PR #10, opened 2015-07-08
Install all sub-packages recursively - fixes #7

### LINKED ISSUE #7: 'python setup.py install' leaves out several subpackages
Hi. 

Running the current version of setup.py will leave out several of the subpackages, e.g. arctic.store and arctic.tickstore.

Instead of using "packages=['arctic', 'tests']" in setup.py, would it not be better to use "packages=find_packages()"? After applying this change it installs nicely.

I'm on OSX so I don't have a pull that fixes this due to clang and OpenMP lacking compatibility leading to other changes, and thus lots of failing unittests.

alessio:


jamesblackburn:
Thanks for the PR!

PR #12, opened 2015-07-10
Create a logo for arctic

burrowsa:
To cement the arctic brand (and to allow us to make stickers!!!!) we need a logo. Here is a first stab at something, it is supposed to look like an arctic landscape that morphs into a timeseries chart. Of course we could just have a cute polar bear. I won't be at all offended if you don't like it.


jamesblackburn:
Great, thanks Andy!

PR #18, opened 2015-07-24
Snapshot specific symbol versions

mstampfer:
Added versions parameter to snapshot specific versions of symbols in version store.

### DIFF
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -678,6 +678,8 @@ class VersionStore(object):
             an optional dictionary of metadata to persist along with the symbol.
         skip_symbols : `collections.Iterable`
             optional symbols to be excluded from the snapshot
+        versions: `dict`
+            an optional dictionary of versions of the symbols to be snapshot
         """
         # Ensure the user doesn't insert duplicates
         snapshot = self._snapshots.find_one({'name': snap_name})
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -666,9 +666,9 @@ class VersionStore(object):
         # Create the audit entry
         mongo_retry(self._audit.insert_one)(audit)
 
-    def snapshot(self, snap_name, metadata=None, skip_symbols=None):
+    def snapshot(self, snap_name, metadata=None, skip_symbols=None, versions=None):
         """
-        Snapshot the current versions of symbols in the library.  Can be used like:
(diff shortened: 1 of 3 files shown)


arctic/store/version_store.py
```diff
@@ -666,9 +666,9 @@ def _write_audit(self, user, message, changed_version):
         # Create the audit entry
         mongo_retry(self._audit.insert_one)(audit)
 
-    def snapshot(self, snap_name, metadata=None, skip_symbols=None):
+    def snapshot(self, snap_name, metadata=None, skip_symbols=None, versions=None):
         """
-        Snapshot the current versions of symbols in the library.  Can be used like:
+        Snapshot versions of symbols in the library.  Can be used like:
```
jamesblackburn: Would be good to document the form for the `versions` argument

arctic/store/version_store.py
```diff
@@ -688,22 +688,23 @@ def snapshot(self, snap_name, metadata=None, skip_symbols=None):
...
-            skip_symbols = set(skip_symbols)
+        if versions is None:
+            versions = {sym: None for sym in set(self.list_symbols()) - skip_symbols}
 
         # Loop over, and snapshot all versions except those we've been asked to skip
-        for sym in set(self.list_symbols()) - skip_symbols:
+        for sym in versions:
             try:
-                sym = self._read_metadata(sym, read_preference=ReadPreference.PRIMARY)
+                sym = self._read_metadata(sym, read_preference=ReadPreference.PRIMARY, as_of=versions[sym])
```
jamesblackburn: Assuming here that every symbol in the library is in the `versions` argument?  What if the library contains additional symbols (or a symbol isn't specified in skip_symbols)?  Perhaps use `get()`?

jamesblackburn:
Would be good to have a test in `test_version_store.py` for the new functionality

arctic/store/version_store.py
richardbounds: I don't think it is necessary for every symbol to be in `versions` - it is likely to be true, but in our use-case we care about including the symbols that we've passed in.

richardbounds:
The change looks good, but agree it needs a test :)

arctic/store/version_store.py
jamesblackburn: Ah, sorry I misread the code, I missed the `versions is None` bit...

PR #20, opened 2015-07-31
Tidy up logging

eeaston:
Makes the library a bit more downstream-user friendly wrt the use of logging - basically removes the singleton logger and only calls logging.basicConfig for the console scripts.
Also a bunch of little PEP8 fixes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -125,6 +125,7 @@ It wouldn't be possible without the work of the AHL Data Engineering Team includ
  * Drake Siard
  * [Slavi Marinov](https://github.com/slavi)
  * [Wilfred Hughes](https://github.com/wilfred)
+ * [Edward Easton](https://github.com/eeaston)
  * ... and many others ...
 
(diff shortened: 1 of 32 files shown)


Wilfred:
Looks good to me. :+1: 

If possible, please keep PEP8ification in a separate commit to functionality changes -- it makes changes easier to understand, especially if using git-bisect.

jamesblackburn:
Looks great, thanks!

Is it possible to configure the logging level for the whole arctic library in one place, or does it get configured module by module now?

eeaston:
@Wilfred - agreed, will separate them out for clarity
@jamesblackburn  - yes, having logging namespaces (created by the 'logger = logging.getLogger(__name__)' where __name__ is 'arctic.x.y'  means that you can set up the whole package's logging with a qualname of 'arctic'.

jamesblackburn:
great, thanks @eeaston

PR #21, opened 2015-08-03
Append functionality with additional functions for naive/user-friendly datetime read/write

theahura:
Summary: 

Functions added to make arctic more usable and consistent.

Details: 
Added API 'tickstore_addons' to be used instead of the native arctic library read/write, and added append functionality. These functions all function with datetime objects that can be either timezone aware or naive, making it easier for the end user to quickly pull or write data. 

Within arctic, added function to get updated date ranges for a given tickstore. Added ability to cleanly overwrite previous dates with a write request, while leaving the non-date-override intact. Added function to convert a tickstore bucket to a version store image. 

Testing: 
tickstore_addons can be imported with 'from arctic.tickstore import tickstore_addons'. The rest of the functions can be tested by setting up a tickstore instance and running various functions. 

Documentation: 
Comments included in code.


jamesblackburn:
Many thanks for the pull request!  Would definitely like to get this new functionality in, but will need to break it into smaller pieces and add tests for the new code first.

jamesblackburn:
Needs tests etc. to be merged back in.

PR #38, opened 2015-10-29
Speedup saving of Pandas Series/Dataframes with MultiIndex containing a date

jamesblackburn:
From Tom: Saving a 3GB multi-index Pandas DF takes >10 minutes:

```
from pandas import DataFrame, date_range
from datetime import datetime as dt
from itertools import product
import numpy as np
from pandas.tools.util import cartesian_product
 
dates = date_range(dt(2000, 1, 1), dt(2015, 1, 1))
ids = range(6000)
idx = cartesian_product((dates, ids))
 
df = DataFrame(index=idx)
for c in range(12):
    df['c_%s' % c] = np.random.randn(len(df))
 
size = (df.values.nbytes + df.index.nbytes + df.columns.nbytes) / 1e6
print size
 
from arctic import Arctic
from datetime import datetime as dt
m = Arctic('research')['tptaylor.alphas']
now = dt.now()
m.write('xx', df)
print dt.now() – now
```

output

### DIFF
--- a/arctic/store/_pandas_ndarray_store.py
+++ b/arctic/store/_pandas_ndarray_store.py
@@ -21,7 +21,7 @@ def _to_primitive(arr):
     if arr.dtype.hasobject:
         if len(arr) > 0:
             if isinstance(arr[0], Timestamp):
-                return arr.astype(DTN64_DTYPE)
+                return np.array([t.value for t in arr], dtype=DTN64_DTYPE)
         return np.array(list(arr))
     return arr
 
@@ -35,7 +35,7 @@ class PandasStore(NdarrayStore):
         if isinstance(index, MultiIndex):
             # array of tuples to numpy cols. copy copy copy
             if len(df) > 0:
-                ix_vals = map(np.array, zip(*index.values))
+                ix_vals = map(np.array, [index.get_level_values(i) for i in range(index.nlevels)])
             else:
                 # empty multi index has no size, create empty arrays for recarry..
                 ix_vals = [np.array([]) for n in index.names]
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,16 @@
 
 ## Changelog
 
+### 1.11 (2015-10-29)
+
+  * Bugfix: Improve performance of saving multi-index Pandas DataFrames
+    by 9x
+
+### 1.10 (2015-10-28)
+
+  * Bugfix: VersionStore.read(date_range=...) could do the wrong thing with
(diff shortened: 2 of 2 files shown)


arctic/store/_pandas_ndarray_store.py
```diff
@@ -21,7 +21,7 @@ def _to_primitive(arr):
     if arr.dtype.hasobject:
         if len(arr) > 0:
             if isinstance(arr[0], Timestamp):
-                return arr.astype(DTN64_DTYPE)
+                return np.array([t.value for t in arr], dtype=DTN64_DTYPE)
```
jamesblackburn: @TomTaylorLondon 
> arr.astype() is hellishly slow, simply getting the values runs in less than second.
> numpy.fromiter was a little slower so I kept it as a list comprehension (you're going to end up with a copy of the index anyway...)

arctic/store/_pandas_ndarray_store.py
```diff
@@ -35,7 +35,7 @@ def _index_to_records(self, df):
         if isinstance(index, MultiIndex):
             # array of tuples to numpy cols. copy copy copy
             if len(df) > 0:
-                ix_vals = map(np.array, zip(*index.values))
+                ix_vals = map(np.array, [index.get_level_values(i) for i in range(index.nlevels)])
```
jamesblackburn: @TomTaylorLondon
> zip(*index.values) has two issues:
> (1) Firstly its slow, much slower than get_level_values
> (2) It returns the datetime index as TimeStamp. get_level_values converts to dt64 under the hood in Cython removing the TimeStamp bottleneck.

jamesblackburn:
>These speedups bring the saving of a 550mb DataFrame with a multiindex (see below) from 2.16s to 15s (same performance as w/o the index).

```
dates = date_range(dt(2000, 1, 1), dt(2015, 1, 1))
ids = range(1000)
idx = cartesian_product((dates, ids))
df = DataFrame(index=idx)
for c in range(12):
    df['c_%s' % c] = np.random.randn(len(df))
```

## Original

original

## With TS fix

multi_fix_tstamp_conversion

## With both fixes

multi_fix_tstamp_conversion_and_zip

maximilianr:
What tool generated the charts? They're superb.

bmoscon:
@TomTaylorLondon used https://github.com/jrfonseca/gprof2dot

TomTaylorLondon:
@maximilianr gprof2dot. What's neat is that it's written in Python so you can always modify the output to include additional information.

maximilianr:
Thanks!

PR #40, opened 2015-10-29
EASY-511 return deleted state in list_versions

eeaston:
This way we can check whether or not the leading version is just a sentinel put in by version_store.delete().


jamesblackburn:
+1 LGTM

PR #42, opened 2015-11-01
Fixed missing packages in building (.store etc), removed cython and

bjorskog:
compiling of extension compress for OS X support.


bjorskog:
Ooops - fatfinger syndrome.

PR #44, opened 2015-11-04
Remove the version pin for mock

burrowsa:
this was there because mockextras didn't work with the latest version of mock but now it does


burrowsa:
Turns out that there are other problems using the latest mock with artic

PR #47, opened 2015-11-12
Bugfix/mdp 621 should mongo retry in the mongoose.  getitem  

mildbyte:


mildbyte:
This is to mitigate an issue we've had (https://jira.mongodb.org/browse/PYTHON-1012) where, if a server is overloaded, we might be able to connect to it, but the authentication would time out (raising an AutoReconnect). We now retry the whole connection initialisation if issues occur.

Issue #50, opened 2015-11-20
Allow read_metadata to expose internal information about the data


bmoscon:
per @TomTaylorLondon  sometimes you want to determine information about the data without actually reading it back (i.e. type, number of columns, column names, etc). This allows the user to get this data via VersionStore::read_metadata (or via any VersionedItem).

bmoscon:
fixed with merge #51

PR #51, opened 2015-11-20
Expose data info via read_metadata

bmoscon:
see issue #50

### DIFF
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -106,6 +106,16 @@ class VersionStore(object):
     def __repr__(self):
         return str(self)
     
+    def _get_data_info(self, version):
+        ret = {}
+        if type in version:
+            ret['type'] = version['type']
+        if 'dtype_metadata' in version:
+            ret['col_names'] = version['dtype_metadata']
+        if 'dtype' in version:
+            ret['col_dtypes'] = version['dtype']
+        return ret
+
     def _read_preference(self, allow_secondary):
         """ Return the mongo read preference given an 'allow_secondary' argument
         """
@@ -366,7 +376,7 @@ class VersionStore(object):
         if data is None:
             raise NoDataFoundException("No data found for %s in library %s" % (symbol, self._arctic_lib.get_name()))
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=version['version'],
-                             metadata=version.pop('metadata', None), data=data)
+                             metadata=version.pop('metadata', None), data=data, data_info=self._get_data_info(version))
     _do_read_retry = mongo_retry(_do_read)
 
     @mongo_retry
@@ -392,7 +402,7 @@ class VersionStore(object):
         """
         _version = self._read_metadata(symbol, as_of=as_of, read_preference=self._read_preference(allow_secondary))
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=_version['version'],
-                             metadata=_version.pop('metadata', None), data=None)
+                             metadata=_version.pop('metadata', None), data=None, data_info=self._get_data_info(_version))
 
     def _read_metadata(self, symbol, as_of=None, read_preference=None):
         if read_preference is None:
@@ -462,7 +472,7 @@ class VersionStore(object):
 
         if len(data) == 0 and previous_version is not None:
             return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=previous_version,
-                                 metadata=version.pop('metadata', None), data=None)
+                                 metadata=version.pop('metadata', None), data=None, data_info=self._get_data_info(previous_version))
 
         if upsert and previous_version is None:
             return self.write(symbol=symbol, data=data, prune_previous_version=prune_previous_version, metadata=metadata)
@@ -513,7 +523,7 @@ class VersionStore(object):
             self._prune_previous_versions(symbol)
 
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=version['version'],
-                             metadata=version.pop('metadata', None), data=None)
+                             metadata=version.pop('metadata', None), data=None, data_info=self._get_data_info(version))
 
     def _publish_change(self, symbol, version):
         if self._publish_changes:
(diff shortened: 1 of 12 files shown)


bmoscon:
I'll wait to hear from @TomTaylorLondon and @jamesblackburn before merging this in

arctic/store/version_store.py
```diff
@@ -106,6 +106,16 @@ def __str__(self):
     def __repr__(self):
         return str(self)
     
+    def _get_data_info(self, version):
```
TomTaylorLondon: Delegate this down to the _arctic_lib (in our instance PandasNDArrayStore) - that way we can implement different implementations for multiple types.

TomTaylorLondon:
Thanks Bryant. Can you push the implementation down to Pandas/NDArray layer and add in specific support for dataframes/series - index cols, index names, date range (you can re-use the existing logic to find the first date column & you'll need to be clever on handling the range on appends)

bmoscon:
@TomTaylorLondon, those fields (names, index, etc) are already populated in _ndarray_store.py, so I didn't need to add anything for them. 

I will move "get_data_info" to the lower levels. For now I'll probably just stub it out for the other types.

tests/integration/store/test_pandas_store.py
```diff
@@ -797,3 +796,27 @@ def test_daterange_fails_with_timezone_start(library):
     with pytest.raises(ValueError):
         library.read('MYARR', date_range=DateRange(start=dt(2015, 1, 1, tzinfo=mktz())))
 
+def test_data_info_series(library):
```
jamesblackburn: For these new `data_info` tests, it would be good to assert the shape and content of the data_info property

bmoscon:
@jamesblackburn, will do. 

@TomTaylorLondon, I've spent some time looking into what you requested. I don't think its possible. the other storage types all populate different parts of the version dict. When we are creating a VersionedItem, we have no access to the underlying libraries, just the data that has been returned from Mongo, in the form of this dict. Unfortunately, whatever populates the data info will need to basically know about all the other lower layers, or it can return an unformatted version dict back that an upper layer can make sense of. We can discuss.

bmoscon:
actually, @TomTaylorLondon, I think I do have a way. Get in touch with me so we can chat.

jamesblackburn:
@bmoscon  How are you intending to use this dtype info? Presumably the easy way to do push the behaviour down is to use `_read_handler()` to get the implementation class and call the method on that...

bmoscon:
@TomTaylorLondon and @jamesblackburn - changes made. Let me know what you think

tests/integration/store/test_pandas_store.py
```diff
@@ -797,3 +796,26 @@ def test_daterange_fails_with_timezone_start(library):
...
+    s = Series(data=[1, 2, 3], index=[4, 5, 6])
+    library.write('pandas', s)
+    md = library.read('pandas').data_info
+    assert md == library.read_metadata('pandas').data_info
+    assert md == {'col_dtypes': "[('index', '<i8'), ('values', '<i8')]", 'col_names': {'columns': ['values'], 'index': ['index']}, 'type': 'pandasseries'}
```
jamesblackburn: Would it be useful for the `col_dtypes` to be a list of tuples (rather than a string __repr__)?

jamesblackburn:
LGTM, thanks!

tests/integration/store/test_pandas_store.py
bmoscon: Ah, I didn't even notice it was a string. I'll fix that and merge it in the AM. thanks!

TomTaylorLondon:
Looks good, thanks.

richardbounds:
What is the difference between get_info() and get_data_info()?

jamesblackburn:
That's a good point.  Probably worth having just one method 'get_info' and have the results of that be part of the VersionedItem

bmoscon:
So i assume then, @jamesblackburn and @richardbounds, that I can change the result that get_info returns? It currently returns a formatted string

bmoscon:
i don't see get_info being used in many places. Really its just in _show_info, which does not appear to be used anywhere at all. I could really just totally change get_info and remove the show_info method since its not in use

richardbounds:
get_info was only ever used for interactive debugging, changing it is fine.

arctic/store/versioned_item.py
```diff
@@ -1,19 +1,20 @@
 from collections import namedtuple
 
 
-class VersionedItem(namedtuple('VersionedItem', ['symbol', 'library', 'data', 'version', 'metadata'])):
+class VersionedItem(namedtuple('VersionedItem', ['symbol', 'library', 'data', 'version', 'metadata', 'info'])):
```
richardbounds: Why is info on VersionedItem? What is it used for?

arctic/store/version_store.py
```diff
@@ -366,7 +364,7 @@ def _do_read(self, symbol, version, from_version=None, **kwargs):
         if data is None:
             raise NoDataFoundException("No data found for %s in library %s" % (symbol, self._arctic_lib.get_name()))
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=version['version'],
-                             metadata=version.pop('metadata', None), data=data)
+                             metadata=version.pop('metadata', None), data=data, info=self._get_info(symbol, version=version))
```
richardbounds: Just call the handler here.

arctic/store/version_store.py
```diff
@@ -391,8 +389,9 @@ def read_metadata(self, symbol, as_of=None, allow_secondary=None):
             `False` : only allow reads from primary members
         """
         _version = self._read_metadata(symbol, as_of=as_of, read_preference=self._read_preference(allow_secondary))
+        handler = self._read_handler(_version, symbol)
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=_version['version'],
-                             metadata=_version.pop('metadata', None), data=None)
+                             metadata=_version.pop('metadata', None), data=None, info=self._get_info(symbol, version=_version))
```
richardbounds: Just call the handler here

arctic/store/version_store.py
```diff
@@ -461,8 +460,9 @@ def append(self, symbol, data, metadata=None, prune_previous_version=True, upser
                                                    sort=[('version', pymongo.DESCENDING)])
 
         if len(data) == 0 and previous_version is not None:
+            handler = self._read_handler(previous_version, symbol)
             return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=previous_version,
-                                 metadata=version.pop('metadata', None), data=None)
+                                 metadata=version.pop('metadata', None), data=None, info=self._get_info(symbol, version=previous_version))
```
richardbounds: Just call the handler...

PR #55, opened 2015-11-25
Add support for returning additional information in the VersionedItem

jamesblackburn:
I've put this back on a branch for now @bmoscon @TomTaylorLondon @richardbounds until we can agree on the implementation.

### DIFF
--- a/arctic/store/_base_store.py
+++ b/arctic/store/_base_store.py
@@ -1,0 +1,10 @@
+class BaseStore(object):
+    def read(self, lib, version, symbol, **kwargs):
+        pass
+
+    def write(self, lib, version, symbol, item, previous_version):
+        pass
+
+    def get_info(self, lib, version, symbol, **kwargs):
+        pass
+
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -7,6 +7,7 @@ import numpy as np
 import pymongo
 from pymongo.errors import OperationFailure, DuplicateKeyError
 
+from ._base_store import BaseStore
 from ..decorators import mongo_retry, dump_bad_documents
 from ..exceptions import UnhandledDtypeException
 from ._version_store_utils import checksum
@@ -36,7 +37,7 @@ def _promote_struct_dtypes(dtype1, dtype2):
     return np.dtype([(n, _promote(dtype1.fields[n][0], dtype2.fields.get(n, (None,))[0])) for n in dtype1.names])
 
 
-class NdarrayStore(object):
+class NdarrayStore(BaseStore):
     """Chunked store for arbitrary ndarrays, supporting append.
     
     for the simple example:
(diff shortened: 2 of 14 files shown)


arctic/store/_ndarray_store.py
```diff
@@ -151,26 +152,20 @@ def _index_range(self, version, symbol, from_version=None, **kwargs):
...
-
-        est_size = dtype.itemsize * length
-        return """Handler: %s
-
-dtype: %s
+                'segment': {'$lt': ret['length']}}
 
-%d rows in %d segments
-Data size: %s bytes
+        ret['n_segments'] = collection.find(spec).count()
```
jamesblackburn: This DB query may add a bit of overhead on the query perf.  We've got the `segment_count` field on version document - can probably just do that instead?

arctic/store/_ndarray_store.py
```diff
@@ -151,26 +152,20 @@ def _index_range(self, version, symbol, from_version=None, **kwargs):
...
 
-Version document:
-%s""" % (self.__class__.__name__, dtype, length, n_segments, est_size, pprint.pformat(version))
+        ret['est_size'] = ret['dtype'].itemsize * ret['length']
+        ret['handler'] = self.__class__.__name__
+	return ret
```
jamesblackburn: Whitespace here - we should only be using spaces

arctic/store/version_store.py
```diff
@@ -352,13 +352,11 @@ def _show_info(self, symbol, as_of=None):
...
         print self._get_info(symbol, as_of)
 
-    def _get_info(self, symbol, as_of=None):
-        _version = self._read_metadata(symbol, as_of=as_of)
-        handler = self._read_handler(_version, symbol)
-        if hasattr(handler, "get_info"):
-            return handler.get_info(self._arctic_lib, _version, symbol)
-        else:
-            return """Handler: %s\n\nVersion document:\n%s""" % (handler.__class__.__name__, pprint.pformat(_version))
+    def _get_info(self, symbol, as_of=None, version=None):
```
jamesblackburn: Elsewhere in the API we use `as_of` as the selector we use when slicing out a particular version:
   https://github.com/manahl/arctic/blob/master/arctic/store/version_store.py#L186
Is there a reason to make this take `version` too?

arctic/store/versioned_item.py
```diff
@@ -1,19 +1,20 @@
 from collections import namedtuple
 
 
-class VersionedItem(namedtuple('VersionedItem', ['symbol', 'library', 'data', 'version', 'metadata'])):
+class VersionedItem(namedtuple('VersionedItem', ['symbol', 'library', 'data', 'version', 'metadata', 'info'])):
```
jamesblackburn: A lot of the change comes about due to the addition of `info` to the API here.  I was wondering if it might be better just to have a method `get_info` on the `VersionStore` and let clients that care use that?

arctic/store/_base_store.py
```diff
@@ -0,0 +1,10 @@
+class BaseStore(object):
```
jamesblackburn: Would be great to document the base class

arctic/store/versioned_item.py
richardbounds: Agree with @jamesblackburn - I'd prefer to avoid pushing this into everything if possible. I guess it is in here because the our graphs just pull the VersionedItem and do everything with that.

bmoscon:
I'd say just jettison this. There really isnt a good reason to add this data to the VersionedItem. We just need a reasonable way to get the information back without having to read the entire object back from Mongo. I've submitted a new pull request, take a look at that.

PR #56, opened 2015-11-25
new VersionStore API - get_data_info

bmoscon:
allows user to get information about stored data without reading back the whole df/series/etc

see: NI-370

### DIFF
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -150,27 +150,17 @@ class NdarrayStore(object):
             from_index = from_version['up_to']
         return from_index, None
 
-    def get_info(self, arctic_lib, version, symbol, **kwargs):
-        collection = arctic_lib.get_top_level_collection()
+    def get_info(self, version):
+        ret = {}
         dtype = self._dtype(version['dtype'], version.get('dtype_metadata', {}))
         length = int(version['up_to'])
-
-        spec = {'symbol': symbol,
-                'parent': version.get('base_version_id', version['_id']),
-                'segment': {'$lt': length}}
-
-        n_segments = collection.find(spec).count()
-
-        est_size = dtype.itemsize * length
-        return """Handler: %s
-
-dtype: %s
-
-%d rows in %d segments
-Data size: %s bytes
-
-Version document:
-%s""" % (self.__class__.__name__, dtype, length, n_segments, est_size, pprint.pformat(version))
+        ret['size'] = dtype.itemsize * length
+        ret['segment_count'] = version['segment_count']
+        ret['dtype'] = version['dtype']
+        ret['type'] = version['type']
+        ret['handler'] = self.__class__.__name__
+        ret['rows'] = int(version['up_to'])
+        return ret
 
     def read(self, arctic_lib, version, symbol, read_preference=None, **kwargs):
         index_range = self._index_range(version, symbol, **kwargs)
--- a/arctic/store/_pandas_ndarray_store.py
+++ b/arctic/store/_pandas_ndarray_store.py
@@ -4,6 +4,7 @@ from bson.binary import Binary
 from pandas import DataFrame, MultiIndex, Series, DatetimeIndex, Panel
 from pandas.tslib import Timestamp, get_timezone
 import numpy as np
+import ast
 
 from .._compression import compress, decompress
 from ..exceptions import ArcticException
@@ -194,6 +195,17 @@ class PandasStore(NdarrayStore):
         if date_range:
             item = self._daterange(item, date_range)
         return item
+
+    def get_info(self, version):
+        """
+        parses out the relevant information in version
+        and returns it to the user in a dictionary
+        """
+        ret = super(PandasStore, self).get_info(version)
+        ret['col_names'] = version['dtype_metadata']
+        ret['handler'] = self.__class__.__name__
+        ret['dtype'] = ast.literal_eval(version['dtype'])
+        return ret
 
 
 def _start_end(date_range, dts):
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -335,10 +335,9 @@ class VersionStore(object):
             raise
 
     @mongo_retry
-    def _show_info(self, symbol, as_of=None):
+    def get_info(self, symbol, as_of=None):
         """
-        Print details on the stored symbol: the underlying storage handler
-        and the version_document corresponding to the specified version.
+        Reads and returns information about the data stored for symbol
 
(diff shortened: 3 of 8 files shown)


.pydevproject
```diff
@@ -3,6 +3,6 @@
 <pydev_property name="org.python.pydev.PYTHON_PROJECT_INTERPRETER">Default</pydev_property>
 <pydev_property name="org.python.pydev.PYTHON_PROJECT_VERSION">python 2.7</pydev_property>
 <pydev_pathproperty name="org.python.pydev.PROJECT_SOURCE_PATH">
-<path>/arctic</path>
+<path>/arctic_working</path>
```
jamesblackburn: Not needed for this change

arctic/store/_pandas_ndarray_store.py
```diff
@@ -4,6 +4,7 @@
 from pandas import DataFrame, MultiIndex, Series, DatetimeIndex, Panel
 from pandas.tslib import Timestamp, get_timezone
 import numpy as np
+import ast
```
jamesblackburn: Unused import

arctic/store/_pandas_ndarray_store.py
```diff
@@ -195,6 +196,17 @@ def read(self, arctic_lib, version, symbol, read_preference=None, date_range=Non
             item = self._daterange(item, date_range)
         return item
 
+    def get_data_info(self, version):
+        """
+        parses out the relevant information in version
+        and returns it to the user in a dictionary
+        """
+        ret = {}
+        ret['dtype'] = ast.literal_eval(version['dtype'])
```
jamesblackburn: Gosh looks magical :)

arctic/store/_pandas_ndarray_store.py
```diff
@@ -195,6 +196,17 @@ def read(self, arctic_lib, version, symbol, read_preference=None, date_range=Non
             item = self._daterange(item, date_range)
         return item
 
+    def get_data_info(self, version):
```
jamesblackburn: `get_info` would probably be a good simple name for this

arctic/store/version_store.py
```diff
@@ -360,6 +360,27 @@ def _get_info(self, symbol, as_of=None):
         else:
             return """Handler: %s\n\nVersion document:\n%s""" % (handler.__class__.__name__, pprint.pformat(_version))
 
+    def get_data_info(self, symbol):
```
jamesblackburn: `get_info`, should take `as_of`, as the person using this will have a `VerisonedItem` in their hand which will have a : `(symbol, version_number)` in it.

arctic/store/_pandas_ndarray_store.py
bmoscon: I agree get_info is nicer, but its already used by the parent class and other storage types for something different.

jamesblackburn:
Would it be possible to replace the `_show_info` / `_get_info` internal methods on VersionStore?

As a minor thing: would be good to have Jira / Issue numbers in the commit comments.

Otherwise looks like the right approach!

arctic/store/_pandas_ndarray_store.py
bmoscon: is there a change you'd like to me make? Wasn't sure the best way to convert a string of a list of tuples to a list of tuples.

arctic/store/version_store.py
bmoscon: just as of? The method is intended to be used like read_metadata

bmoscon:
Yes, I can add the JIRA ticket number. Why do you want me to replace the show info? Doing that will break other things that use arctic (as I discovered with the original pull request). Nothing IN arctic uses that method, but plenty of other AHL modules do.

arctic/store/_pandas_ndarray_store.py
jamesblackburn: Nope just learned something new. Perhaps @Wilfred has thoughts

arctic/store/version_store.py
jamesblackburn: Looks like the only other parameter is allow_secondary - which is a bit of an implementation detail... Symbol and as_of are the key ones I think.

jamesblackburn:
I think @richardbounds said _show_info was only used for debugging and not actually used in any apps. This seems to expose similar information? If so would be good to fold one into the other.

bmoscon:
I just remember having to fix unit tests because there were tests in mongo or signals that checked its output

richardbounds:
I'd vote for folding everything into get_info and changing it as you need to. I'm pretty sure "_show_info" isn't actually used in anger anywhere, particularly because it only returns a formatted string at the moment. I guess any tests are just asserting that the output hasn't changed and it doesn't blow up.

arctic/store/_pandas_ndarray_store.py
Wilfred: This is the right way to parse literals. The only alternative is a serialisation format like JSON.

bmoscon: thanks for the confirmation @Wilfred

bmoscon:
alright @jamesblackburn, let me know what you think

arctic/store/_ndarray_store.py
```diff
@@ -150,27 +150,12 @@ def _index_range(self, version, symbol, from_version=None, **kwargs):
...
-        return """Handler: %s
-
-dtype: %s
-
-%d rows in %d segments
-Data size: %s bytes
-
-Version document:
-%s""" % (self.__class__.__name__, dtype, length, n_segments, est_size, pprint.pformat(version))
+    def get_info(self, version):
```
jamesblackburn: Can we preserver some of the data from above into this?

I suspect we can get the following:
 - number of segments `version['segment_count']`
 - number of rows
 - data size

from just the version document.

bmoscon: sure thing

arctic/store/_ndarray_store.py
```diff
@@ -150,27 +150,12 @@ def _index_range(self, version, symbol, from_version=None, **kwargs):
...
+        ret = {}
+        ret['dtype'] = version['dtype']
```
jamesblackburn: _padnas_ndarray_store interprets the dtype, whereas this base-class just includes it as a string, do we want to `ast.literal_eval` here too?

bmoscon: no, because the dtype of a numpy array will more than likely be a string like float64, or uint8. it wont be a list.

richardbounds:
Looks ok to me.

Issue #37, opened 2015-10-28
how_to_use_arctic.py fails with SyntaxError: unexpected EOF while parsing on library.read()


jeffneuen:
Attempting to run the demo as a test, and getting errors on library.read().  Also, I get an error the first time I run store.initialize_library('username.scratch'), but the second time it did works.  I did have to make the following change to get import Arctic to run:

store/_version_store_utils.py, line 66:    if pd.__version__.startswith("0.14"):

See paste of output included the EOF error below.  I'm not sure if it's a numpy bug or arctic, but I'm not sure where to go from here.

Python 2.7.10 |Anaconda 2.3.0 (64-bit)| (default, Oct 19 2015, 18:04:42) 
[GCC 4.4.7 20120313 (Red Hat 4.4.7-1)] on linux2
Type "help", "copyright", "credits" or "license" for more information.
Anaconda is brought to you by Continuum Analytics.
Please check out: http://continuum.io/thanks and https://anaconda.org
>>> from arctic import Arctic
>>> from datetime import datetime as dt
>>> import pandas as pd
>>> store = Arctic('localhost')
>>> store.list_libraries()
[u'NASDAQ', u'username.scratch']
>>> store.initialize_library('username.scratch')
No handlers could be found for logger "arctic.store.version_store"
>>> store.initialize_library('username.scratch')
>>> library = store['username.scratch']
>>> df = pd.DataFrame({'prices': [1, 2, 3]},[dt(2014, 1, 1), dt(2014, 1, 2), dt(2014, 1, 3)])
>>> library.write('SYMBOL', df)
VersionedItem(symbol=SYMBOL,library=arctic_username.scratch,data=<type 'NoneType'>,version=19,metadata=None
>>> library.read('SYMBOL')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/jeff/anaconda/lib/python2.7/site-packages/arctic/store/version_store.py", line 319, in read
    date_range=date_range, read_preference=read_preference, **kwargs)
  File "/home/jeff/anaconda/lib/python2.7/site-packages/arctic/store/version_store.py", line 363, in _do_read
    data = handler.read(self._arctic_lib, version, symbol, from_version=from_version, **kwargs)
  File "/home/jeff/anaconda/lib/python2.7/site-packages/arctic/store/_pandas_ndarray_store.py", line 279, in read
    item = super(PandasDataFrameStore, self).read(arctic_lib, version, symbol, **kwargs)
[...]

PR #58, opened 2015-12-06
Fix issue with new versions of NumPy that cause dtype metadata to be lost

bmoscon:
This addresses issue #37 

In new versions of NumPy there is an issue ( see https://github.com/numpy/numpy/issues/6771 ) that causes the dtype metadata to be lost. Additionally, the actual dtype has changed when using NumPy records as well. This pull request addresses these issues and works with the current supported versions of NumPy and Pandas (1.9.2 and 0.16.2) as well as the latest versions (1.10.1 and 0.17.1)

### DIFF
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -210,35 +210,38 @@ class NdarrayStore(object):
         rtn = np.fromstring(data, dtype=dtype).reshape(version.get('shape', (-1)))
         return rtn
 
-    def _promote_types(self, item, dtype_str):
-        if dtype_str == str(item.dtype):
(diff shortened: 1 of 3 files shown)


bmoscon:
@jamesblackburn let me know what you think. I'm not sure who else should be asked to review the code.  The changes are not very significant, but it does touch some core components.

jamesblackburn:
Looks good to me @bmoscon 

Regarding the unit test removed, I see there's already an integration test for storing a DF with an object column:
https://github.com/manahl/arctic/blob/master/tests/integration/store/test_pandas_store.py#L514

So I think I agree this integration test is superfluous.

PR #60, opened 2015-12-15
Add support non-audited transactions.

jamesblackburn:
This is useful for batch jobs that require 'transactional' access to
VersionStore (i.e. read-modify-write without intervening update), but
don't necessarily want to keep all versions in perpetuity.


richardbounds:
lgtm

bmoscon:
looks good!

PR #70, opened 2015-12-24
Use setuptools-git

eeaston:
Removed MANIFEST file and use setuptools-git instead, its more fool-proof to include all the header and source files. Updated changelog.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,11 @@
 
 ## Changelog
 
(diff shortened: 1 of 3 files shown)


AdrianTeng:
LGTM.
Are we changing CircleCI build so that it installs from sdist as well?

eeaston:
Yes, but this release needs to get out pronto so I'll do that first :)

PR #85, opened 2016-01-05
Unit tests need to be prefixed with test_

bmoscon:
The integration tests in integration/tickstore/test_toplevel.py are not prefixed with test. Py.test will not run any of these tests. 

I added "test_" to the tests and fixed a few errors in the code. Some tests still will not pass (either because of a bug in the test or in the code), so these have been marked xfail. Whoever wrote them should go back and take a look at the broken tests. (I'll open a ticket for them once this gets merged in).

### DIFF
--- a/tests/integration/tickstore/test_toplevel.py
+++ b/tests/integration/tickstore/test_toplevel.py
@@ -26,7 +26,7 @@ FEED_2012_LEVEL1 = toplevel.TickStoreLibrary('FEED_2012.LEVEL1', DateRange(dt(20
                           (dt(2009, 2, 1), dt(2009, 12, 31), []),
                           (dt(2013, 2, 1), dt(2013, 12, 31), []),
                           ])
-def should_return_libraries_for_the_given_daterange(toplevel_tickstore, start, end, expected):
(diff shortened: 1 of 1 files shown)


AdrianTeng:
FYI we had customised pytest to identify a function as test with both `test` and `should` as prefix. I guess that was lost when we open source arctic and re-write `setup.py` from scratch

bmoscon:
Thanks for the info. If those tests all were passing at somepoint, then that means somewhere along the way something changed that resulted in them breaking :(

jamesblackburn:
I'd guess the tests are sloppy with timezones mixing naive and London.

Top-Level Tickstore is used by us for poor man's sharding.  We store the tick data in a library per year, e.g. `FEED_2013`, `FEED_2014`...

The Top level Tickstore makes them appear as a single library `FEED` and redirects reads with a date range to the right underlying library.

We did this to:
- prevent having very large databases with MMAP
- to aid backup and restore
- to make the sharing and balancing transparent

It's a real pain though.  Ideally we would be able to use MongoDBs sharing for this in the future...

bmoscon:
Could be. most the test failures were leftside dataframe had 1 extra row compared to rightside dataframe

PR #89, opened 2016-01-08
Python3 support

bmoscon:
Support for Python 3.3

Some unit tests (not integration tests) had to be xfailed if the version of python is 3 because we require an old version of mock that doesnt work with python3.3

I tried to squash as many of the commits as I could (I got about 47 squashed).


bmoscon:
@TomTaylorLondon @jamesblackburn finally! Circle Ci wont pass unless you pre-install cython

arctic/scripts/arctic_delete_library.py
```diff
@@ -28,7 +28,7 @@ def main():
     if not opts.library:
         parser.error('Must specify the full path of the library e.g. arctic_jblackburn.lib!')
 
-    print "Deleting: %s on mongo %s" % (opts.library, opts.host)
+    print("Deleting: %s on mongo %s" % (opts.library, opts.host))
```
AdrianTeng: `from __future__ import print_function`?

arctic/store/_ndarray_store.py
```diff
@@ -20,6 +19,11 @@
 _APPEND_SIZE = 1 * 1024 * 1024  # 1MB
 _APPEND_COUNT = 60  # 1 hour of 1 min data
 
+try:
```
AdrianTeng: `from six.moves import xrange` :)

arctic/store/_pickle_store.py
```diff
@@ -1,11 +1,14 @@
 import bson
 from bson.binary import Binary
 from bson.errors import InvalidDocument
-import cPickle
-import cStringIO as stringio
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
```
AdrianTeng: `from six.moves import cPickle`

howtos/how_to_custom_arctic_library.py
```diff
@@ -1,7 +1,11 @@
 from datetime import datetime as dt
 from bson.binary import Binary
-import cPickle
 
+try:
+    import cPickle as pickle
+except ImportError:
+    import pickle
```
AdrianTeng: ditto

AdrianTeng:
Good stuff!

arctic/scripts/arctic_delete_library.py
bmoscon: I could have done that, but python2.7 supports print as a function, i.e. print()

bmoscon:
Code updated. All tests passing (CircleCI with specific dependency versions on 2.7, and travis with latest dependencies on 2.7 and 3.3)

arctic/scripts/arctic_delete_library.py
AdrianTeng: I don't think python 2.7 supports print function without the future import, [as documented here](https://docs.python.org/2/library/functions.html#print). The brackets here are treated as grouping (for 2.7), which has no effect in this case.
Anyway, it works, so that'll do :)

femtotrader: Python 2.7 supports print function without the future import but the way it handles arguments is different if you use future import and if you don't use future import.

See using Python 2.7

```python
$ source activate py2
discarding //anaconda/bin from PATH
prepending //anaconda/envs/py2/bin to PATH
$ ipython
Python 2.7.11 |Anaconda 2.4.1 (x86_64)| (default, Dec  6 2015, 18:57:58)
Type "copyright", "credits" or "license" for more information.

IPython 4.0.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: print("hello ", 2, "people")
('hello ', 2, 'people')

In [2]: from __future__ import print_function

In [3]: print("hello ", 2, "people")
hello  2 people
```

Using Python 3.4 now

```python
$ source deactivate
discarding //anaconda/envs/py2/bin from PATH
$ ipython
Python 3.4.3 |Anaconda 2.3.0 (x86_64)| (default, Oct 20 2015, 14:27:51)
Type "copyright", "credits" or "license" for more information.

IPython 4.0.1 -- An enhanced Interactive Python.
?         -> Introduction and overview of IPython's features.
%quickref -> Quick reference.
help      -> Python's own help system.
object?   -> Details about 'object', use 'object??' for extra details.

In [1]: print("hello ", 2, "people")
hello  2 people
```

So I suggest using future import for print function to have same behavior with Python 2 and Python 3 when you are using several arguments for print (this is not the case here)

bmoscon:
ok, I added 
```
__future__ import print_function
```
to all the files that use print()

AdrianTeng:
Done. Also added OS classifiers. (Didn't add windows as I'm not sure are we supporting it or not)

bmoscon:
Windows is on my list,  but I don't think it currently works.

.travis.yml
```diff
@@ -0,0 +1,26 @@
+sudo: false
+language: python
+python:
+  - "2.7"
+  - "3.3"
+install:
```
jamesblackburn: I realise none of these versions are pinned. Do the versions change regularly? Presumably happy for the build to break as a result?

arctic/store/_pandas_ndarray_store.py
```diff
@@ -186,7 +186,7 @@ def _index_range(self, version, symbol, date_range=None, **kwargs):
                     return -1, -1
                 idxstart = min(np.searchsorted(dts, start), len(dts) - 1)
                 idxend = min(np.searchsorted(dts, end), len(dts) - 1)
-                return index['index'][idxstart], index['index'][idxend] + 1
+                return int(index['index'][idxstart]), int(index['index'][idxend] + 1)
```
jamesblackburn: Would these not be integers by default?

arctic/store/_pickle_store.py
```diff
@@ -57,7 +57,7 @@ def write(self, arctic_lib, version, symbol, item, previous_version):
         version['blob'] = _MAGIC_CHUNKED
         pickled = lz4.compressHC(cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL))
 
-        for i in xrange(len(pickled) / _CHUNK_SIZE + 1):
+        for i in xrange(int(len(pickled) / _CHUNK_SIZE + 1)):
```
jamesblackburn: Or `//`

arctic/store/_version_store_utils.py
```diff
@@ -28,9 +30,13 @@ def checksum(symbol, doc):
     Checksum the passed in dictionary
     """
     sha = hashlib.sha1()
-    sha.update(symbol)
-    for k in sorted(doc.iterkeys(), reverse=True):
-        sha.update(str(doc[k]))
+    sha.update(symbol.encode('ascii'))
```
jamesblackburn: Did you verify this produces the same SHA for the `Binary` chunk data?

Not sure that the string should be re-endcoded - can't it just be passed as a byte-array?

tests/unit/date/test_datetime_to_ms_roundtrip.py
```diff
@@ -45,6 +46,8 @@ def test_pytz_London():
     assert_roundtrip(tz)
 
 
+@pytest.mark.xfail(sys.version_info >= (3,),
+                   reason="known issue with dateutil and python3")
```
jamesblackburn: Which issue is this?

tests/unit/store/test_pickle_store.py
```diff
@@ -12,7 +15,10 @@
 from arctic.store._pickle_store import PickleStore
 from arctic.store._version_store_utils import checksum
 
-PANDAS_VERSION = LooseVersion(pd.version.version)
+try:
+    PANDAS_VERSION = LooseVersion(pd.version.version)
```
jamesblackburn: I think `__version__` is there in all cases

tests/unit/store/test_pickle_store.py
```diff
@@ -63,7 +69,8 @@ def test_read_object_2():
     assert PickleStore.read(self, arctic_lib, version, sentinel.symbol) == object
     assert coll.find.call_args_list == [call({'symbol': sentinel.symbol, 'parent': sentinel._id}, sort=[('segment', 1)])]
 
-
+@pytest.mark.xfail(sys.version_info >= (3,),
+                   reason="lz4 data written with python2 not compatible with python3")
```
jamesblackburn: `lz4 data written with python2 not compatible with python3` 

What's going on here?  Is lz4 the issue?

tests/unit/tickstore/test_toplevel.py
```diff
@@ -131,6 +132,8 @@ def test_slice_list_of_dicts(start, end, expected_start_index, expected_end_inde
     assert expected == result
 
 
+@pytest.mark.xfail(sys.version_info >= (3,),
+                   reason="python3 issue with mock 1.0.1")
```
jamesblackburn: There are a bunch of these mock issues - can we resolve before merge so we don't lose these tests?

arctic/store/_version_store_utils.py
```diff
@@ -28,9 +30,13 @@ def checksum(symbol, doc):
...
+    for k in sorted(iter(doc.keys()), reverse=True):
```
jamesblackburn: Can we add a unit test that, given a known document `checksum` returns a know good checksum?

jamesblackburn:
Looking really good, a few comments along the way

.travis.yml
bmoscon: Circle CI has the versions pinned - thats what we want I assume to ensure support with the AHL libraries. I think having Travis use the latest version is good, this way we'll know when something breaks due to a change in another library

arctic/store/_pandas_ndarray_store.py
TomTaylorLondon: In the most recent numpy they broke the link between numpy ints and python ints (they no longer subclass the python type). This confuses the mongodb driver..

tests/unit/date/test_datetime_to_ms_roundtrip.py
bmoscon: https://github.com/dateutil/dateutil/issues/158

tests/unit/store/test_pickle_store.py
bmoscon: no the issue is bytes vs string in Python3. The file you have was encoded by LZ4 in python2 that returned a string encoding.

tests/unit/tickstore/test_toplevel.py
bmoscon: If we can upgrade mock most of them will work. Most the errors I was getting show they were resolved in much later versions than what we use. If we can do that, we can also support python 3.4

arctic/store/_version_store_utils.py
TomTaylorLondon: hashlib takes byte/char arrays - you have to convert the symbol (in P3 unicode str) into ascii to encode now. I can't see why it wouldn't return the same value as all strings in P2.7 have to be ascii.

tests/unit/store/test_pickle_store.py
jamesblackburn: So we don't support reading these pickles anymore?

arctic/store/_version_store_utils.py
jamesblackburn: Not sure - but presumably string encoding will be doing some additional work in this path, which would be non ideal.

arctic/store/_version_store_utils.py
TomTaylorLondon: added

tests/unit/store/test_pickle_store.py
TomTaylorLondon: I'm not even sure if you can read pickles between python 2 and 3...

arctic/store/_version_store_utils.py
TomTaylorLondon: In python 2.X the default string encoding is ascii - therefore the conversion to ascii is a noop. 

In Python 3.X we have to encode to ascii to pass it into shalib. This means we don't support unicode string names, but we don't in 2.X already.

tests/unit/store/test_pickle_store.py
femtotrader: requests-cache is facing same issue.

Response stored in DB using Python 2 can't be retrieved using Python 3 (and vice versa).

bmoscon:
@jamesblackburn @TomTaylorLondon I fixed the mock tests so everything works with mock 1.3.0 and I removed the xfails I added. Ok to merge?

tests/unit/scripts/test_initialize_library.py
```diff
@@ -9,10 +9,10 @@
 def test_init_library():
     # Create the user agains the current mongo database
     with patch('pymongo.MongoClient') as MongoClient, \
-         patch('arctic.scripts.arctic_init_library.logger', autospec=True) as logger, \
-         patch('arctic.scripts.arctic_init_library.Arctic', autospec=True) as Arctic, \
-         patch('arctic.scripts.arctic_init_library.get_mongodb_uri', autospec=True) as get_mongodb_uri, \
-         patch('arctic.scripts.arctic_init_library.do_db_auth', autospec=True) as do_db_auth:
+         patch('arctic.scripts.arctic_init_library.logger') as logger, \
```
jamesblackburn: Why no more `autospec` ?

Without autospec these tests don't really test antyhing - it's possible for the API to change and these wiring tests will happily pass...

jamesblackburn:
This is an epic piece of work @bmoscon 

Well done and +1!

tests/unit/scripts/test_initialize_library.py
bmoscon: When Mock went to 1.1.0 they changed how autospecing works, which now affects methods wrapped in decorators. I'm putting these back to see if they work, if not it will just take some more massaging to get them working. Specin'g seems to work, but autospecing does not (when decorated methods come into play). see http://stackoverflow.com/questions/34800832/issue-mocking-with-a-decorated-method-call-in-python-2-7

bmoscon:
ok @jamesblackburn I fixed the autospec issue. Anything else? :)

jamesblackburn:
All good from me! Presumably after merge we'll switch the build from circleci to travis?

bmoscon:
@jamesblackburn We can if you want. I do think there is some value in having Circle CI, since the versions it uses are pinned to what we use internally. It would be a pain if we merge something in that works great with the latest version of numpy but breaks with what we use internally.

tests/unit/date/test_datetime_to_ms_roundtrip.py
pganssle: FYI, I tracked this down and it is an issue with Python that should be OK for Python 3.4.4+.

PR #94, opened 2016-01-24
new unit tests

bmoscon:
Addresses gaps in code coverage

### DIFF
--- a/tests/unit/date/test_util.py
+++ b/tests/unit/date/test_util.py
@@ -86,3 +86,23 @@ def test_to_dt_dt_tz():
 
 def test_to_dt_dt_tz_default():
     assert to_dt(dt(1970, 1, 1, tzinfo=mktz('UTC')), mktz('Europe/London')) == dt(1970, 1, 1, tzinfo=mktz('UTC'))
+
+
+def test_daterange_raises():
+    with pytest.raises(ValueError):
+        assert(DateRange(dt(2013, 1, 1), dt(2000, 1, 1)))
+
+
+def test_daterange_eq():
+    dr = DateRange(dt(2013, 1, 1))
+    assert((dr == None) == False)
+    assert(dr == dr)
+
+
+def test_daterange_lt():
+    dr = DateRange(dt(2013, 1, 1))
+    dr2 = DateRange(2001, 1, 1)
+
+    assert(dr2 < dr)
+    dr.start = None
+    assert((dr2 < dr) == False)
--- a/tests/unit/date/test_util.py
+++ b/tests/unit/date/test_util.py
@@ -101,7 +101,7 @@ def test_daterange_eq():
 
 def test_daterange_lt():
     dr = DateRange(dt(2013, 1, 1))
-    dr2 = DateRange(2001, 1, 1)
+    dr2 = DateRange(dt(2001, 1, 1))
 
     assert(dr2 < dr)
     dr.start = None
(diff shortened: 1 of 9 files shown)


bmoscon:
@jamesblackburn want to take a look? Just new unit tests

tests/unit/date/test_util.py
```diff
@@ -86,3 +86,23 @@ def test_to_dt_dt_tz():
...
+
+def test_daterange_eq():
+    dr = DateRange(dt(2013, 1, 1))
+    assert((dr == None) == False)
+    assert(dr == dr)
+
+
+def test_daterange_lt():
+    dr = DateRange(dt(2013, 1, 1))
+    dr2 = DateRange(2001, 1, 1)
```
jamesblackburn: Did you mean to construct it like this?

tests/unit/store/test_version_store_audit.py
```diff
@@ -3,11 +3,17 @@
...
 
-from arctic.store.audit import ArcticTransaction
+from arctic.store.audit import ArcticTransaction, DataChange
 from arctic.store.version_store import VersionedItem, VersionStore
 from arctic.exceptions import ConcurrentModificationException, NoDataFoundException
 
 
+def test_data_change():
+    d = DataChange("a", "b")
+    assert(d.date_range == "a")
```
jamesblackburn: `setinel....` can be used when you don't care about the type of the value

jamesblackburn:
+1

tests/unit/date/test_util.py
bmoscon: ah, good catch! I'll fix it

tests/unit/store/test_version_store_audit.py
bmoscon: will change

tests/unit/store/test_version_store_audit.py
```diff
@@ -3,11 +3,17 @@
...
-from arctic.store.audit import ArcticTransaction
+from arctic.store.audit import ArcticTransaction, DataChange
 from arctic.store.version_store import VersionedItem, VersionStore
 from arctic.exceptions import ConcurrentModificationException, NoDataFoundException
 
 
+def test_data_change():
+    d = DataChange(sentinel, sentinel)
+    assert(d.date_range == sentinel)
+    assert(d.new_data == sentinel)
```
AdrianTeng: From your previous commit, I think this should be something like:

```
d = DateChange(sentinel.spam, sentinel.egg)
assert(d.date_range == sentinel.spam)
assert(d.new_data == sentinel.egg)
```

tests/unit/date/test_util.py
```diff
@@ -86,3 +86,23 @@ def test_to_dt_dt_tz():
...
+
+
+def test_daterange_raises():
+    with pytest.raises(ValueError):
+        assert(DateRange(dt(2013, 1, 1), dt(2000, 1, 1)))
+
+
+def test_daterange_eq():
+    dr = DateRange(dt(2013, 1, 1))
+    assert((dr == None) == False)
```
AdrianTeng: `assert` is a statement (in both python 2 & 3),  and since the statement inside evaluates to a single `True` or `False`, the outer bracket pair is redundent.

(applies to all the `assert` in this PR)

AdrianTeng:
...And I just realised you've already merged this #facepalm

bmoscon:
@AdrianTeng feel free to push any changes you want. I did know that assert is a statement and does not need the ()

Issue #95, opened 2016-01-26
toplevel read raises NoDataFoundException if any lower level library raises this error


reasto:
Only raise the error from the top level read if no data is returned from all the low level reads put together.

bmoscon:
@reasto i assume this is fixed now that the pull request has been merged in? Can we close this?

reasto:
@bmoscon all done.

PR #97, opened 2016-01-28
Top level read fix for #95

reasto:
A read across two or more lower level tickstore libraries where one of them contained no data would raise a NoDataFoundException even though the others had returned data. This change fixes the bug and returns data.
@bmoscon @jamesblackburn

### DIFF
--- a/arctic/tickstore/toplevel.py
+++ b/arctic/tickstore/toplevel.py
@@ -100,8 +100,16 @@ overlapping libraries: {}""".format(library_name, [l.library for l in library_me
 
     def read(self, symbol, date_range, columns=['BID', 'ASK', 'TRDPRC_1', 'BIDSIZE', 'ASKSIZE', 'TRDVOL_1'], include_images=False):
         libraries = self._get_libraries(date_range)
-        dfs = [l.library.read(symbol, l.date_range.intersection(date_range), columns,
-                              include_images=include_images) for l in libraries]
+        dfs = []
+        for l in libraries:
+            try:
+                df = l.library.read(symbol, l.date_range.intersection(date_range), columns,
+                                    include_images=include_images)
+                dfs.append(df)
+            except NoDataFoundException, e:
+                continue
+        if len(dfs) == 0:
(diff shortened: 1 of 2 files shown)


tests/integration/tickstore/test_toplevel.py
```diff
@@ -88,6 +88,28 @@ def test_should_return_data_when_date_range_spans_libraries(toplevel_tickstore,
     assert_frame_equal(expected_df, res.tz_convert(mktz('Europe/London')))
 
 
+def test_should_return_data_when_date_range_spans_libraries_even_if_one_returns_nothing(toplevel_tickstore, arctic):
+    arctic.initialize_library('FEED_2010.LEVEL1', tickstore.TICK_STORE_TYPE)
+    arctic.initialize_library('FEED_2011.LEVEL1', tickstore.TICK_STORE_TYPE)
+    tickstore_2010 = arctic['FEED_2010.LEVEL1']
+    tickstore_2011 = arctic['FEED_2011.LEVEL1']
+    toplevel_tickstore._collection.insert_one({'start': dt(2010, 1, 1),
```
jamesblackburn: Can we not use `TopLevelTickStore.add` here rather than crafting a mapping by hand?

bmoscon:
also, the builds will have to pass for this to be merged.

arctic/tickstore/toplevel.py
```diff
@@ -100,8 +100,16 @@ def add(self, date_range, library_name):
...
         libraries = self._get_libraries(date_range)
-        dfs = [l.library.read(symbol, l.date_range.intersection(date_range), columns,
-                              include_images=include_images) for l in libraries]
+        dfs = []
+        for l in libraries:
+            try:
+                df = l.library.read(symbol, l.date_range.intersection(date_range), columns,
+                                    include_images=include_images)
+                dfs.append(df)
+            except NoDataFoundException, e:
```
bmoscon: thats the python2 style, please update so this works in python3

tests/integration/tickstore/test_toplevel.py
reasto: Updating now

PR #102, opened 2016-02-01
Fix pandas write

reasto:
Adding tests to detect doc['e'] being set incorrectly. Changing code in tickstore.py to fix the bug.

It seems that assert_frame_equal occasionally raises in error, I don't think my changes have caused this.


bmoscon:
@reasto yes, there is an intermittent issue. I've already opened an issue for this. all you can do is re-run the build (which I have done)

PR #103, opened 2016-02-02
Tickstore write initial images

reasto:
The tickstore code has supported initial images with each document for some time but the write() call could not write the data. This PR allows for an initial_image field. We also change the init to enable the chunk_size to be set on creation. A number of methods get changed to staticmethods.

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -69,6 +69,7 @@ IMAGE = 'i'
 COLUMNS = 'cs'
 DATA = 'd'
 DTYPE = 't'
+IMAGE_TIME = 't'
 ROWMASK = 'm'
 
 COUNT = 'c'
@@ -344,7 +345,7 @@ class TickStore(object):
 
     def _prepend_image(self, document, im, rtn_length, column_dtypes, column_set, columns):
         image = im[IMAGE]
-        first_dt = im[DTYPE]
+        first_dt = im[IMAGE_TIME]
         if not first_dt.tzinfo:
             first_dt = first_dt.replace(tzinfo=mktz('UTC'))
         document[INDEX] = np.insert(document[INDEX], 0, np.uint64(datetime_to_ms(first_dt)))
@@ -578,7 +579,7 @@ class TickStore(object):
                 start = to_dt(df.index[0].to_datetime())
             image_start = initial_image.get('index', start)
             image = {k: v for k, v in initial_image.items() if k != 'index'}
-            rtn[IMAGE_DOC] = {DTYPE: image_start, START: 0, IMAGE: initial_image}
+            rtn[IMAGE_DOC] = {IMAGE_TIME: image_start, IMAGE: initial_image}
             final_image = TickStore._pandas_compute_final_image(df, initial_image, end)
         else:
             start = to_dt(df.index[0].to_datetime())
@@ -639,7 +640,7 @@ class TickStore(object):
         if initial_image:
             image_start = initial_image.get('index', start)
             start = min(start, image_start)
-            rtn[IMAGE_DOC] = {DTYPE: image_start, START: 0, IMAGE: initial_image}
+            rtn[IMAGE_DOC] = {IMAGE_TIME: image_start, IMAGE: initial_image}
         rtn[END] = end
         rtn[START] =  start
         rtn[INDEX] = Binary(lz4.compressHC(np.concatenate(([data['index'][0]], np.diff(data['index']))).tostring()))
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -561,7 +561,7 @@ class TickStore(object):
         # Compute the final image with forward fill of df applied to the image
         final_image = copy.copy(image)
         last_values = df.ffill().tail(1).to_dict()
-        last_dict = {i: a.values()[0] for i, a in last_values.items()}
+        last_dict = {i: list(a.values())[0] for i, a in last_values.items()}
         final_image.update(last_dict)
         final_image['index'] = end
         return final_image
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -468,10 +468,10 @@ class TickStore(object):
         ----------
         symbol : `str`
             symbol name for the item
-        data : list of dicts or a panda.DataFrame
+        data : list of dicts or a pandas.DataFrame
             List of ticks to store to the tick-store.
             if a list of dicts, each dict must contain a 'index' datetime
-            if a panda.DataFrame the index must be a Timestamp that can be converted to a datetime
+            if a pandas.DataFrame the index must be a Timestamp that can be converted to a datetime
         initial_image : dict
             Dict of the initial image at the start of the document. If this contains a 'index' entry it is
             assumed to be the time of the timestamp of the index
@@ -500,6 +500,7 @@ class TickStore(object):
         mongo_retry(self._collection.insert_many)(buckets)
         t = (dt.now() - start).total_seconds()
         ticks = len(buckets) * self._chunk_size
+        logger.debug("%d buckets in %s: approx %s ticks/sec" % (len(buckets), t, int(ticks / t)))
 
     def _pandas_to_buckets(self, x, symbol, initial_image):
         rtn = []
@@ -638,7 +639,7 @@ class TickStore(object):
         if initial_image:
             image_start = initial_image.get('index', start)
             start = min(start, image_start)
-            rtn[IMAGE_DOC] = {DTYPE: image_start, START: 0, IMAGE: final_image}
+            rtn[IMAGE_DOC] = {DTYPE: image_start, START: 0, IMAGE: initial_image}
         rtn[END] = end
         rtn[START] =  start
         rtn[INDEX] = Binary(lz4.compressHC(np.concatenate(([data['index'][0]], np.diff(data['index']))).tostring()))
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -10,6 +10,7 @@ import pandas as pd
 from pandas.core.frame import _arrays_to_mgr
 import pymongo
 from pymongo.errors import OperationFailure
+import copy
 
 from ..date import DateRange, to_pandas_closed_closed, mktz, datetime_to_ms, CLOSED_CLOSED, to_dt
 from ..decorators import mongo_retry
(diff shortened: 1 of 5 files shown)


tests/integration/tickstore/test_ts_write.py
```diff
@@ -79,28 +79,3 @@ def test_ts_write_pandas(tickstore_lib):
 
     read = tickstore_lib.read('SYM', columns=None)
     assert_frame_equal(read, data, check_names=False)
-
```
reasto: Moved to unit tests as the methods are now statismethods

arctic/tickstore/tickstore.py
```diff
@@ -458,8 +468,13 @@ def write(self, symbol, data):
         ----------
         symbol : `str`
             symbol name for the item
-        data : list of dicts
+        data : list of dicts or a panda.DataFrame
```
jamesblackburn: minor: `pandas.DataFrame`

arctic/tickstore/tickstore.py
```diff
@@ -475,38 +490,41 @@ def write(self, symbol, data):
...
-            buckets = self._to_buckets(data, symbol)
+            buckets = self._to_buckets(data, symbol, initial_image)
         self._write(buckets)
 
     def _write(self, buckets):
         start = dt.now()
         mongo_retry(self._collection.insert_many)(buckets)
         t = (dt.now() - start).total_seconds()
-        ticks = len(buckets) * self.chunk_size
-        print("%d buckets in %s: approx %s ticks/sec" % (len(buckets), t, int(ticks / t)))
```
jamesblackburn: Agree this might just be noise - but maybe having it at `logger.debug`?

arctic/tickstore/tickstore.py
```diff
@@ -538,42 +556,68 @@ def _ensure_supported_dtypes(self, array):
...
+        # Compute the final image with forward fill of df applied to the image
+        final_image = copy.copy(image)
+        last_values = df.ffill().tail(1).to_dict()
+        last_dict = {i: list(a.values())[0] for i, a in last_values.items()}
+        final_image.update(last_dict)
+        final_image['index'] = end
+        return final_image
+
+    @staticmethod
+    def _pandas_to_bucket(df, symbol, initial_image):
```
jamesblackburn: Is it useful to make this `static`? 

This is the main method that constructs a chunk document so making it static makes it harder for anyone to extend `TickStore` and override the chunk writing implementation

arctic/tickstore/tickstore.py
bmoscon: I think we should keep it and change it from a print to a debug

arctic/tickstore/tickstore.py
```diff
@@ -538,42 +556,68 @@ def _ensure_supported_dtypes(self, array):
...
+        rtn = {SYMBOL: symbol, VERSION: CHUNK_VERSION_NUMBER, COLUMNS: {}, COUNT: len(df)}
         end = to_dt(df.index[-1].to_datetime())
-        rtn = {START: start, END: end, SYMBOL: symbol}
-        rtn[VERSION] = CHUNK_VERSION_NUMBER
-        rtn[COUNT] = len(df)
-        rtn[COLUMNS] = {}
+        if initial_image :
+            if 'index' in initial_image:
+                start = min(to_dt(df.index[0].to_datetime()), initial_image['index'])
```
jamesblackburn: When would we need the `min`? The image should never be timestamped after the first tick, should it?

arctic/tickstore/tickstore.py
```diff
@@ -538,42 +556,68 @@ def _ensure_supported_dtypes(self, array):
...
+            else:
+                start = to_dt(df.index[0].to_datetime())
+            image_start = initial_image.get('index', start)
+            image = {k: v for k, v in initial_image.items() if k != 'index'}
+            rtn[IMAGE_DOC] = {DTYPE: image_start, START: 0, IMAGE: initial_image}
```
jamesblackburn: Is this the wrong way round? `START: 0` ; `DTYPE: image_start`?

arctic/tickstore/tickstore.py
jamesblackburn: Oh I see the image.start is based on the end from the previous chunk.  Presumably `image_start == start` always?

arctic/tickstore/tickstore.py
```diff
@@ -583,21 +627,22 @@ def _to_bucket(self, ticks, symbol):
...
+                v = TickStore._ensure_supported_dtypes(v)
                 rtn[COLUMNS][k] = {DATA: Binary(lz4.compressHC(v.tostring())),
-                                   DTYPE: self._str_dtype(v.dtype),
+                                   DTYPE: TickStore._str_dtype(v.dtype),
                                    ROWMASK: rowmask[k]}
 
+        if initial_image:
+            image_start = initial_image.get('index', start)
+            start = min(start, image_start)
+            rtn[IMAGE_DOC] = {DTYPE: image_start, START: 0, IMAGE: final_image}
```
jamesblackburn: Is this the wrong way round? START: 0 ; DTYPE: image_start?

jamesblackburn: Shouldn't `rtn[IMAGE_DOC][IMAGE]` be `initial_image` rather than `final_image`?

tests/unit/tickstore/test_tickstore.py
```diff
@@ -43,3 +45,48 @@ def test_strify_tickstore():
...
+    assert IMAGE_DOC not in bucket
+    assert not final_image
+
+
+def test_tickstore_to_bucket_with_image():
+    symbol = 'SYM'
+    initial_image = {'index': dt(2014, 1, 1, 0, 0, tzinfo=mktz()), 'A': 123, 'B': 54.4, 'C': 'DESC'}
+    data = [{'index': dt(2014, 1, 1, 0, 1, tzinfo=mktz()), 'A': 124, 'D': 0},
+            {'index': dt(2014, 1, 1, 0, 2, tzinfo=mktz()), 'A': 125, 'B': 27.2}]
+    bucket, final_image = TickStore._to_bucket(data, symbol, initial_image)
```
jamesblackburn: I think we should be asserting the contents of the `bucket` here. It would be useful to lock down the contents of the `IMAGE` subdocument.

tests/unit/tickstore/test_tickstore.py
```diff
@@ -43,3 +45,48 @@ def test_strify_tickstore():
...
+    assert final_image == {'index': dt(2014, 1, 1, 0, 2, tzinfo=mktz()), 'A': 125, 'B': 27.2, 'C': 'DESC', 'D': 0}
+
+
+def test_tickstore_pandas_to_bucket_image():
+    symbol = 'SYM'
+    initial_image = {'index': dt(2014, 1, 1, 0, 0, tzinfo=mktz()), 'A': 123, 'B': 54.4, 'C': 'DESC'}
+    data = [{'A': 120, 'D': 1}, {'A': 122, 'B': 2.0}, {'A': 3, 'B': 3.0, 'D': 1}]
+    data = pd.DataFrame(data, index=[dt(2014, 1, 2, 0, 0, tzinfo=mktz()), dt(2014, 1, 3, 0, 0, tzinfo=mktz()),
+                                     dt(2014, 1, 4, 0, 0, tzinfo=mktz())])
+    bucket, final_image = TickStore._pandas_to_bucket(data, symbol, initial_image)
```
jamesblackburn: I think we should be asserting the contents of the bucket here. It would be useful to lock down the contents of the IMAGE subdocument.

tests/unit/tickstore/test_tickstore.py
```diff
@@ -43,3 +45,48 @@ def test_strify_tickstore():
...
+    data = [{'index': dt(2014, 1, 1, 0, 1, tzinfo=mktz()), 'A': 124, 'D': 0},
+            {'index': dt(2014, 1, 1, 0, 2, tzinfo=mktz()), 'A': 125, 'B': 27.2}]
+    bucket, final_image = TickStore._to_bucket(data, symbol, initial_image)
+    assert bucket[COUNT] == 2
+    assert bucket[END] == dt(2014, 1, 1, 0, 2, tzinfo=mktz())
+    assert bucket[SYMBOL] == symbol
+    assert bucket[START] == dt(2014, 1, 1, 0, 0, tzinfo=mktz())
+    assert final_image == {'index': dt(2014, 1, 1, 0, 2, tzinfo=mktz()), 'A': 125, 'B': 27.2, 'C': 'DESC', 'D': 0}
+
+
```
jamesblackburn: Would be good to test `_to_buckets` for the pandas and list of dicts case to check that the image is updated and propagated between multiple chunks as they're written, and in particular that the timestamps, and fields of the final chunks image are correct

arctic/tickstore/tickstore.py
reasto: changed

arctic/tickstore/tickstore.py
reasto: Agreed

arctic/tickstore/tickstore.py
reasto: The logic here came from the case where I was loading data from a file that spanned many months. I'd load up just less than the chunk size and write it as a document. Then I'd set the next initial image to be the end of the previous document +1 millisecond, so the image could have a time before the first tick time. The issue then arises, if you have an image before the first tick would you expect to see it in a read(date_range)? So I choose to put the start date at the min of the image and the first tick time.

arctic/tickstore/tickstore.py
reasto: I'd like to leave this one as it is for now. Perhaps we can discuss later.

arctic/tickstore/tickstore.py
reasto: In our Java code, 's' or START is sequence No and 't' or DTYPE is the datetime. So I copied this.. Happy for your call on this one. I think I should just drop the START: 0.

arctic/tickstore/tickstore.py
reasto: Thank you. Test added.

arctic/tickstore/tickstore.py
jamesblackburn: So the read will be consistent with the image stored on the chunk, and include the image if its within the date range. It also looks from the other logic that the start time of the image will be the timestamp of the last tick in the previous chunk.
Sounds reasonable to me.

arctic/tickstore/tickstore.py
jamesblackburn: Ah I see, that's unfortunate. Can we call it `IMAGE_TIME` rather than DTYPE, and drop the sequence number field if it's always 0.

reasto: So we don't have to change existing data:
DTYPE = 't'
IMAGE_TIME = 't'
Will happily drop the sequence no

jamesblackburn: +1

reasto: If the checks pass and you're ok with it, I think I'm ready to merge.

PR #106, opened 2016-03-08
Fix pd.Panel storage.

jvermaak:
Ensure rows with nans are not dropped by passing ``filter_observations=False``
to ``panel.to_frame``.


bmoscon:
looks good to me. Not sure why coveralls is complaining.

mildbyte:
Thanks!

PR #111, opened 2016-03-16
Quick stab at handling Apples default gcc (clang) compiler incompatibility

joshuaostrom:
So perhaps simple update to the ReadMe.MD will suffice.  The default gcc compiler on OSX (clang) doesn't support openmp and won't fly.  This PR not-so-elegantly detects when this compiler is in use and issues a warning

`Apple LLVM detected you many need to install (non clang) gcc`

If gcc-\* is found, it will instead use that.  

Alternatives would be 
1) Do nothing, expect OSX users to not have their gcc pointing to the Apple supplied one.
2) Updated Readme.md with note for Apple devs `CC=gcc-4.2 pip install` etc 

No worries if you trash the PR, if some part of it proves useful, great.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,10 @@
 ## Changelog
 
+### 1.27
+
+  * Bugfix: #187 Compatibility with latest version of pytest-dbfixtures
+  * Feature: #182 Improve ChunkStore read/write performance
+
 ### 1.26 (2016-07-20)
 
   * Bugfix: Faster TickStore querying for multiple symbols simultaneously
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,10 +1,12 @@
 ## Changelog
 
-### 1.27
+### 1.27 (2016-08-05)
 
   * Bugfix: #187 Compatibility with latest version of pytest-dbfixtures
   * Feature: #182 Improve ChunkStore read/write performance
   * Feature: #162 Rename API for ChunkStore
+  * Feature: #186 chunk_range on update
+  * Bugfix: #189 range delete does not update symbol metadata
 
 ### 1.26 (2016-07-20)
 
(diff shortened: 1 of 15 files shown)


AdrianTeng:
I'm neutral on whether we should change `setup.py` xor `README.md`. @jamesblackburn @Wilfred?

Wilfred:
Aha, we're using OpenMP for the `cython.parallel` stuff (I was initially expecting to see openmp pragmas like `#pragma omp parallel` but it's only used through cython).

I certainly think we should do something to help out apple users. Putting a note in readme.md would be a reasonable approach, as would this patch. However, recent clang versions do support openmp. Apple's LLVM will probably have this in the future, so we ought to have a version check too.

@jamesblackburn I presume you're developing this on a mac? How are you handling this?

Wilfred:
In principle, it should be possible to use OpenMP with the system compiler on Apple: http://stackoverflow.com/a/36000632/509706

jamesblackburn:
The way I've built this before on MacOS is using conda and the gcc that comes with it. I hacked the cython link line to add `-static` to statically link against the openmp .a. The resulting .so then runs, passes the tests and appears to be relocatable.

bmoscon:
not sure what happened when you updated your fork, but now its near impossible to tell the changes you made vs the changes made since you forked . . .

joshuaostrom:
@bmoscon sorry about that, had updated the fork with the latest from arctic's master.  Should be all set now.  (Been tracking down some oddities I'm only seeing when running arctic from within a docker container).

bmoscon:
@joshuaostrom you'll probably need to update your branch/fork to get the build to pass. I'll test it out this weekend and try building a wheel.

bmoscon:
macOS now supported, so closing this.

Issue #113, opened 2016-03-23
unlabeled dataframes and series changed when read back from arctic


bmoscon:
Here is a test case (add to tests/integration/test_pandas_store)

````
df = DataFrame(data=[1, 2, 3],
                   index=Index(data=[dt(2016, 01, 01),
                                     dt(2016, 01, 02),
                                     dt(2016, 01, 03)],
                               name='date'))
    library.write('test', df)
    ret = library.read('test').data
    assert_frame_equal(ret, df)
````

basically the to/from records functionality will hard code in some colum/series names if none are present. Series get named 'values', and columns become u'0' to u'N'

Series default name is None, and dataframe column names are 0 -> N (integer, not string). Not sure if this is intentional or not, or if its even something to care about (most series and columns in dataframes will be named). 

From the above example here are the values of df.columns and ret.columns, respectively:

````
Int64Index([0], dtype='int64')
Index([u'0'], dtype='object')
````

note 0 goes from int64 to string object.

bmoscon:
@jamesblackburn

richardbounds:
Pretty sure this was intentional, although I can't remember exactly why. At one point I think we stored the column names in the dtype, so it may be something to do with the type checking/promotion code in _ndarry_store. I wonder whether @jjbmatthews remembers?

bmoscon:
@richardbounds we do store the names IF they are there. If you provide a series name or a column name it works perfectly. Its only if they are absent that there is an issue.

bmoscon:
I have done some digging. The dataframe converter maps all the column names through the str function, so any column name that is a non string (i.e. the default column names) will be converted to a string. 

The converter for Series just uses the name 'values' if one is not supplied.

bmoscon:
resolved by logging when we change column/series/index names

PR #114, opened 2016-03-23
Fix column naming issue #113

bmoscon:
@TomTaylorLondon indicated this has been an issue for him in the past, so I've created a fix for #113 

Because the dtype operations in the record conversions require a string for the record name, I opted to use a "magic number" (string in this case) to indicate we've converted to a string from an integer (or a None). I'm open to other methods to fix this.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### 1.22 (2016-03-23)
+
+  * Bugfix: #113 Fix dataframe default column and series default name issues
+
 ### 1.21 (2016-03-08)
 
   * Bugfix: #106 Fix Pandas Panel storage for panels with different dimensions
--- a/arctic/date/_mktz.py
+++ b/arctic/date/_mktz.py
@@ -4,7 +4,7 @@ import dateutil
 import tzlocal
 
 DEFAULT_TIME_ZONE_NAME = tzlocal.get_localzone().zone  # 'Europe/London'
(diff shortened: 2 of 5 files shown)


bmoscon:
@TomTaylorLondon 
@jamesblackburn 
@richardbounds

richardbounds:
It isn't a big change, and it looks OK but I'm not convinced it is necessary - I don't think it is unreasonable to insist on naming the columns - we don't necessarily have to persist the entire universe of possible dataframes perfectly.

bmoscon:
@richardbounds im inclined to agree with you. @TomTaylorLondon mentioned it had been an issue for him in the past and asked me to fix it, so I'll let him argue the merits :)

TomTaylorLondon:
@bmoscon @richardbounds I think it's useful to fix this - it's not a corner case to have un-labeled data-frames. Could you also handle the case when the index isn't named:

https://github.com/manahl/arctic/blob/master/arctic/store/_pandas_ndarray_store.py#L57

richardbounds:
How is it not a corner case? When is it a good idea to have unlabelled dataframes?

TomTaylorLondon:
@richardbounds I think it would be good to fix - it's not critical but we have plenty of tests, nodes in graphs etc. where it would have been helpful. Do you think strongly enough about this to not merge the request?

bmoscon:
@richardbounds per a conversation with @TomTaylorLondon I'm going to scrap this and just emit a long line when we convert any names. That ok with you?

richardbounds:
Fine by me.

Issue #118, opened 2016-03-28
updated python-dateutil breaks unit tests and arctic/date


bmoscon:
newest version of dateutil brings in a lot of changes in its parser, which breaks:

https://github.com/manahl/arctic/blob/master/arctic/date/_parse.py

from my initial inspection it seems that an agnostic parser is no longer possible. For example:

parse("20111231") will break everytime, because of the check with dayfirst=True:

````
arctic/date/_parse.py:7: in parse
    == _parse(string, dayfirst=True, **kwargs)):
../../../virtualenv/python2.7.9/lib/python2.7/site-packages/dateutil/parser.py:1164: in parse
    return DEFAULTPARSER.parse(timestr, **kwargs)
../../../virtualenv/python2.7.9/lib/python2.7/site-packages/dateutil/parser.py:577: in parse
    ret = default.replace(**repl)
E   ValueError: month must be in 1..12

````

I'm not sure what the past behavior was, but currently if it cannot parse a date based on the given information it now raises an error

(see these logs for more info: https://travis-ci.org/manahl/arctic/jobs/119019881 )

bmoscon:
after looking at this a bit more, it looks unclear how to handle this:

1. enforce YYYYMMDD format. remove attempts to distinguish between ambiguous dates

2. Change the unit tests. Honestly I'm not sure how they ever worked. 20111231 with dayfirst = True? How did that ever work? Either way, ambiguous dates that once worked, now no longer work. 

@jamesblackburn

bmoscon:
i think 2 and 3 would have to go together if you wanted to try and keep the same functionality. 

consider this date:
20110102

just calling parsed gives 2011-01-02
parsed with yearfirst gives 2011-01-02
parsed with dayfirst gives 2011-02-01

bmoscon:
Eventually I found the PR that caused this:
https://github.com/dateutil/dateutil/pull/229

Looks like there was a bug before that caused all dates to default to YYYYMMDD when separators were missing.

bmoscon:
resolved

Issue #119, opened 2016-03-29
Numpy 1.11.0 datetime parsing is not timezone aware


bmoscon:
see this commit:

https://github.com/numpy/numpy/commit/426114879da49bf9a586b1991dcaf38ce594c4b6

note this causes several unit tests to fail, and likely will affect anything that tries to use a timezone with numpy

bmoscon:
resolved

PR #120, opened 2016-03-29
Fixes to support latest versions of dateutil and Numpy

bmoscon:
Take a look @jamesblackburn 

this resolves issues #118 and #119


bmoscon:
chatted with James about this before merging.

PR #122, opened 2016-03-30
Support for new chunkable Version Store type

bmoscon:
New store type allows user to specify the chunking method for the dataframe/series (Daily, Monthly, Yearly). Requires a datetime-indexed dataframe/series

### DIFF
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -482,7 +482,11 @@ class NdarrayStore(object):
         version['sha'] = self.checksum(item)
 
         if previous_version:
-            return
+            if version['dtype'] == str(dtype) \
+                    and version['chunk_size'] == previous_version['chunk_size'] \
+                    and 'sha' in previous_version \
+                    and self.checksum(item[:previous_version['up_to']]) == previous_version['sha']:
+                raise Exception("Append on write is handled by PandasStore")
 
         version['base_sha'] = version['sha']
         self._do_chunked_write(collection, version, symbol, records, ranges, previous_version)
@@ -539,7 +543,7 @@ class NdarrayStore(object):
         version['type'] = self.TYPE
         version['up_to'] = len(item)
         version['sha'] = self.checksum(item)
-        
+
         if previous_version:
             if version['dtype'] == str(dtype) \
                     and 'sha' in previous_version \
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -175,7 +175,7 @@ class NdarrayStore(object):
 
         start = date_range[0].strftime('%Y-%m-%d')
         end = date_range[-1].strftime('%Y-%m-%d')
-        
+
         spec = {'symbol': symbol,
                 'parent': version.get('base_version_id', version['_id']),
                 'segment': {'$lte': end, '$gte': start}
@@ -187,9 +187,10 @@ class NdarrayStore(object):
             try:
                 segments.append(decompress(x['data']) if x['compressed'] else x['data'])
             except Exception:
-                dump_bad_documents(x, collection.find_one({'_id': x['_id']}),
-                                      collection.find_one({'_id': x['_id']}),
-                                      collection.find_one({'_id': x['_id']}))
+                dump_bad_documents(x,
+                                   collection.find_one({'_id': x['_id']}),
+                                   collection.find_one({'_id': x['_id']}),
+                                   collection.find_one({'_id': x['_id']}))
                 raise
         data = b''.join(segments)
 
@@ -211,8 +212,8 @@ class NdarrayStore(object):
 
     def _do_read(self, collection, version, symbol, index_range=None):
         '''
-        index_range is a 2-tuple of integers - a [from, to) range of segments to be read. 
-            Either from or to can be None, indicating no bound. 
+        index_range is a 2-tuple of integers - a [from, to) range of segments to be read.
+            Either from or to can be None, indicating no bound.
         '''
         from_index = index_range[0] if index_range else None
         to_index = version['up_to']
@@ -268,7 +269,7 @@ class NdarrayStore(object):
 
         if not dtype:
             dtype = item.dtype
-        
+
         if previous_version['up_to'] == 0:
             dtype = dtype
         elif len(item) == 0:
@@ -340,7 +341,7 @@ class NdarrayStore(object):
                     '''If we get a duplicate key error here, this segment has the same symbol/parent/segment
                        as another chunk, but a different sha. This means that we have 'forked' history.
                        If we concat_and_rewrite here, new chunks will have a different parent id (the _id of this version doc)
-                       ...so we can safely write them. 
+                       ...so we can safely write them.
                        '''
                     self._concat_and_rewrite(collection, version, symbol, item, previous_version)
                     return
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -3,7 +3,6 @@ import hashlib
 
 from bson.binary import Binary
 import numpy as np
-import pandas as pd
 import pymongo
 from pymongo.errors import OperationFailure, DuplicateKeyError
 
@@ -178,9 +177,7 @@ class NdarrayStore(object):
                 }
 
         if date_range is not None:
-            start = date_range[0].strftime('%Y-%m-%d')
-            end = date_range[-1].strftime('%Y-%m-%d')
-            spec['segment'] = {'$lte': end, '$gte': start}
+            spec['segment'] = date_range.mongo_query()
 
         segments = []
         i = -1
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -414,8 +414,7 @@ class NdarrayStore(object):
     def chunked_write(self, arctic_lib, version, symbol, records, ranges, previous_version, dtype):
         collection = arctic_lib.get_top_level_collection()
 
-        item = np.array(records).flatten()
-
+        item = np.array([r for record in records for r in record]).flatten()
         if item.dtype.hasobject:
             raise UnhandledDtypeException()
 
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -11,7 +11,6 @@ from ..exceptions import UnhandledDtypeException
 from ._version_store_utils import checksum
 
 from .._compression import compress_array, decompress
-from ..exceptions import ConcurrentModificationException
 from six.moves import xrange
 
 
@@ -55,7 +54,6 @@ class NdarrayStore(object):
       u'up_to': 10,  # no. of rows included in the data for this version
       u'append_count': 0,
       u'append_size': 0,
-      u'base_sha': Binary('........', 0),
       u'dtype': u'float64',
       u'dtype_metadata': {},
       u'segment_count': 1, #only 1 segment included in this version
@@ -70,7 +68,6 @@ class NdarrayStore(object):
       u'up_to': 20, # no. of rows included in the data for this version
       u'append_count': 1, # 1 append operation so far
       u'append_size': 80, # 80 bytes appended
-      u'base_sha': Binary('.........', 0), # equal to sha for version 1
       u'base_version_id': ObjectId('55fa9a7781f12654382e58b8'), # _id of version 1
       u'dtype': u'float64',
       u'dtype_metadata': {},
@@ -145,9 +142,6 @@ class NdarrayStore(object):
         """
         from_index = None
         if from_version:
-            if version['base_sha'] != from_version['base_sha']:
-                #give up - the data has been overwritten, so we can't tail this
-                raise ConcurrentModificationException("Concurrent modification - data has been overwritten")
             from_index = from_version['up_to']
         return from_index, None
 
@@ -291,7 +285,6 @@ class NdarrayStore(object):
     def _do_chunked_append(self, collection, version, symbol, records, ranges, item, previous_version):
 
         data = item.tostring()
-        version['base_sha'] = previous_version['base_sha']
         version['up_to'] = previous_version['up_to'] + len(item)
         if len(item) > 0:
             version['segment_count'] = previous_version['segment_count'] + len(records)
@@ -355,7 +348,6 @@ class NdarrayStore(object):
             item = np.concatenate([old_arr, item])
             version['up_to'] = len(item)
             version['sha'] = self.checksum(item)
-            version['base_sha'] = version['sha']
             self._do_write(collection, version, symbol, item, previous_version)
         else:
             version['dtype'] = previous_version['dtype']
@@ -366,7 +358,6 @@ class NdarrayStore(object):
     def _do_append(self, collection, version, symbol, item, previous_version):
 
         data = item.tostring()
-        version['base_sha'] = previous_version['base_sha']
         version['up_to'] = previous_version['up_to'] + len(item)
         if len(item) > 0:
             version['segment_count'] = previous_version['segment_count'] + 1
@@ -479,16 +470,7 @@ class NdarrayStore(object):
         version['dtype_metadata'] = dict(dtype.metadata or {})
         version['type'] = self.TYPE
         version['up_to'] = len(item)
-        version['sha'] = self.checksum(item)
 
-        if previous_version:
-            if version['dtype'] == str(dtype) \
-                    and version['chunk_size'] == previous_version['chunk_size'] \
-                    and 'sha' in previous_version \
-                    and self.checksum(item[:previous_version['up_to']]) == previous_version['sha']:
-                raise Exception("Append on write is handled by PandasStore")
-
-        version['base_sha'] = version['sha']
         self._do_chunked_write(collection, version, symbol, records, ranges, previous_version)
 
     def _do_chunked_write(self, collection, version, symbol, records, ranges, previous_version):
@@ -552,7 +534,6 @@ class NdarrayStore(object):
                 self._do_append(collection, version, symbol, item[previous_version['up_to']:], previous_version)
                 return
 
-        version['base_sha'] = version['sha']
         self._do_write(collection, version, symbol, item, previous_version)
 
     def _do_write(self, collection, version, symbol, item, previous_version, segment_offset=0):
@@ -648,6 +629,7 @@ class NdarrayStore(object):
             end = rng[1]
             orig_start = orig_rng[0]
             if orig_start is None:
+                version['up_to'] += rec_len
                 seg_count += 1
                 seg_len += rec_len
             segment = {'data': Binary(chunk), 'compressed': True}
@@ -669,4 +651,3 @@ class NdarrayStore(object):
             version['segment_count'] += seg_count
             version['append_size'] += seg_len
             version['append_count'] += seg_count
-            version['sha'] = sha
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -129,6 +129,9 @@ class NdarrayStore(object):
     def can_write(self, version, symbol, data, **kwargs):
         return isinstance(data, np.ndarray) and not data.dtype.hasobject
 
+    def can_update(self, version, symbol, data, **kwargs):
+        return False
+
     def _dtype(self, string, metadata=None):
         if metadata is None:
             metadata = {}
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -37,16 +37,39 @@ def _promote_struct_dtypes(dtype1, dtype2):
     return np.dtype([(n, _promote(dtype1.fields[n][0], dtype2.fields.get(n, (None,))[0])) for n in dtype1.names])
 
 
+class NumpySizeChunker(object):
+    TYPE = 'numpysize'
+    def __call__(self, item, **kwargs):
+        sze = int(item.dtype.itemsize * np.prod(item.shape[1:]))
+        length = len(item)
+        chunk_size = int(_CHUNK_SIZE / sze)
+        idxs = xrange(int(np.ceil(float(length) / chunk_size)))
+        for i in idxs:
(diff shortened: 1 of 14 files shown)


bmoscon:
@TomTaylorLondon take a first look for me. May not be 100% done yet, but I think all the base functionality is here. I'm trying to make sure that I cover all the current Atlas use cases.

bmoscon:
@TomTaylorLondon i added the initial support for updating a chunk. let me know what you think. (Still need to add a lot of tests)

bmoscon:
This is basically done now. I still need to add some unit tests and potentially correct any issues those uncover, but otherwise ready for review.

take a look @jamesblackburn and assign anyone else you think might want to take a look

arctic/store/_ndarray_store.py
```diff
@@ -163,6 +157,47 @@ def get_info(self, version):
...
+        if date_range is not None:
+            spec['segment'] = date_range.mongo_query()
+
+        segments = []
+        i = -1
+        for i, x in enumerate(collection.find(spec, sort=[('segment', pymongo.ASCENDING)],)):
+            try:
+                segments.append(decompress(x['data']))
+            except Exception:
+                dump_bad_documents(x,
```
richardbounds: Do we still need dump_bad_documents()?

bmoscon: probably not. It doesnt seem like we need it in the normal read path either, but I wasnt sure, so I just left it in. Do you think we can remove this from both paths @richardbounds ?

jamesblackburn: Yep, this is dead - it was there to debug a network corruption issue we used to experience and was subsequently fixed by infra.

arctic/store/_ndarray_store.py
```diff
@@ -129,9 +126,12 @@ def can_delete(self, version, symbol):
     def can_read(self, version, symbol):
         return version['type'] == self.TYPE
 
-    def can_write(self, version, symbol, data):
+    def can_write(self, version, symbol, data, **kwargs):
         return isinstance(data, np.ndarray) and not data.dtype.hasobject
 
+    def can_update(self, version, symbol, data, **kwargs):
```
jamesblackburn: Can we push this default implementation up into a generic base-class?  Our TimeSeries chunk store (derived from `VariableChunkStore`) will also need changing, so it should probably inherit basic functionality from a generic `ABC`

bmoscon: NdarrayStore is the base class. I can just remove it from here (its covered in the PandasStore class), but that could cause options in some rare situations. Are you saying you'd want an even "baser" class? Something like "VersionedObjectStore" or something like that that everything here and ahl.mongo can inherit from?

jamesblackburn: Yeah, I was thinking of something 'Baser'.  As the classes get bigger it becomes hard to know what the API is with `VersionStore`

arctic/store/_ndarray_store.py
```diff
@@ -163,6 +160,41 @@ def get_info(self, version):
         ret['rows'] = int(version['up_to'])
         return ret
 
+    def chunked_read(self, arctic_lib, version, symbol, date_range, read_preference=None):
```
jamesblackburn: Only called from a derived class - perhaps prefix with `_`?

arctic/store/_ndarray_store.py
```diff
@@ -163,6 +160,41 @@ def get_info(self, version):
...
+
+    def _do_chunked_read(self, collection, version, symbol, date_range):
+        segment_count = version.get('segment_count', None)
+
+        spec = {'symbol': symbol,
+                'parent': version.get('base_version_id', version['_id']),
+                }
+
+        if date_range is not None:
+            spec['segment'] = date_range.mongo_query()
```
jamesblackburn: So `segment` here are dates, whereas elsewhere in this class it's an `int`?

bmoscon: 'segment' is the name that was chosen to identify the ordering of the individual segments. Not sure why. So yes, for the other segments, its an integer, but for this storage type its a date.

arctic/store/_ndarray_store.py
bmoscon: good call.

arctic/store/_ndarray_store.py
bmoscon: ok, will do.

arctic/store/_ndarray_store.py
```diff
@@ -145,9 +143,6 @@ def _index_range(self, version, symbol, from_version=None, **kwargs):
         """
         from_index = None
         if from_version:
-            if version['base_sha'] != from_version['base_sha']:
```
jamesblackburn: Isn't this still a worry?  A `write` fighting an `append` might cause corruption?

arctic/store/_ndarray_store.py
```diff
@@ -163,6 +158,41 @@ def get_info(self, version):
         ret['rows'] = int(version['up_to'])
         return ret
 
+    def _chunked_read(self, arctic_lib, version, symbol, date_range, read_preference=None):
```
jamesblackburn: This new hunk of code is only used by `PandasDateTimeIndexedStore` what's the motivation for pushing it up to `NdarrayStore` ?

jamesblackburn: So looks like this shares the unpacking logic with `_do_read` the difference being the query spec

arctic/store/_ndarray_store.py
jamesblackburn: Is this right?

If a query spans some DateRange: t -> t' 
then the query will be: `{segment: {'$gt': t, '$lt': t'}}`

If there is to segment=`t0` which chunk, which contains data for `t`, `t0` will be less than `t` and won't be picked up by the query, will it?

arctic/store/_ndarray_store.py
```diff
@@ -163,6 +158,41 @@ def get_info(self, version):
...
+        segments = []
+        i = -1
+        for i, x in enumerate(collection.find(spec, sort=[('segment', pymongo.ASCENDING)],)):
+            segments.append(decompress(x['data']))
+
+        data = b''.join(segments)
+
+        # we can retrieve any number of segments due to how we chunk and query the chunks
+        # so anything from 0 to segment_chunks (inclusive) is ok
+        if segment_count is not None and i + 1 > segment_count:
```
jamesblackburn: The size chunked code was trying to assert, in the presence of `append` that the returned number of documents matches the `segment_count` expected.  

This condition looks sufficiently weaker that perhaps we delete - I can't think of a mechanism by which more documents would come back than we expected.

arctic/store/_store.py
```diff
@@ -0,0 +1,12 @@
+class BaseStore(object):
```
richardbounds: shouldn't read(), write() and maybe update() and append() be in here too?

arctic/store/_ndarray_store.py
```diff
@@ -467,9 +597,48 @@ def _segment_index(self, new_data, existing_index, start, new_segments):
...
+            segment['end'] = end
+            sha = checksum(symbol, segment)
+            segment['sha'] = sha
+            if orig_start is None:
+                # new chunk
+                bulk.find({'symbol': symbol, 'sha': sha, 'segment': segment['segment']}
+                          ).upsert().update_one({'$set': segment, '$addToSet': {'parent': version['_id']}})
+            else:
+                bulk.find({'symbol': symbol, 'segment': orig_start}
+                          ).update_one({'$set': segment, '$addToSet': {'parent': version['_id']}})
```
richardbounds: Are we modifying an existing chunk at this point?

arctic/store/_pandas_ndarray_store.py
```diff
@@ -253,8 +262,8 @@ def from_records(self, recarr):
         name = recarr.dtype.names[-1]
         return Series.from_array(recarr[name], index=index, name=name)
 
-    def can_write(self, version, symbol, data):
-        if isinstance(data, Series):
+    def can_write(self, version, symbol, data, **kwargs):
+        if isinstance(data, Series) and 'chunk_size' not in version:
```
richardbounds: Why is `chunk_size` here? It doesn't seem right that PandasSeriesStore should know about the implementation of the chunked store.

arctic/store/version_store.py
```diff
@@ -517,6 +522,31 @@ def append(self, symbol, data, metadata=None, prune_previous_version=True, upser
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=version['version'],
                              metadata=version.pop('metadata', None), data=None)
 
+    @mongo_retry
+    def update(self, symbol, data, **kwargs):
```
richardbounds: I don't really think we should support this :-(

bmoscon: I can always make a whole new store type instead of trying to shoehorn this into the version store. Is that preferable?

arctic/store/_pandas_ndarray_store.py
bmoscon: To resolve ambiguity between the handlers, the only difference between series store and the chunking one is the chunk_size arg. The actual data is the same.

richardbounds: We've had multiple handlers for the same data type in the past - they are checked in the order that they appear in _TYPE_HANDLERS.

arctic/store/version_store.py
```diff
@@ -559,6 +589,9 @@ def write(self, symbol, data, metadata=None, prune_previous_version=True, **kwar
                                                   sort=[('version', pymongo.DESCENDING)],
                                                   )
 
+        if 'chunk_size' in  kwargs:
```
richardbounds: I'd rather not propagate `chunk_size` up into VersionStore.

arctic/store/version_store.py
richardbounds: It might be, but why do you want to update chunk documents, instead of creating a new version with a different chunk(s)?

arctic/store/_ndarray_store.py
jamesblackburn: I tend to agree ideally the semantics of versions and versions under a snapshot should be preserved even with 'update'. The frozen nature of a version of a symbol allows you to have temporal history of changes which is so useful when trying to work out how-and-why data has changed.

arctic/store/_pandas_ndarray_store.py
```diff
@@ -346,3 +355,218 @@ def read(self, arctic_lib, version, symbol, **kwargs):
...
+        dates = df.index.get_level_values('date')
+        start = dates.min()
+        end = dates.max()
+        if isinstance(start, Timestamp):
+            start = start.to_pydatetime()
+        if isinstance(end, Timestamp):
+            end = end.to_pydatetime()
+        return start, end
+
+    def to_records(self, df, chunk_size):
```
jamesblackburn: Would be good to test this

arctic/store/_pandas_ndarray_store.py
```diff
@@ -346,3 +355,218 @@ def read(self, arctic_lib, version, symbol, **kwargs):
...
+        """
+        chunks the dataframe/series by dates
+
+        returns
+        -------
+        A list of tuples - (start date, end date, dataframe/series)
+        """
+        dates = [pd.to_datetime(d) for d in df.index.get_level_values('date').drop_duplicates()]
```
jamesblackburn: Is this expensive?

arctic/store/_pandas_ndarray_store.py
```diff
@@ -346,3 +355,218 @@ def read(self, arctic_lib, version, symbol, **kwargs):
...
+        key_array = [self.get_date_chunk(d, chunk_size) for d in dates]
```
jamesblackburn: Default for `chunk_size` in `write` below seems to be `None`.  In that case doesn't that mean key_array will be a a list of `None`s?

arctic/store/version_store.py
```diff
@@ -288,6 +287,14 @@ def _write_handler(self, version, symbol, data, **kwargs):
             handler = self._bson_handler
         return handler
 
+    def _update_handler(self, version, symbol, data, **kwargs):
+        handler = None
+        for h in _TYPE_HANDLERS:
```
jamesblackburn: I guess for the case of updates, we already have a version document so can we just dispatch directly to the correct handler for the chunks?

arctic/store/version_store.py
```diff
@@ -517,6 +522,31 @@ def append(self, symbol, data, metadata=None, prune_previous_version=True, upser
...
+        version = self._read_metadata(symbol)
+        old_ver = dict(version)
+        handler = self._update_handler(version, symbol, data, **kwargs)
+
+        if handler is None:
+            Exception("Update not implemented for handler %s" % handler)
+
+        mongo_retry(handler.update)(self._arctic_lib, version, symbol, data, **kwargs)
+        mongo_retry(self._versions.delete_one)(old_ver)
+        mongo_retry(self._versions.insert_one)(version)
```
jamesblackburn: Why do you delete and reinsert the version document if nothing has changed?

arctic/store/version_store.py
```diff
@@ -517,6 +522,31 @@ def append(self, symbol, data, metadata=None, prune_previous_version=True, upser
...
+        """
+        version = self._read_metadata(symbol)
+        old_ver = dict(version)
+        handler = self._update_handler(version, symbol, data, **kwargs)
+
+        if handler is None:
+            Exception("Update not implemented for handler %s" % handler)
+
+        mongo_retry(handler.update)(self._arctic_lib, version, symbol, data, **kwargs)
+        mongo_retry(self._versions.delete_one)(old_ver)
```
jamesblackburn: This may or may not work as you expect. If there are nested documents within the version document then they will be byte compared by the mongoserver. The Python dictionary order may well be different to the mongo server order... 
https://jira.mongodb.org/browse/SERVER-5030

arctic/store/version_store.py
```diff
@@ -517,6 +522,31 @@ def append(self, symbol, data, metadata=None, prune_previous_version=True, upser
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=version['version'],
                              metadata=version.pop('metadata', None), data=None)
 
+    @mongo_retry
+    def update(self, symbol, data, **kwargs):
+        """
+        Find an existing chunk of data and update (overwrite) it. This WILL NOT
```
jamesblackburn: It would be good to know from the docstring to detail what update means in this context. Is it a replacement of all data in a particular date-range chunk, or a merge, or a replacement in the date-range bounds contained within `data`

arctic/store/_ndarray_store.py
bmoscon: give me some example dates and I'll add a unit test :)

arctic/store/version_store.py
jamesblackburn: And I guess it depends on the chunk_size of the data as previously written.

arctic/store/_pandas_ndarray_store.py
```diff
@@ -346,3 +355,218 @@ def read(self, arctic_lib, version, symbol, **kwargs):
...
+                can't slice with partial date on multi-index. Support coming in 
+                pandas 0.18.1
+                '''
+                ret = df.xs(slice(date, date), level='date', drop_level=False)
+            else:
+                ret = df[date : date]
+            start, end = self.get_range(ret)
+            yield start, end, ret
+
+    def write(self, arctic_lib, version, symbol, item, previous_version, chunk_size=None):
```
jamesblackburn: Do subsequent writes preserve the chunk_size from the previous write?

arctic/store/_ndarray_store.py
bmoscon: per a conversation with Richard:

As you found, base_sha is used as a consistency check when ‘tailing’ the contents of a time series – we would read only the difference between one version and a subsequent version, assuming that there had only been appends between versions. We removed that functionality, so base_sha can probably be deleted.

arctic/store/_ndarray_store.py
bmoscon: I can of course bring all this code back into the pandas storage class. I'm not even sure why its in the ndarray storage class since the majority of it seems to be mongo/schema specific and not really related to pandas, numpy, etc.

arctic/store/_ndarray_store.py
bmoscon: yes, I can delete it.

arctic/store/_store.py
bmoscon: yes, probably. Read and Write for sure, but I dont know about the others. Not all classes can append (eg Pickle store doesnt seem to have an append).

arctic/store/_ndarray_store.py
bmoscon: If I update a chunk, how do I make the new version doc point to it and not the old one? You'll have two docs with the same segment (i.e. index). The only way i can see to do it would be to add another field to the segment docs (version) and it would have to match a version field in the version doc, so when you read for a given date range, you only read dates in the range that also match the version number of the parent version doc.

arctic/store/_pandas_ndarray_store.py
bmoscon: I can adjust the ordering then and remove this. The code will then rely on them to be in the correct behavior in order to work correctly.

arctic/store/version_store.py
bmoscon: See previous comment about this. I can add a new field to the segment docs that maps to the version number in the version doc.

arctic/store/version_store.py
bmoscon: noted. will change.

arctic/store/_pandas_ndarray_store.py
bmoscon: Do you mean unit tests?

arctic/store/_pandas_ndarray_store.py
bmoscon: compared to what? I'm sure its not cheap, but its how the Atlas code has been doing it. Honestly I spent A LONG time messing with chunking by date, and I am open to any and all suggestions.

arctic/store/_pandas_ndarray_store.py
bmoscon: I thought exceptions would be raised at other points if it were none by now. I could be wrong, ill check and fix if necessary.

arctic/store/version_store.py
bmoscon: I modeled this after append. Isnt the same thing true for appends?

arctic/store/version_store.py
bmoscon: It does change. The version number doesnt, but the number of appended segments can change. If you update with a DF that requires new chunks to be added, it will update the existing ones and append the new ones.

arctic/store/version_store.py
bmoscon: I'm not sure what line you meant this comment to be on. I assume it was not on the delete_one line, or was it? shouldnt it delete based on the key?

arctic/store/version_store.py
bmoscon: it can do a lot of things, so I should update the docstring.

arctic/store/_pandas_ndarray_store.py
bmoscon: yes, you cannot change the chunk size, or should not be able to. Sounds like some new unit tests are in order.

arctic/store/version_store.py
jamesblackburn: Sounds good - looks like historical cruft ;)

arctic/store/version_store.py
jamesblackburn: Ah, OK.  I guess a delete followed by an insert does (briefly) corrupt the symbol - an update here based no the version `_id` might be safer?

arctic/store/_ndarray_store.py
jamesblackburn: All the chunks in arctic (currently) have a parent field that points back up to the version document(s) that refer to the chunk.

This is somewhat backwards and exists as a result of the experiments we did when trying to figure out how to maximise read performance. Querying based on an indexed parent field was much faster than querying a set of chunks, for example. So in this case we would need to write a new version doc, and reference that version in you new chunks + all the existing chunks  (- the chunk you're replacing).  This would be as expensive as a `write()`.

Ideally it would be the other way round: we would write a new version document, and a new chunk, and be able to efficiently query the existing and new chunk using the new version document (I think this might be possible doing something clever with 2d indexes to pull out the latest chunk version across some date range).  This might not be easy without some serious playing...

arctic/store/_ndarray_store.py
jamesblackburn: Would something like:

- Chunked by year
- TS containing: 2015/01/01  ;  2015/06/01 ; 2016/01/01
- Query 2015/05/30 -> 2016/01/02 should return two rows 

work?

arctic/store/_ndarray_store.py
jamesblackburn: Ah, that sounds familiar. On master I was looking at the write (which sets this field anew) and the append which copies this, and was worried that this was protecting a collision between the two method calls.  It's quite an unlikely issue anyhow.

arctic/store/version_store.py
jamesblackburn: Yah, sorry for not being clear.  Doing a mongo query (or delete in this case) given a fully specified document will surprisingly not work in some cases, because subdocument ordering may not be consistent in the subdocument is in a python hash map.

Something like: `delete_one({'_id': old_ver['_id']})`

arctic/store/_pandas_ndarray_store.py
jamesblackburn: I was wondering whether this iterated over every row in pure python and whether this had an impact (at the scale of 1minute - 4Mln row time series).

Sorry perhaps that was a note for myself to dig into this more ;). I find doing larger code reviews hard and my pandas isn't all that strong.

bmoscon: Sounds like we should get the benchmarks branch merged in and find out ;)


One PR at a time! :)

PR #123, opened 2016-03-31
Address column naming with logging

bmoscon:


bmoscon:
@TomTaylorLondon does this satisfy you?

bmoscon:
since i changed the logic in ndarray, can you take a look too @jamesblackburn

arctic/store/_ndarray_store.py
```diff
@@ -387,7 +387,7 @@ def write(self, arctic_lib, version, symbol, item, previous_version, dtype=None)
         version['sha'] = self.checksum(item)
         
         if previous_version:
-            if version['dtype'] == str(dtype) \
+            if 'dtype' in previous_version and previous_version['dtype'] == str(dtype) \
```
jamesblackburn: I think the new code probably matches the intent.

However I'm wondering if this is a bug at all - and whether we should just delete the first condition in this if altogether. 

If the dtype has changed, and the previous `write`s `checksum` given `up_to` still matches (i.e. they're byte-identical) then an `_append` is still a reasonable thing to do?  

- A version document without a `sha` (created as part of an `append` call) will always be re-written on `write`.  
- `write`s which happen to share common initial data should work as the version document `dtype` will be correct as per line 382.

jamesblackburn:
Question regarding the validity of the `if` check altogether, otherwise LGTM

bmoscon:
@jamesblackburn i think you are correct and we can probably remove the whole condition. I can do that too if you like.

PR #131, opened 2016-04-29
Correct chunk start bounds calculation

### LINKED ISSUE #129: Tickstore: Reading multiple symbols at the same time with a daterange omits some chunks
Don't have an end-to-end reproduction yet, but consider this:

When reading several symbols at the same time from the tickstore, we determine the actual start date to stream the chunks from by taking the chunk with the latest start date that's still before our cutoff:

```
# Find the start bound
start_range = {}
first = last = None
if date_range.start:
    assert date_range.start.tzinfo
    start = date_range.start
    startq = self._symbol_query(symbol)
    startq.update({START: {'$lte': start}})
    first = self._collection.find_one(startq,
                                      # Service entirely from the index
                                      projection={START: 1, ID: 0},
                                      sort=[(START, pymongo.DESCENDING)])
if first:
    start_range['$gte'] = first[START]


                  ]
sym1   |      |   ]     | .....
sym2 |     |      ] |     .....
                  ]
          (1)(2)  ^Start
```

so in this case the query for sym1 and sym2 would return points 1 and 2 and find_one would return point 2. So when streaming the actual chunks we would consider only those which start on or after point 2, hence the first chunk for sym2 (starting at point 1) would be dropped.

I think we could either: 

* craft an aggregation query that groups by symbols (for each symbol, take the latest chunk start point that's still before the read start, then take the earliest of those points for all symbols). I tried this but it seems like the query doesn't use just the index anymore and times out: 

```
coll.aggregate([
                {'$match': {'e': {'$gte': datetime.datetime(2016, 4, 19, 0, 0)},
 's': {'$lte': datetime.datetime(2016, 4, 19, 0, 0)},
 'sy': {'$in': ['sym1', 'sym2']}}},
                {'$group': {'_id': 'sy', 'start': {'$max': START}}}
                ])
```
* iterate over all symbols, do the "find start bound" query for each one of them and then find the earliest one in Python.

mildbyte:
I think this fixes #129 -- make sure we get earliest start of the (for each symbol) closest chunk to our start datetime.

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -157,7 +157,10 @@ class TickStore(object):
         # We're assuming CLOSED_CLOSED on these Mongo queries
         assert date_range.interval == CLOSED_CLOSED
 
-        # Find the start bound
+        # Since we only index on the start of the chunk,
+        # we do a pre-flight aggregate query to find the point where the
+        # earliest relevant chunk starts.
+
         start_range = {}
         first_dt = last_dt = None
         if date_range.start:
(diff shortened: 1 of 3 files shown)


mildbyte:
Still needs some unit test fixing and a comment for the aggregation query :)

bmoscon:
looks like the test that failed was only due to the change you made in tickstore. Looks like just a tweak of the existing test is needed

mildbyte:
@bmoscon Yep, thanks, I think I fixed the unit test -- let's see if CI is okay with it.

jamesblackburn:
+1 thanks @mildbyte

PR #134, opened 2016-05-03
Allow replicaSet

pchavanne:
Hi,
Could you please allow replicaset?
This is one quick way of implementing it that is working here.
Regards,
Philippe

### DIFF
--- a/arctic/arctic.py
+++ b/arctic/arctic.py
@@ -62,7 +62,7 @@ class Arctic(object):
 
     def __init__(self, mongo_host, app_name=APPLICATION_NAME, allow_secondary=False,
                  socketTimeoutMS=10 * 60 * 1000, connectTimeoutMS=2 * 1000,
-                 serverSelectionTimeoutMS=30 * 1000):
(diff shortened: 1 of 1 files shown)


bmoscon:
I don't see any problem with something like this, though maybe @jamesblackburn can comment on it, this definitely will not be merged in until the unit tests are fixed and passing.

jamesblackburn:
I think the recommended way of doing this is to pass the `replicaSet` parameter in as part of the `mongodb://` URI:
https://api.mongodb.org/python/current/api/pymongo/mongo_client.html#pymongo.mongo_client.MongoClient
https://docs.mongodb.org/manual/reference/connection-string/

We also provide a hook for you to provide custom MongoDB URI resolution - so you can take a custom short-name and expand it to a fully-qualified MongoDB instance, for example:
`mongo` -> `mongodb://host1,host2?replicaSet=rs0`

See here:
https://github.com/manahl/arctic/blob/master/arctic/hooks.py#L8

pchavanne:
I will use the URI then, Thank you.

PR #139, opened 2016-05-10
Write a dummy base_sha if we don't have any in the previous version d…

mildbyte:
…ocument


bmoscon:
looks good, assuming no one uses the old from_version code path

PR #140, opened 2016-05-16
windows msvc stdint.h absent compatibility

mmitkevich:
pip install arctic complained about absent stdint.h on msvc
Used python 64 bit from MiniConda distribution.

### DIFF
--- a/src/_compress.pyx
+++ b/src/_compress.pyx
@@ -18,7 +18,9 @@ cimport libc.stdio
 cimport openmp
 
 from libc.stdlib cimport malloc, free, realloc
-from libc.stdint cimport uint8_t, uint32_t
+ctypedef unsigned char  uint8_t
+ctypedef unsigned int   uint32_t
+
 from libc.stdio cimport printf
 from cython.view cimport array as cvarray
 from cython.parallel import prange


src/_compress.pyx
```diff
@@ -18,7 +18,9 @@ cimport libc.stdio
 cimport openmp
 
 from libc.stdlib cimport malloc, free, realloc
-from libc.stdint cimport uint8_t, uint32_t
+ctypedef unsigned char  uint8_t
+ctypedef unsigned int   uint32_t
```
jamesblackburn: Should we make this `unsigned long int` to remove the ambiguity?

mmitkevich: Which ambiguity? `ctypedef unsigned long int uint32_t` ?
I think that would be uint64_t not uint32_t then
Actually ms vc++ doesn't have stdint.h on windows so only manual `ctypedef ` could help

jamesblackburn: `int` is compiler specific, so can be 2 or 4 bytes (or maybe 8?). Perhaps this is fine, I just thought `long int` is nearly always 32-bit (and `long long int` is 64 bits), maybe this isn't he case though...

jamesblackburn: Certainly:
http://linux.die.net/include/stdint.h
seem to agree with you - does this fix the build on Window?

PR #141, opened 2016-05-17
FSF-264 Allow reading of None values from arctic

lyrickh:
Will throw an error if the symbol exists but has been deleted (by checking the 'deleted' metadata) - tests for this behavior already exist.
Will no longer throw an exception when reading a legitimate None from version store.

### DIFF
--- a/tests/integration/store/test_version_store.py
+++ b/tests/integration/store/test_version_store.py
@@ -427,6 +427,11 @@ def test_delete_bson_versions(library):
     assert coll.versions.count() == 0
 
 
+def test_read_none_does_not_exception(library):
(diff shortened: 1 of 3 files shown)


tests/integration/store/test_version_store.py
```diff
@@ -427,6 +427,11 @@ def test_delete_bson_versions(library):
     assert coll.versions.count() == 0
 
 
+def test_read_none_does_not_exception(library):
+    library.write(symbol, None)
+    library.read(symbol)
```
jamesblackburn: Assert the result `is None` here?

richardbounds:
lgtm

PR #143, opened 2016-05-20
MDP-934 Ensure that the time always goes forwards in a tickstore bucket

jamesblackburn:
We now raise an arctic.exceptions.UnorderedDataException if data goes
backwards within one chunk.


jamesblackburn:
@bmoscon @richardbounds @mildbyte all ok?

bmoscon:
looks good to me.

PR #146, opened 2016-05-27
Make sure we don't fetch chunks that don't span the start point

mildbyte:


arctic/tickstore/tickstore.py
```diff
@@ -176,17 +176,28 @@ def _mongo_date_range_query(self, symbol, date_range):
...
+                        break
             except StopIteration:
                 pass
+            
         if first_dt:
             start_range['$gte'] = first_dt
+        else:
+            # If all chunks start inside of the range, make sure
+            # we don't fetch everything
+            start_range['$gte'] = start
```
jamesblackburn: Is this edge-case tested too?

mildbyte: It is now :)

bmoscon:
looks good to me.

jamesblackburn:
LGTM @mildbyte 

Don't forget to update the Changelog :)

Issue #105, opened 2016-03-08
Quandl data-set in README doesn't exist


NickAltmann:
The README makes reference to "NASDAQ/AAPL" on Quandl.  Perhaps they have re-shuffled the data but no such set exists.  All worked fine for me using "WIKI/AAPL".

jamesblackburn:
Fix merged

PR #149, opened 2016-06-17
quandl changes

xgdgsc:
make the demo work with current quandl


jamesblackburn:
Many thanks, looks like this fixes #105

PR #154, opened 2016-06-23
Re-cache the library quota periodically

rob256:
Previously the quota was cached indefinitely, and if the quota was increased, the new quota wasn't being picked up until the application was restarted. Re-checking the quota after every countdown (half-life) now picks up any changes made to the library whilst the application is still running.


jamesblackburn:
LGTM

PR #161, opened 2016-07-12
TickStore.max_date: convert to local timezone

mckelvin:
The return value of `TickStore.max_date()` was a `datetime` of UTC timezone without `tzinfo`. It should be converted to local timezone. Please correct if this's a feature.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /build
+*.eggs/
 *.pyc
 /src/_compress.c
 *.egg
@@ -13,3 +14,5 @@ htmlcov
 coverage.xml
 junit.xml
 /tmp/
+
+*.sw[op]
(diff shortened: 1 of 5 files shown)


bmoscon:
@jamesblackburn do you know if the lack of conversion is intentional or not?

bmoscon:
@mckelvin latest merge into master changed a lot of the TZ stuff. You might need to re-test this. Not sure. Definitely needs to have the merge conflicts resolved.

mckelvin:
@bmoscon Thanks for reminding. Updated now.

bmoscon:
based on the latest PR that went in, I believe that it was their intent to store everything in UTC in tickstore.  @jamesblackburn @reasto can probably confirm or deny this.

mckelvin:
@bmoscon Yes. Every value of `datetime` type is stored in UTC, and that's the default behaviour of MongoDB. When calling `.read`, datetime index in the returning DateFrame is converted and presented in user's local timezone. So I think it's better to conform by converting to user local timezone.

jamesblackburn:
Seem reasonable to me.  The main issue is returning a non-naive datetime which isn't comparable to a naive one.

I don't think we use this max_date method anywhere here...   @reasto can you think will making this non-naive have any impact for you?

bmoscon:
@jamesblackburn @reasto what do you want to do with this one?

jamesblackburn:
This is minor API change - but seems reasonable to me especially given the behaviour of `read`.

Issue #166, opened 2016-07-12
ChunkStore missing index on SHA


bmoscon:
when overwriting a large dataframe there is a significant delay when removing the old chunks. We're missing an index on the SHA (which is how the cleanup of old, unused chunks takes place).

bmoscon:
fixed

PR #167, opened 2016-07-12
Chunkstore SHA Index

bmoscon:
see issue #166 for complete details.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -6,6 +6,7 @@
   * Bugfix: TickStore.read now respects `allow_secondary=True`
   * Bugfix: #147 Add get_info method to ChunkStore
   * Bugfix: Periodically re-cache the library.quota to pick up any changes
+  * Bugfix: #166 Add index on SHA for ChunkStore
 
 ### 1.25 (2016-05-23)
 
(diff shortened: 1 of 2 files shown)


bmoscon:
@jamesblackburn take a look and let me know

jamesblackburn:
Don't we query on (symbol, sha) when we cleanup old chunks?  I wouldn't think we would need an additional index here...

bmoscon:
@jamesblackburn check out:

https://github.com/manahl/arctic/blob/master/arctic/chunkstore/chunkstore.py#L222

we query just on SHA's to remove old chunks that are no longer used. No reason why I couldn't query on symbol / sha, but also no reason why I can't add a new index,  right?

jamesblackburn:
It's just another index to maintain on each chunk write.  

We previously found that `'sha': {'$in': list(previous_shas)` is also quite slow, so adding the symbol in there will likely help (as the number of SHAs in the sub btree is smaller).

bmoscon:
ok, will change it.

bmoscon:
@jamesblackburn take a look now

jamesblackburn:
+1

Issue #172, opened 2016-07-15
Bug: ChunkStore read with chunk_range not working in all cases


bmoscon:
example:

```
df = DataFrame(data={'data': [1, 2, 3, 4, 5, 6]},
                   index=MultiIndex.from_tuples([(dt(2016, 1, 1), 1),
                                                 (dt(2016, 1, 2), 1),
                                                 (dt(2016, 2, 1), 1),
                                                 (dt(2016, 2, 2), 1),
                                                 (dt(2016, 3, 1), 1),
                                                 (dt(2016, 3, 2), 1)],
                                                names=['date', 'id'])
                   )

chunkstore_lib.write('test', df, 'M')
chunkstore_lib.read('test', chunk_range=DateRange(dt(2016, 1, 2), dt(2016, 3, 1))')
```

should read data back from Jan 2nd, 2016 to March 1st, 2016. Instead it reads out data from Feb 1st to March 3rd. This because its only looking at the START key in Mongo, and not the end key (should look at both).

bmoscon:
fixed

PR #173, opened 2016-07-15
Fixes #171 and #172

### LINKED ISSUE #171: Feature: add optional chunk_range to ChunkStore's delete method
Expected Behavior - delete everything between the range, even if it means only partially deleting a chunk

bmoscon:
See issues #171 and #172  for more information.

### DIFF
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -73,18 +73,32 @@ class ChunkStore(object):
     def __repr__(self):
         return str(self)
 
-    def delete(self, symbol):
+    def delete(self, symbol, chunk_range=None):
         """
-        Delete all chunks for a symbol.
+        Delete all chunks for a symbol, or optionally, chunks within a range
 
         Parameters
         ----------
         symbol : `str`
             symbol name for the item
         """
-        query = {"symbol": symbol}
-        self._collection.delete_many(query)
-        self._collection.symbols.delete_many(query)
+        if chunk_range:
+            # read out chunks that fall within the range and filter out
+            # data within the range
+            df = self.read(symbol, chunk_range=chunk_range, no_filter=True)
+            df = df[(df.index.get_level_values('date') < chunk_range[0]) | (df.index.get_level_values('date') > chunk_range[1])]
+
+
+            # remove chunks, and update any remaining data
+            query = {'symbol': symbol}
+            query.update(self.chunker.to_mongo(chunk_range))
+            self._collection.delete_many(query)
+            self.update(symbol, df)
+
+        else:
+            query = {"symbol": symbol}
+            self._collection.delete_many(query)
+            self._collection.symbols.delete_many(query)
 
     def list_symbols(self):
         """
@@ -99,7 +113,7 @@ class ChunkStore(object):
     def _get_symbol_info(self, symbol):
         return self._symbols.find_one({'symbol': symbol})
 
-    def read(self, symbol, chunk_range=None):
+    def read(self, symbol, chunk_range=None, no_filter=False):
         """
         Reads data for a given symbol from the database.
 
@@ -110,6 +124,9 @@ class ChunkStore(object):
         chunk_range: object
             corresponding range object for the specified chunker (for
             DateChunker it is a DateRange object)
+        no_filter: boolean
+            perform chunk level filtering on the data (see filter() in _chunker)
+            only applicable when chunk_range is specified
 
         Returns
         -------
@@ -124,7 +141,7 @@ class ChunkStore(object):
                 }
 
         if chunk_range:
-            spec['start'] = self.chunker.to_mongo(chunk_range)
+            spec.update(self.chunker.to_mongo(chunk_range))
 
         segments = []
         for _, x in enumerate(self._collection.find(spec, sort=[('start', pymongo.ASCENDING)],)):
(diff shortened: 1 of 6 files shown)


bmoscon:
@jamesblackburn

tests/integration/chunkstore/test_chunkstore.py
```diff
@@ -638,3 +638,51 @@ def test_dtype_mismatch_error(chunkstore_lib):
...
+                                                 (dt(2016, 2, 2), 1),
+                                                 (dt(2016, 2, 3), 1),
+                                                 (dt(2016, 3, 1), 1),
+                                                 (dt(2016, 3, 2), 1),
+                                                 (dt(2016, 3, 3), 1)],
+                                                names=['date', 'id'])
+                   )
+
+    chunkstore_lib.write('test', df, 'M')
+    assert(chunkstore_lib.read('test', chunk_range=DateRange(dt(2016, 1, 1), dt(2016, 1, 1))).index.get_level_values('date')[0] == dt(2016,1,1))
```
jamesblackburn: Is `DateChunker` the only implementation of `Chunker` currently?  The API here is a little inconsistent with `date_range=` which is used for range-slicing on all the other classes in arctic.

arctic/chunkstore/date_chunker.py
```diff
@@ -74,7 +74,17 @@ def to_start_end(self, data):
         return self._get_date_range(data)
 
     def to_mongo(self, range_obj):
-        return range_obj.mongo_query()
+        if range_obj[0] and range_obj[1]:
+            return {'$and': [{'start': {'$lte': range_obj[1]}}, {'end': {'$gte': range_obj[0]}}]}
+        elif range_obj[0]:
+            return {'end': {'$gte': range_obj[0]}}
```
jamesblackburn: I think this query will end up scanning all chunks for a symbol...

arctic/chunkstore/date_chunker.py
```diff
@@ -74,7 +74,17 @@ def to_start_end(self, data):
         return self._get_date_range(data)
 
     def to_mongo(self, range_obj):
-        return range_obj.mongo_query()
+        if range_obj[0] and range_obj[1]:
+            return {'$and': [{'start': {'$lte': range_obj[1]}}, {'end': {'$gte': range_obj[0]}}]}
```
jamesblackburn: Would be good to do an `.explain()` on these queries... I suspect Mongo won't do anything efficient with a query that contains a `start` and `end` like this

jamesblackburn: Are `range_obj` `DateRange`s?  Might be more explicit to use`range_obj.start` ; `range_obj.end`

tests/integration/chunkstore/test_chunkstore.py
bmoscon: Yes, it is the only implementation. I didnt want a chunker to be tied to a date, its supposed to be very extensible, so it doesnt have to be a date range. You could write a chunker that did things by some other ordering (like alphabetical? By market, security id, whatever).

arctic/chunkstore/date_chunker.py
bmoscon: yes, but chunkstore modifies this and adds in the symbol. This is only responsible for returning the date portion of the query.

arctic/chunkstore/date_chunker.py
bmoscon: I have an index on symbol, start, end, so I think its ok. Also it is a DateRange so I'll change it to use start and end.

arctic/chunkstore/date_chunker.py
```diff
@@ -74,7 +74,17 @@ def to_start_end(self, data):
...
+        elif range_obj.start:
+            return {'end': {'$gte': range_obj.start}}
+        elif range_obj.end:
+            return {'start': {'$lte': range_obj.end}}
+        else:
+            return {}
 
     def filter(self, data, range_obj):
-        return data.ix[range_obj[0]:range_obj[1]]
+        return data.ix[range_obj.start:range_obj.end]
```
jamesblackburn: I think using `to_pandas_closed_closed` would be ensure this is right for the `OPEN` / `CLOSED`

arctic/chunkstore/chunkstore.py
```diff
@@ -110,6 +123,9 @@ def read(self, symbol, chunk_range=None):
         chunk_range: object
             corresponding range object for the specified chunker (for
             DateChunker it is a DateRange object)
+        no_filter: boolean
+            perform chunk level filtering on the data (see filter() in _chunker)
```
jamesblackburn: _**don't** perform_ 

When would this be used?

jamesblackburn:
Few minor comments. LGTM!

arctic/chunkstore/chunkstore.py
bmoscon: 1. testing
2. its also used when deleting a range. If you filter out the data during a read, you can't preserve parts of the chunk that you might want to save. (ex: if the delete range is March 15 - April 15 and your chunking strat is monthly. You need to not filter the chunks when reading so you can preserve the March 1-14 data and the April 16-30 data).

arctic/chunkstore/date_chunker.py
bmoscon: Will look into that and test it out.

bmoscon: seems to work fine with every range - no need to call that API. I added some unit tests for it.

arctic/chunkstore/chunkstore.py
bmoscon: note that i decided i didnt like calling it no_filter, so its no longer a negative and the default behavior is now true rather than false.

PR #175, opened 2016-07-18
[WIP] Bootstrap setup process

mckelvin:
`Cython` and `six` are required when building arctic. Lacking of these two may break `pip install git+https://github.com/manahl/arctic.git`.

The problem can be reproduced in a clean and initial Python environment:

    docker run --rm -it python:2.7.12 pip install git+https://github.com/manahl/arctic.git

And it's solved in this PR:

    docker run --rm -it python:2.7.12 pip install git+https://github.com/mckelvin/arctic.git@setup

@bmoscon @jamesblackburn Please review.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -21,8 +21,6 @@ from setuptools import setup
 from setuptools.extension import Extension
 from setuptools import find_packages
 from setuptools.command.test import test as TestCommand
-from Cython.Build import cythonize
-import six
 import sys
 
 # Convert Markdown to RST for PyPI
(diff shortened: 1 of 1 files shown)


mckelvin:
This solution only works in Python 2.*.

bmoscon:
@mckelvin thats exactly why we don't have it in there :) Used to, but once we supported 2 and 3, we couldn't find a way to get it to work with both. Admittedly I haven't spent a ton of time on it, so its probably possible, I just haven't looked into it enough.

Issue #178, opened 2016-07-19
ChunkStore - Append not working correctly


bmoscon:
Behavior of append is incorrect - should simply append whatever data is in the dataframe to the symbol.

bmoscon:
fixed

PR #179, opened 2016-07-19
Fix issue #178

bmoscon:
- Fix append
- Rework Append and Update to share code (since 99% of their code is identical)
- Change how start/end are stored in Mongo. Previously was the actual start/end of the data, now is start/end of the chunk size. This makes the code less complex.


bmoscon:
@jamesblackburn

jamesblackburn:
looks good

PR #180, opened 2016-07-20
_compression: Fix uninitialized logger

mckelvin:
While importing arctic, a warning message `No handlers could be found for logger "arctic._compression"` will occur, which is quite annoying. I checked the code and found it's caused by calling `logger.warning` before `setup_logging()`. To avoid this behaviour, I guess it's a better choice to import clz4 lazily.

### DIFF
--- a/arctic/_compression.py
+++ b/arctic/_compression.py
@@ -3,13 +3,7 @@ import os
 import logging
 
 logger = logging.getLogger(__name__)
-
-try:
-    from . import _compress as clz4
-except ImportError:
-    logger.warning("Couldn't import cython lz4")
-    import lz4 as clz4
-
+clz4 = None
 
 # switch to parallel LZ4 compress (and potentially other parallel stuff), Default True
 ENABLE_PARALLEL = not os.environ.get('DISABLE_PARALLEL')
@@ -40,6 +34,7 @@ def compress_array(str_list):
     if not ENABLE_PARALLEL:
         return [lz4.compress(s) for s in str_list]
 
+    _lazy_import_clz4()
     # Less than 50 chunks its quicker to compress sequentially..
     if len(str_list) > LZ4_N_PARALLEL:
         return clz4.compressarr(str_list)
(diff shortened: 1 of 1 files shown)


jamesblackburn:
It looks like this is in the exception handler path for the import.  Is this only happening because that `.so` isn't building for you? I would have thought that as we use logging in lots of places this might be a general problem?

mckelvin:
> It looks like this is in the exception handler path for the import. Is this only happening because that .so isn't building for you? 

This happens to me when using arctic in Ubuntu 14.04. The arctic is built in [wheel](http://pythonwheels.com) from a [python:2.7.12 docker image](https://hub.docker.com/_/python/).

```
$ python -c "import arctic._compress"
No handlers could be found for logger "arctic._compression"
  File "<string>", line 1, in <module>
ImportError: /usr/lib/x86_64-linux-gnu/libgomp.so.1: version `GOMP_4.0' not found (required by /vagrant/projects/hom/venv/local/lib/python2.7/site-packages/arctic/_compress.so)
```

Actually that's not a big deal and I'd like to ignore the broken `_compress.so` since it will fallback to use `lz4`.

* * *

> I would have thought that as we use logging in lots of places this might be a general problem?

Agree but I'd appreciate it if we can take control of what writes to stderr.

mckelvin:
> so you're importing arctic before you've setup your logger

Yes. Calling `setup_logging` in a main function should be a better choice. And that's exactly the way how `setup_logging` is use in [scripts/arctic_init_library](https://github.com/manahl/arctic/blob/6d4acca/arctic/scripts/arctic_init_library.py#L21).

jamesblackburn:
ok it sounds like w're in agreement @mckelvin 

I'd rather not make this lazy as:
 - users should have a logger configured when using this library
 - we expect this `.so` to be built and exist for our users
 - the additional code does increase the complexity in this module

mckelvin:
@jamesblackburn Thanks for your response. 

> the additional code does increase the complexity in this module

Agree.

> we expect this .so to be built and exist for our users

Agree. But if it's not existed, user should be warned with detailed message instead of `No handlers could be found for logger "arctic._compression"`, which is meaningless if he doesn't inspect the code. 

**May be we can use [`warnings`](https://docs.python.org/2/library/warnings.html) module to show this message ?**

> users should have a logger configured when using this library

Users should configure in the beginning of `main()`, just like what https://github.com/manahl/arctic/blob/6e1f0fe/arctic/scripts/arctic_copy_data.py#L65 does. But this warning happened in import-time. The user have to write dirty code like:

```python
import foo
import bar
import logging
logging.basicConfig(...)  ## calling this at import-time
import arctic

def naive(...):
    ...

def main():
    ...

...
``` 

So, I think use `warnings` should be a better idea. What do you think @jamesblackburn ?

Issue #162, opened 2016-07-12
ChunkStore - Rename API


bmoscon:
now that we are using this more internally, we've discovered it would be nice to have a rename API - that is, the ability to rename a symbol. Would be faster than copying data out and then back in (presumably).

bmoscon:
@jamesblackburn i assume there is no opposition to this?

jamesblackburn:
None from me

bmoscon:
fixed

PR #188, opened 2016-08-04
Feature #162

bmoscon:
See #162 - Adds a rename method to ChunkStore to allow symbol renaming.

### DIFF
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -126,6 +126,42 @@ class ChunkStore(object):
     def _get_symbol_info(self, symbol):
         return self._symbols.find_one({SYMBOL: symbol})
 
+    def rename(self, from_symbol, to_symbol):
+        """
+        Rename a symbol
+
+        Parameters
+        ----------
+        from_symbol: str
+            the existing symbol that will be renamed
+        to_symbol: str
+            the new symbol name
+        """
+
+        sym = self._get_symbol_info(from_symbol)
+        if not sym:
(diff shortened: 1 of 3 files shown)


bmoscon:
@jamesblackburn

bmoscon:
@jamesblackburn I had to change the travis and circle scripts to update Mongo. The default version is some ancient 2.4 version. I'll squash the merge :)

bmoscon:
@jamesblackburn note that the rename method only works on newer versions of Mongo. Not sure if thats important or not. Based on some things  I read online it looks like there was some bug in older versions that was fixed in 2.6. I could change the code to still support older versions of Mongo, or we could say we only support newer versions than 2.4 (or we could say everything except rename works in 2.4). Or something else?

arctic/chunkstore/chunkstore.py
```diff
@@ -126,6 +126,44 @@ def list_symbols(self):
...
+
+        chunks = []
+        for x in self._collection.find({SYMBOL: from_symbol}, sort=[(START, pymongo.ASCENDING)],):
+            chunks.append(x)
+
+        bulk = self._collection.initialize_unordered_bulk_op()
+        for chunk in chunks:
+            chunk.pop('_id', None)
+            chunk[SYMBOL] = to_symbol
+            chunk[SHA] = self._checksum(to_symbol, chunk)
```
jamesblackburn: Could be optimised to remove the symbol from the SHA, then this turns into a simple `update` operation.

jamesblackburn:
LGTM!

PR #196, opened 2016-08-12
Tickstore top level write fix

reasto:
Date ranges for top level libraries are stored as timezone naive datetimes. These are now converted into UTC before using these dates to split up the tick data into slices to writing to the low level tickstore libraries. We also fix an xfailed test and add an extra check that data written to in one timezone can be fetched out in another.


jamesblackburn:
LGTM!

bmoscon:
:+1:

PR #202, opened 2016-08-25
Pytest fix

bmoscon:
Latest version of PyTest deprecates a call that pytest-dbfixtures makes, resulting in an error when running pytest. This forces setup.py and Travis to use the previous version of pytest


bmoscon:
@jamesblackburn will squash all the commits into a single commit message ;)

jamesblackburn:
Ok

Issue #209, opened 2016-09-07
Add upsert capability to update method in ChunkStore


bmoscon:

bmoscon:
resolved with #210

PR #210, opened 2016-09-07
Upsert capability for update in ChunkStore

bmoscon:
fix for #209

### DIFF
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -346,7 +346,7 @@ class ChunkStore(object):
         """
         self.__update(symbol, item, combine_method=self.__concat)
 
-    def update(self, symbol, item, chunk_range=None):
+    def update(self, symbol, item, chunk_range=None, upsert=False, chunk_size=None):
         """
         Overwrites data in DB with data in item for the given symbol.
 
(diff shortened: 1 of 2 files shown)


bmoscon:
@jamesblackburn @TomTaylorLondon

bmoscon:
@Wilfred

arctic/chunkstore/chunkstore.py
```diff
@@ -363,8 +363,15 @@ def update(self, symbol, item, chunk_range=None):
             range and overwrite it with the data in item. This allows the user
             to update with data that might only be a subset of the
             original data.
+        upsert: True or False
```
Wilfred: True or False: in other words, bool :)

arctic/chunkstore/chunkstore.py
```diff
@@ -363,8 +363,15 @@ def update(self, symbol, item, chunk_range=None):
...
+            if True, will write the data even if the symbol does not exist.
+            If true and an upsert is performed, chunk_range will be ignored
+        chunk_size: ?
```
Wilfred: Presumably this is an integer?

arctic/chunkstore/chunkstore.py
bmoscon: point taken, will change ;)

arctic/chunkstore/chunkstore.py
```diff
@@ -346,7 +346,7 @@ def append(self, symbol, item):
         """
         self.__update(symbol, item, combine_method=self.__concat)
 
-    def update(self, symbol, item, chunk_range=None):
+    def update(self, symbol, item, chunk_range=None, upsert=False, chunk_size=None):
```
Wilfred: Super pedantic: I might be tempted to keep the chunking arguments together.

arctic/chunkstore/chunkstore.py
bmoscon: it depends on the chunker. for the date chunker, its an arctic DateRange object, or a Pandas DatetimeIndex

Wilfred:
Apart from my minor nits, LGTM :+1:

arctic/chunkstore/chunkstore.py
bmoscon: I can see your point, to me it makes more sense like so, after all you only have to specify a chunksize if you override upsert and set it to True

Issue #212, opened 2016-09-09
New "pass thru" chunk type for ChunkStore


bmoscon:
We want a way to write one off dataframes that dont have any datetime information (i.e. like a look up table that maps country codes to currency codes).

Issue #193, opened 2016-08-05
Complete support for multiple chunker/serializers in ChunkStore


bmoscon:
Should be able to set different chunkers and serializers per symbol (and also a default one for the library). Currently hard coded to use the date chunker and the numpy array serializer.

PR #213, opened 2016-09-09
Support for different chunkers/serializers by symbol

bmoscon:
Addresses issue #193  and issue #212 
Also #222 

take a look @jamesblackburn and @miketkelly

### DIFF
--- a/arctic/chunkstore/_chunker.py
+++ b/arctic/chunkstore/_chunker.py
@@ -6,7 +6,7 @@ class Chunker(object):
 
     def to_chunks(self, data, **kwargs):
         """
-        Chunks data
+        Chunks data. keyword args passed in from write API
 
         returns
         -------
--- a/arctic/chunkstore/_chunker.py
+++ b/arctic/chunkstore/_chunker.py
@@ -4,7 +4,7 @@ END = 'e'
 
 class Chunker(object):
 
-    def to_chunks(self, data, *args, **kwargs):
+    def to_chunks(self, data, chunk_size, **kwargs):
         """
         Chunks data
 
--- a/arctic/chunkstore/_chunker.py
+++ b/arctic/chunkstore/_chunker.py
@@ -4,15 +4,16 @@ END = 'e'
 
 class Chunker(object):
 
-    def to_chunks(self, data, chunk_size, **kwargs):
+    def to_chunks(self, data, **kwargs):
         """
         Chunks data
 
         returns
         -------
-        generator that produces 3-tuples
+        generator that produces 4-tuples
             (chunk start index/marker/key,
             chunk end index/marker/key,
+            chunk_size,
             chunked data)
         """
         raise NotImplementedError
--- a/arctic/chunkstore/_chunker.py
+++ b/arctic/chunkstore/_chunker.py
@@ -36,7 +36,7 @@ class Chunker(object):
 
         returns
         -------
-        string
+        dict
         """
         raise NotImplementedError
 
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -50,7 +50,7 @@ class ChunkStore(object):
         self._collection.create_index([(SYMBOL, pymongo.HASHED)],
                                       background=True)
         self._collection.create_index([(SYMBOL, pymongo.ASCENDING),
-                                      (SHA, pymongo.ASCENDING)],
+                                       (SHA, pymongo.ASCENDING)],
                                       unique=True,
                                       background=True)
         self._collection.create_index([(SYMBOL, pymongo.ASCENDING),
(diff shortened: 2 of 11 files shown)


bmoscon:
Alright, this is ready for review and merge @jamesblackburn @miketkelly

arctic/chunkstore/_chunker.py
```diff
@@ -4,15 +4,16 @@
 
 class Chunker(object):
 
-    def to_chunks(self, data, *args, **kwargs):
```
jamesblackburn: Maybe worth saying that `kwargs` is passed through from the `write` public API

arctic/serialization/numpy_arrays.py
```diff
@@ -120,28 +77,64 @@ def docify(self, df):
...
+
         for c in df:
-            meta = {NAME: str(c)}
             try:
-                doc[DATA][str(c)] = self.converter.docify(df[c].values, meta)
+                columns.append(str(c))
+                arr, mask = self._convert_types(df[c].values)
+                dtypes[str(c)] = arr.dtype.str
+                if mask is not None:
+                    masks[str(c)] = Binary(compress(mask.tostring()))
```
jamesblackburn: Opportunity for parallel compression here

arctic/chunkstore/chunkstore.py
```diff
@@ -188,19 +203,31 @@ def read(self, symbol, chunk_range=None, columns=None, filter_data=True):
...
 
         segments = []
-        for x in self._collection.find(spec, sort=[(START, pymongo.ASCENDING)],):
-            segments.append(x[DATA])
+        parts = []
+        for x in self._collection.find(spec, sort=[(START, pymongo.ASCENDING), (SEGMENT, pymongo.ASCENDING)],):
+            if x[SEGMENT] > -1:
+                parts.append(x[DATA])
+            else:
+                if parts:
```
jamesblackburn: I don't understand what this branch of the else is for.

jamesblackburn: Ah, I see that segments of chunks are treated specially.  Is it not possible to to treat the segments as just another chunk - this make the code a fair bit more complicated...

arctic/chunkstore/chunkstore.py
```diff
@@ -266,67 +308,58 @@ def write(self, symbol, item, chunk_size):
...
+                    chunk[SEGMENT] = i
+                else:
+                    chunk[SEGMENT] = -1
+                chunk[START] = start
+                chunk[END] = end
+                chunk[SYMBOL] = symbol
+                dates = [chunker.chunk_to_str(start), chunker.chunk_to_str(end), str(chunk[SEGMENT]).encode('ascii')]
+                sha = self._checksum(dates, data[DATA])
+                chunk[SHA] = sha
+                bulk.find({SYMBOL: symbol, START: start, END: end, SEGMENT: chunk[SEGMENT]}
```
jamesblackburn: What happens if someone does a chunk update, the old chunk is segmented, and the new chunk has fewer segments than the old one?  I think we'll have a rogue segment remaining?

arctic/chunkstore/passthrough_chunker.py
```diff
@@ -0,0 +1,75 @@
+from pandas import DataFrame, Series
+
+from ._chunker import Chunker, START, END
+
+
+class PassthroughChunker(Chunker):
```
jamesblackburn: Is this just for tests?

arctic/chunkstore/passthrough_chunker.py
```diff
@@ -0,0 +1,75 @@
...
+        returns
+        -------
+        string
+        """
+        return b'NA'
+
+    def to_mongo(self, range_obj):
+        """
+        returns mongo query against range object.
+        since range object is not valid, returns empty dict
```
jamesblackburn: Can / should this just return an empty dict as the docstring says?

arctic/chunkstore/date_chunker.py
```diff
@@ -5,14 +5,16 @@
 
 
 class DateChunker(Chunker):
-    def to_chunks(self, df, chunk_size):
+    TYPE = 'date'
+
+    def to_chunks(self, df, chunk_size='D', **kwargs):
```
jamesblackburn: I think `Y` or `M` might be a better default.  I found you get a big performance wins for larger and fewer chunks.

jamesblackburn: And I found that future minute bars generally easily fit into `Y` buckets.  Maybe you've found otherwise?

bmoscon: I put 'daily' because thats just what most of our data is. The user should really specify one, not just reply on the default, but its possible that a chunker might not need a user specified size, so i didnt want to make it a requirement of the design

arctic/chunkstore/passthrough_chunker.py
bmoscon: yes, thats fine. there is only one "chunk" per symbol, so matching on symbol alone is fine. the range keys (start/end) arent necessary or used really.

arctic/chunkstore/chunkstore.py
bmoscon: I'll add a unit test

arctic/chunkstore/_chunker.py
bmoscon: fixed

arctic/serialization/numpy_arrays.py
bmoscon: did you mean i should compress mask and arr in paralle? or did you mean i should save off all the arrs and compress all of them in parallel?

Issue #206, opened 2016-08-31
How to store strings in a pd.DataFrame in tickstore?  add 'O' dtype?


waterkirin:
Hi, I am new to arctic and I am trying to write my pandas.dataframe to tickstore.

Pandas uses native python strings, which require an object dtype. Therefore an array of strings in pd.dataframe's dtype is dtype('O'), which is an object.
See the following links.

[http://stackoverflow.com/questions/34881079/pandas-distinction-between-str-and-object-types?noredirect=1&lq=1](url)

[http://stackoverflow.com/questions/21018654/strings-in-a-dataframe-but-dtype-is-object](url)

[http://docs.scipy.org/doc/numpy/reference/generated/numpy.dtype.kind.html#numpy.dtype.kind](url)

And I notice that in tickstore.py, function   _ensure_supported_dtypes(array)

```
    def _ensure_supported_dtypes(array):
        # We only support these types for now, as we need to read them in Java
        if (array.dtype.kind) == 'i':
            array = array.astype('<i8')
        elif (array.dtype.kind) == 'f':
            array = array.astype('<f8')
        elif (array.dtype.kind) in ('U', 'S', 'O'):        # add 'O' here, is this OK? 
            array = array.astype(np.unicode_)
```

So I just use python, could I just add dtype 'O' there? Is that OK?

Thank you guys!

bmoscon:
What happens when you try and write strings to it now?

richardbounds:
Artic stores all of its data (including strings) as compressed fixed-width C-style arrays. In an array of type 'O', the array contains pointers to objects that are elsewhere in memory, so Arctic can't do anything with that.

Your change may work if the .astype(np.unicode_) successfully converts the object array to a fixed-width Unicode array, but it will probably raise some other error if the objects in the array are not strings. Personally, I wouldn't do that here - I'd rather do the conversion of any object-type columns before trying to write to Arctic.

jonasleotech:
Richard,

Can you please elaborate on how to do the conversation of an object-type string column to a numpy dtype string in a data-frame before writing it to the TickStore?

It seems like Pandas data-frames always store strings as native python strings even when I try to force them to 'S' and therefore perhaps we should allow the solution of adding 'O' as a supported type and let .astype(np.unicode_) raise an error if it's not a supported string?

[...]

PR #215, opened 2016-09-10
Fixes issue 214 - Tickstore not supporting data frames with string cols

jonasleotech:
This pull request fixes issue 214, at least for the specific needs I have.

The fix simply allows for dtype.kind = 'O' to be casted to unicode to support strings.

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -587,7 +587,7 @@ class TickStore(object):
             array = array.astype('<i8')
         elif (array.dtype.kind) == 'f':
             array = array.astype('<f8')
-        elif (array.dtype.kind) in ('U', 'S'):
+        elif (array.dtype.kind) in ('U', 'S', 'O'):
             array = array.astype(np.unicode_)
         else:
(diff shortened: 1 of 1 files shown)


bmoscon:
see #206

jamesblackburn:
I don't think this will work in the general case - what happens when the object numpy column is pointing to non-string objects?

bmoscon:
in reply to @jamesblackburn I think we'd have to do like we do in VersionStore, if its an Object column, check if its a string, otherwise raise an error/abort

arctic/tickstore/tickstore.py
```diff
@@ -587,7 +587,7 @@ def _ensure_supported_dtypes(array):
             array = array.astype('<i8')
         elif (array.dtype.kind) == 'f':
             array = array.astype('<f8')
-        elif (array.dtype.kind) in ('U', 'S'):
+        elif (array.dtype.kind) in ('U', 'S', 'O'):
             array = array.astype(np.unicode_)
```
TomTaylorLondon: Will this blow up for non-string objects - for example a DataFrame of dicts?

jamesblackburn:
AFAICS this patch won't work with other object types; closing for now.

PR #217, opened 2016-09-10
Fixes issue 216 - Tickstore not supporting named index column

jonasleotech:
The tickstore does not support saving dataframes that where the index column have a name.

This is due to an assumption that the index column will always be named "index" in the dictionary created by df.to_records.

The fix is to determine the real name of the index column if there's a name specified and fall back to "index" if the index does not have a name.

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -628,6 +628,7 @@ class TickStore(object):
         logger.warning("NB treating all values as 'exists' - no longer sparse")
         rowmask = Binary(lz4.compressHC(np.packbits(np.ones(len(df), dtype='uint8'))))
 
+        indexcolumnname = df.index.names[0] or "index"
         recs = df.to_records(convert_datetime64=False)
         for col in df:
             array = TickStore._ensure_supported_dtypes(recs[col])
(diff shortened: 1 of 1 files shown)


bmoscon:
looks fine to me. Any other thoughts @manahl/arctic ?

bmoscon:
upon further review of the code in tickstore, it looks like its hardcoded to expect "index" as the index name in many other places. If you change this to allow other index names, those will all need to be modified as well . . .

jonasleotech:
@bmoscon The change in this pull request is purposely only handling the extraction of index values  from the dataframe, and not changing the documented behaviour of a index key in the initial_image. (which are the only places where index values are extracted by name 'index' in the pandas dataframe scenario).

This means that we can save dataframes created by: 
```python
df = pd.read_csv('trades.csv', parse_dates=['trade_timestamp'], index_col='trade_timestamp')
```
 without any special handling, and would still allow us to provide an initial image with a specific timestamp using index as a dictionary key as explained in the doc like: 
```
{'index': dt(2010,1,1,tzinfo=pytc.UTC, 'price': 123, 'volume': 123123123 }.
```

This without breaking any existing scenarios or contradicting the documentation.

Example:
```python
df = pd.read_csv('trades.csv', parse_dates=['trade_timestamp'], index_col='trade_timestamp')
lib.write("SYMBOL", df, {'index': dt(2010,1,1,tzinfo=pytc.UTC, 'price': 123, 'volume': 123123123 })
```

bmoscon:
@jonasleotech thanks for clearing that up. I think its reasonable. What do you think @jamesblackburn ?

jamesblackburn:
Can we get a test that uses this?  We're very keen on unit tests for this codebase.

bmoscon:
fixed

Issue #220, opened 2016-09-13
VersionStore should bypass lz4 when data is too large to compress


bmoscon:
python lz4 is 32bit and cannot compress very large data

PR #221, opened 2016-09-13
Fix issues #219 and #220

### LINKED ISSUE #219: VersionStore pickling code produces checksum collisions in very large data
for very large data (2GB) I'm able to produce collisions in checksums, resulting in overwritten chunks in mongo

bmoscon:
Fixes #219 and #220

@TomTaylorLondon @vmereuta @Wilfred

### DIFF
--- a/arctic/store/_pickle_store.py
+++ b/arctic/store/_pickle_store.py
@@ -9,6 +9,7 @@ import pymongo
 from arctic.store._version_store_utils import checksum, pickle_compat_load
 
 _MAGIC_CHUNKED = '__chunked__'
+_MAGIC_CHUNKED_NO_COMPRESS = _MAGIC_CHUNKED + 'no_compress__'
 _CHUNK_SIZE = 15 * 1024 * 1024  # 15MB
 _MAX_BSON_ENCODE = 256 * 1024  # 256K - don't fill up the version document with encoded bson
 
@@ -28,7 +29,7 @@ class PickleStore(object):
     def read(self, mongoose_lib, version, symbol, **kwargs):
         blob = version.get("blob")
         if blob is not None:
-            if blob == _MAGIC_CHUNKED:
+            if blob == _MAGIC_CHUNKED or blob == _MAGIC_CHUNKED_NO_COMPRESS:
                 collection = mongoose_lib.get_top_level_collection()
                 data = b''.join(x['data'] for x in collection.find({'symbol': symbol,
                                                                    'parent': version['_id']},
@@ -36,7 +37,8 @@ class PickleStore(object):
             else:
                 data = blob
             # Backwards compatibility
-            data = lz4.decompress(data)
+            if blob != _MAGIC_CHUNKED_NO_COMPRESS:
+                data = lz4.decompress(data)
             return pickle_compat_load(io.BytesIO(data))
         return version['data']
 
@@ -54,12 +56,16 @@ class PickleStore(object):
         collection = arctic_lib.get_top_level_collection()
         # Try to pickle it. This is best effort
         version['blob'] = _MAGIC_CHUNKED
-        pickled = lz4.compressHC(cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL))
+        try:
+            pickled = lz4.compressHC(cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL))
+        except OverflowError:
+            pickled = cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL)
+            version['blob'] = _MAGIC_CHUNKED_NO_COMPRESS
 
         for i in xrange(int(len(pickled) / _CHUNK_SIZE + 1)):
             segment = {'data': Binary(pickled[i * _CHUNK_SIZE : (i + 1) * _CHUNK_SIZE])}
-            sha = checksum(symbol, segment)
             segment['segment'] = i
+            sha = checksum(symbol, segment)
             collection.update_one({'symbol': symbol, 'sha': sha}, {'$set': segment,
                                                                '$addToSet': {'parent': version['_id']}},
                                        upsert=True)
--- a/arctic/store/_pickle_store.py
+++ b/arctic/store/_pickle_store.py
@@ -3,13 +3,13 @@ from bson.binary import Binary
 from bson.errors import InvalidDocument
 from six.moves import cPickle, xrange
 import io
-import lz4
+from .._compression import compress, decompress
 import pymongo
 
 from arctic.store._version_store_utils import checksum, pickle_compat_load
 
 _MAGIC_CHUNKED = '__chunked__'
-_MAGIC_CHUNKED_NO_COMPRESS = _MAGIC_CHUNKED + 'no_compress__'
+_MAGIC_CHUNKEDV2 = '__chunked__V2'
 _CHUNK_SIZE = 15 * 1024 * 1024  # 15MB
 _MAX_BSON_ENCODE = 256 * 1024  # 256K - don't fill up the version document with encoded bson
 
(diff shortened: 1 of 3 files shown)


Wilfred:
Looks like there's another test that needs updating following these changes.

arctic/store/_pickle_store.py
```diff
@@ -54,12 +56,16 @@ def write(self, arctic_lib, version, symbol, item, previous_version):
...
+            pickled = lz4.compressHC(cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL))
+        except OverflowError:
+            pickled = cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL)
+            version['blob'] = _MAGIC_CHUNKED_NO_COMPRESS
 
         for i in xrange(int(len(pickled) / _CHUNK_SIZE + 1)):
             segment = {'data': Binary(pickled[i * _CHUNK_SIZE : (i + 1) * _CHUNK_SIZE])}
-            sha = checksum(symbol, segment)
             segment['segment'] = i
+            sha = checksum(symbol, segment)
```
Wilfred: This is changing the `sha` I presume? Will this cause a problem for bad data?

arctic/store/_pickle_store.py
```diff
@@ -54,12 +56,16 @@ def write(self, arctic_lib, version, symbol, item, previous_version):
         collection = arctic_lib.get_top_level_collection()
         # Try to pickle it. This is best effort
         version['blob'] = _MAGIC_CHUNKED
-        pickled = lz4.compressHC(cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL))
+        try:
+            pickled = lz4.compressHC(cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL))
+        except OverflowError:
```
Wilfred: Question: What if we move to 64-bit `lz4` in the future? Would we end up breaking backward compatibility? If so, we could check the size of the pickled data against 2GiB -- what do you think?

bmoscon: that shouldnt be a problem. LZ4 bypass will only happen if you get an overflow error. If that DOES happen, you set the type to _MAGIC_CHUNKED_NO_COMPRESS, so when you read it back, it will not decompress. If you move to Lz4x64, you wont have an overflow error, and thus will bypass all this new code.

arctic/store/_pickle_store.py
bmoscon: it does change the sha. I was getting checksum collisions without this change. 

For pickle store, the SHA is not used when reading the data, so it shouldnt matter as far as i can tell. You should still be able to read back old data no matter how the checksum was produced, for read it just looks it up by symbol name

bmoscon: which kind of begs the question, why even compute and store the sha?

Wilfred: If the checksum isn't checked when reading, when do we get collisions?

Another dumb question: Surely checksum collisions are unavoidable? They can be more or less likely, but it should be impossible to avoid them entirely without making checksums that are at least as big as the underlying data.

arctic/store/_pickle_store.py
Wilfred: I was thinking: try to write data with Lz4x64, succeed, but end up with data that cannot be decompressed with Lz4x32.

(Feel free to overrule me if you don't think this is a worthwhile worry.)

arctic/store/_pickle_store.py
bmoscon: we get collisions when we write, so they overwrite already written data. When the read happens, you dont read enough data back and it starts giving errors when trying to unpickle the data. The key it matches on during write is checksum, symbol

arctic/store/_pickle_store.py
bmoscon: if the python lz4 module supported 64 bit compression, I'd try it. It doesnt look like they expose that. feel free to correct me!

Wilfred:
LGTM!

arctic/store/_pickle_store.py
richardbounds: I'm not sure why we don't compress the chunks separately, which I think would solve this problem. Does anyone else wonder whether storing a 2+GB pickled object is actually a good idea anyway?

bmoscon: I don't disagree with you. @TomTaylorLondon insisted they needed this.

I can re-work the code so it always compresses by chunk. I'm not sure how much overhead compressing once is vs compressing multiple  times though. You also get into the situation where the compressed chunks are significantly smaller, leaving strange sized chunks written to mongo (as opposed to the standard 15MB previously). If these concerns arent major, more than happy to change @richardbounds

arctic/store/_pickle_store.py
richardbounds: The checksum is for dedupe, I think. Is it possible that the data chunks really are identical, in which case we don't handle the 'identical repeating chunk' case correctly?

bmoscon: im pretty sure the data chunks really were identical, hence the checksum match. If you look at the pickle code, i dont see anything in there that would handle a "repeating" chunk case

bmoscon: in another note - i feel like this is because we are not compressing -- hence the repetition of data/chunks. If you compress, its very unlikely you'd have repeating chunks. If I make the change suggested below by @richardbounds, this issue goes away

bmoscon: I obviously wasnt thinking -- compressing identical chunks will not make the repetition of checksums go away :(

jamesblackburn: > Another dumb question: Surely checksum collisions are unavoidable? They can be more or less likely, but it should be impossible to avoid them entirely without making checksums that are at least as big as the underlying data.

Note git assumes SHA-1 hashes never collide unless the content is identical. There's been lots of chat on this:
http://lwn.net/Articles/307281/

bmoscon: i dont think there is any question that data chunks are identical -- which is the issue, it needs to write X of those, and each time they overwrite eachother!

jamesblackburn: I think this is the correct fix for the SHA issue - include the segment offset in the checksum.

arctic/store/_pickle_store.py
jamesblackburn: @bmoscon has hit the nail on the head with why we did it this way. The original mmap storage engine fragmented badly, so identically sized chunks were an advantage for the pickle store.  WT doesn't do this, so this is less of an issue.

I think compression buys us a lot, so lets fix this by compressing the chunks as @richardbounds suggests.

Also agree that this verged on *too much* data in a pickled item.  Mongo have always been surprised by how much data we're pushing into the database...  That said I see the argument that there shouldn't be an artificial limit here.

tests/unit/store/test_pickle_store.py
```diff
@@ -32,10 +32,10 @@ def test_write_object():
     PickleStore.write(self, arctic_lib, version, 'sentinel.symbol', sentinel.item, sentinel.previous_version)
     assert 'data' not in version
 
-    assert version['blob'] == '__chunked__'
+    assert version['blob'] == '__chunked__V2'
```
jamesblackburn: Can we have a unit test that tests the `__chunked__` case?

bmoscon: done

bmoscon:
alright, should be all set @jamesblackburn

arctic/store/_pickle_store.py
```diff
@@ -53,13 +58,14 @@ def write(self, arctic_lib, version, symbol, item, previous_version):
...
         # Try to pickle it. This is best effort
-        version['blob'] = _MAGIC_CHUNKED
-        pickled = lz4.compressHC(cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL))
+        version['blob'] = _MAGIC_CHUNKEDV2
+        pickled = cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL)
 
         for i in xrange(int(len(pickled) / _CHUNK_SIZE + 1)):
-            segment = {'data': Binary(pickled[i * _CHUNK_SIZE : (i + 1) * _CHUNK_SIZE])}
-            sha = checksum(symbol, segment)
+            data = compress(pickled[i * _CHUNK_SIZE: (i + 1) * _CHUNK_SIZE])
```
jamesblackburn: This runs the compression serially - it would be better to call `compress_array` with all the chunks, which will run the compression in parallel if the Cython library is available.

jamesblackburn:
We should flag this as an API break. Pickled objects written by this new store will not be loadable by previous versions of the arctic library. 

Should we make the pickle store handle and error on future API changes?

bmoscon:
Oh i see what you mean. We can add that, but of course no way to get it in the older versions :) So really anything we add is kind of pointless, except it will help make it more future proof should we ever change it again

Issue #229, opened 2016-09-20
update picklestore with a warning if trying to read something stored by a future, unsupported version


bmoscon:

bmoscon:
Assume that we create a __chunked__V3 type of chunked data in pickle store, an older version of arctic that tries to read this will fail. We should put something in arctic now that will produce a warning message if an "unreadable" object from mongo is accessed

bmoscon:
fixed

PR #230, opened 2016-09-20
Issue 229

bmoscon:
Take a look @jamesblackburn this should help 'future proof' picklestore with a readable error


bmoscon:
fixes issue #229

Issue #228, opened 2016-09-19
Mongo fail-over during append can leave a Version in an inconsistent state


jamesblackburn:
We can trigger this assertion:

```
...
  File "/app/AHL/packages/ahl.tickdownsample/1.16.0-py2.7/app/AHL/ahl.tickdownsample/lib/python2.7/site-packages/ahl.mongo-1.297.0-py2.7-linux-x86_64.egg/ahl/mongo/mongoose/store/version_store.py", line 105, in append
    return super(VersionStore, self).append(symbol, data, metadata, prune_previous_version, upsert, **kwargs)
  File "/opt/ahl/app/AHL/packages/ahl.tickdownsample/1.16.0-py2.7/app/AHL/ahl.tickdownsample/lib/python2.7/site-packages/arctic-1.4.0-py2.7-linux-x86_64.egg/arctic/decorators.py", line 50, in f_retry
    return f(*args, **kwargs)
  File "/opt/ahl/app/AHL/packages/ahl.tickdownsample/1.16.0-py2.7/app/AHL/ahl.tickdownsample/lib/python2.7/site-packages/arctic-1.4.0-py2.7-linux-x86_64.egg/arctic/store/version_store.py", line 449, in append
    Append not possible - please call write() to get versions back in sync''')
ArcticException: version_nums is out of sync with previous version document.
            This probably means that either a version document write has previously failed, or the previous version has been deleted.
            Append not possible - please call write() to get versions back in sync
```

in append, as the `version_nums` are updated before the data is actually written.

bmoscon:
So whats the fix? Use a bulk update so its all done in the same transaction?

jamesblackburn:
There are no multi-document transactions in mainstream mongo versions...

I think we probably need to be less strict on the append. In this case we know something bad has happened if the version has increments, but there's no accompanying data. 

I think he fix might be to warn about the missing data instead of raise, and allow the append to proceed.

bmoscon:
fixed

PR #232, opened 2016-09-20
Don't raise during append(...) if the previous append didn't succeed.

jamesblackburn:
It can be the case that the previous append fails (which is reported
back to the caller), and subsequent appends won't be allowed until a
read()+write() are called.  Log an error, but allow the new append to
proceed.

Fixes MDP-1054 / Issue #228


bmoscon:
merge it!

jamesblackburn:
@richardbounds I think this fixes the issue?

richardbounds:
Lgtm

Issue #235, opened 2016-09-26
feature: get list of chunk ranges


bmoscon:
would like a way to know what the dates are in the chunkstore without having to read all the data back.

bmoscon:
fixed

Issue #234, opened 2016-09-26
feature: iterator access to chunkstore


bmoscon:
we're storing 20 years+ of data in chunkstore. Would be nice to have an iterator accessor to allow easy linear movement across data without needing to know date ranges for individual chunks.

bmoscon:
fixed

PR #237, opened 2016-09-26
Features: get_chunk_range + iterator access on ChunkStore

bmoscon:
Addresses #234 and #235 

@jamesblackburn

### DIFF
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -448,7 +448,7 @@ class ChunkStore(object):
 
     def get_chunk_ranges(self, symbol, chunk_range=None, reverse=False):
         """
-        Returns a list of (Start, End) tuples for each chunk in the symbol
+        Returns a generator of (Start, End) tuples for each chunk in the symbol
 
         Parameters
         ----------
@@ -473,7 +473,7 @@ class ChunkStore(object):
 
     def start_iterator(self, symbol, chunk_range=None):
         """
-        Returns a generator that access each chunk in ascending order
+        Returns a generator that accesses each chunk in ascending order
 
         Parameters
         ----------
@@ -494,7 +494,7 @@ class ChunkStore(object):
 
     def end_iterator(self, symbol, chunk_range=None):
         """
-        Returns a generator that access each chunk in descending order
+        Returns a generator that accesses each chunk in descending order
 
         Parameters
         ----------
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -429,9 +429,86 @@ class ChunkStore(object):
             self.__update(sym, item, combine_method=self.__replace, chunk_range=chunk_range)
 
     def get_info(self, symbol):
+        """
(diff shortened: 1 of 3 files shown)


arctic/chunkstore/chunkstore.py
```diff
@@ -429,9 +429,86 @@ def update(self, symbol, item, chunk_range=None, upsert=False, **kwargs):
...
+        c = CHUNKER_MAP[sym[CHUNKER]]
+
+        spec = {SYMBOL: symbol}
+        if chunk_range:
+            spec.update(CHUNKER_MAP[sym[CHUNKER]].to_mongo(chunk_range))
+
+        for x in self._collection.find(spec, projection=[START, END], sort=[(START, pymongo.ASCENDING if not reverse else pymongo.DESCENDING)]):
+            yield (c.chunk_to_str(x[START]), c.chunk_to_str(x[END]))
+
+    def start_iterator(self, symbol, chunk_range=None):
```
jamesblackburn: maybe `iterator` and `reverse_iterator`  - to match Python and C++ naming?  Alternatively, given this just calls `read` could make it parameters on the read method?

bmoscon: I'm fine making it a param on read. How would you suggest doing that? with an iterator keyword arg? How would you differentiate between forwards and backwards?  use a string?

arctic/chunkstore/chunkstore.py
```diff
@@ -429,9 +429,86 @@ def update(self, symbol, item, chunk_range=None, upsert=False, **kwargs):
...
+        if not sym:
+            raise NoDataFoundException("Symbol does not exist.")
+        c = CHUNKER_MAP[sym[CHUNKER]]
+
+        spec = {SYMBOL: symbol}
+        if chunk_range:
+            spec.update(CHUNKER_MAP[sym[CHUNKER]].to_mongo(chunk_range))
+
+        for x in self._collection.find(spec, projection=[START, END], sort=[(START, pymongo.ASCENDING if not reverse else pymongo.DESCENDING)]):
+            yield (c.chunk_to_str(x[START]), c.chunk_to_str(x[END]))
```
jamesblackburn: cursors timeout after 10mins - so depending on how long the clients take to process the data, may take some time...

bmoscon: you can turn the timeout off on the find method. Are you opposed to this?

arctic/chunkstore/chunkstore.py
jamesblackburn: Or maybe tweak the name - the `start` and `end` sounds like actions rather than a direction.

arctic/chunkstore/chunkstore.py
jamesblackburn: maybe it doesn't matter for the first cut.  Think we shouldn't risk leaving cursors hanging forever in the DB...  Could also yield batches of chunk ranges but that might be overkill

bmoscon: yeah i think its ok to just capture them all at once and work on it that way

Issue #236, opened 2016-09-26
Cython routines do not handle errors from LZ4


bmoscon:

bmoscon:
fixed

PR #238, opened 2016-09-26
Fix for decompression issue

bmoscon:
Fix for issue #236

### DIFF
--- a/src/_compress.pyx
+++ b/src/_compress.pyx
@@ -233,8 +233,7 @@ def decompressarr(pStrList):
             result = <char*>malloc(original_size)
             # decompress
             if LZ4_decompress_safe(cString + hdr_size, result, compressed_size - hdr_size, original_size) < 0:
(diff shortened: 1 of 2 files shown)


bmoscon:
@jamesblackburn

src/_compress.pyx
```diff
@@ -229,7 +232,9 @@ def decompressarr(pStrList):
             # malloc 
             result = <char*>malloc(original_size)
             # decompress
-            LZ4_decompress_safe(cString + hdr_size, result, compressed_size - hdr_size, original_size)
+            if LZ4_decompress_safe(cString + hdr_size, result, compressed_size - hdr_size, original_size) < 0:
+                with gil:
```
jamesblackburn: Do you need to gil this here?  It's just a local int above, isn't it?

bmoscon: it wont compile without it

Issue #248, opened 2016-09-30
Chunkstore open ranged read fails if 'date' is not in index


bmoscon:
If there is no datetime index, we cannot slice the dataframe and use a boolean mask to filter the data. This mask fails if the DateRange end is None.

bmoscon:
merged

PR #249, opened 2016-09-30
Chunkstore open range read

bmoscon:
see issue #248 

@jamesblackburn

### DIFF
--- a/arctic/chunkstore/date_chunker.py
+++ b/arctic/chunkstore/date_chunker.py
@@ -92,10 +92,14 @@ class DateChunker(Chunker):
         if 'date' in data.index.names:
             return data[range_obj.start:range_obj.end]
         elif 'date' in data.columns:
-            if range_obj.end:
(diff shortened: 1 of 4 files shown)


arctic/chunkstore/date_chunker.py
```diff
@@ -92,7 +92,14 @@ def filter(self, data, range_obj):
         if 'date' in data.index.names:
             return data[range_obj.start:range_obj.end]
         elif 'date' in data.columns:
-            return data[(data.date >= range_obj.start) & (data.date <= range_obj.end)]
+            if range_obj.start and range_obj.end:
```
jamesblackburn: The `DateRange` objects support `OPEN` vs `CLOSED` time bounds.  We make some effort to get this right here:
https://github.com/manahl/arctic/blob/master/arctic/date/_util.py#L112

Would be nice to respect that in ChunkStore.

PR #251, opened 2016-10-03
update docstrings in chunkstore

bmoscon:


tests/integration/chunkstore/test_chunkstore.py
```diff
@@ -537,7 +537,7 @@ def write_random_data(chunkstore_lib, name, month, days, securities, chunk_size=
     month: integer
     days: list of integers
     securities: list of integers
-    chunk_size: one of 'D', 'M', 'Y'
+    chunk_size: one of 'D', 'M', 'A'
```
jamesblackburn: Why change this?  I would have assumed `Y`?

bmoscon: Deprecated now, only D, M, and A (for annual).

jamesblackburn: Yeah, but I think people would guess 'DMY', rather than 'DMA'.  Nearly everywhere else we talk about years rather than annos.

bmoscon: I guess we can check for 'Y' and convert to A, but pandas only takes 'A' now. Might be moot if i have to re-write the chunker anyways :)

jamesblackburn: Oh I see it's a pandas change.  How horrid.

bmoscon: I changed the code to force 'A' always. Don't want to write data with 'Y' chunksize in Mongo doc, then later read it out with new version of pandas and fail, so now chunksize will be 'A' if user chooses 'Y'

bmoscon:
fixes #252

PR #262, opened 2016-10-20
Mdp 1293 append safety

jamesblackburn:
A few changes here to ensure that append doesn't corrupt data in cases were find returns an incomplete set of chunks.

This might occur if a concurrent writer is updating / pruning the chunks, or during a failover if we're unlucky enough to query the new primary before all oplog entries have been applied to the indexes.  

The below adds additional checks before the append is committed


jamesblackburn:
@richardbounds @bmoscon please take a look

bmoscon:
some test failures

jamesblackburn:
There's definitely one interesting test failure - my append after a failed append test... Digging.

bmoscon:
circle is failing for a known reason, but travis should pass. . .

arctic/store/_ndarray_store.py
```diff
@@ -321,9 +321,27 @@ def _concat_and_rewrite(self, collection, version, symbol, item, previous_versio
...
+        for segment in collection.find(spec, projection={'_id':1,
+                                                         'segment':1,
+                                                         'compressed': 1
+                                                         },
+                                       sort=[('segment', pymongo.ASCENDING)]):
+            # We want to stop iterating when we find the first uncompressed chunks
+            if not segment['compressed']:
+                # We include the last chunk in the recompression
+                unchanged_segment_ids.pop()
+                break
```
jamesblackburn: @richardbounds @bmoscon This is the key bit - when we hit the first `uncompressed` segment, we want to concat and repack the subsequent appended chunks to the previous compressed segment.

bmoscon:
looks good, passed travis so probably ok to merge.

arctic/store/_ndarray_store.py
jamesblackburn: I've added unit tests for this, and there's a test that asserts segment counts:
https://github.com/manahl/arctic/blob/master/tests/integration/store/test_ndarray_store.py#L200

PR #263, opened 2016-10-21
Mongo quota improvements

jamesblackburn:
- Add Library names to the quota waring and info lines
- WARN when library >90% full 

Currently we only warn when we're within 100 (average sized chunks) f
being full.  In addition WARN when the library is >90% (the warning may
still be infrequent if we're many chunks away from filling the library).


bmoscon:
merge it. Circle is failing for a known reason. as long as travis passes we're good.

PR #267, opened 2016-10-28
TickStore.delete: Return result of deletion

mckelvin:
I want to know if anything is deleted, since the `delete` action will only delete fully contained in the passed in date_range.


jamesblackburn:
If you can, please add a unit test :)

mckelvin:
@jamesblackburn There're already some deletion related code in `tests/integration/tickstore/test_ts_delete.py`. So I added the test case on that. LGTU ?

bmoscon:
I'd prefer that you get back what you actually think you're deleting, but I think its fine as is (we're pretty much relying on pymongo to return those values to us, so we don't necessarily need to test them).

PR #269, opened 2016-10-28
Complete unpin of dependencies

bmoscon:


bmoscon:
@jamesblackburn you can take a look if you want, but once it passes im just going to merge it in since it doesnt change any code, just config files

PR #277, opened 2016-11-04
Backout Pandas 0.19.0 workaround

bmoscon:
0.19.1 fixes the issue in 0.19.0, so its time to backout the hack I put in to workaround the issue. 

see: https://github.com/pandas-dev/pandas/issues/14338


bmoscon:
@jamesblackburn

bmoscon:
with more extensive testing, the "fix" in 0.19.1 in pandas only slightly improves the situation. The workaround i have in place in Arctic is about 3X faster :'(

PR #282, opened 2016-11-08
Update Travis Build Parameters

bmoscon:
- more modern linux distro
- newer version of mongo

This seems to fix the incessant mongo connection errors that have been plaguing the builds. 

@jamesblackburn


jamesblackburn:
Does this slow down every build?

PR #285, opened 2016-11-09
update init file in chunkstore

bmoscon:
@jamesblackburn


arctic/chunkstore/__init__.py
```diff
@@ -0,0 +1,3 @@
+from .chunkstore import ChunkStore
```
jamesblackburn: Ideally we shouldn't import modules from `__init__` in new packages.  The main issue is performance.   Doing this causes the python interpreter to parse and run the module scope of these modules on every import which is bad for import performance.

Issue #292, opened 2016-11-28
ChunkStore writes fail when dataframe metadata is very large


bmoscon:
#### Arctic Version

```
# All w/ ChunkStore
```

#### Arctic Store

```
# ChunkStore
```

#### Description of problem and/or code sample that reproduces the issue

Dataframes with a lot of metadata (for Chunkstore this means columns with strings that have missing values) that are very large will not chunk by size properly because the size of the metadata is not accounted for during the write.

bmoscon:
fixed

PR #293, opened 2016-11-28
Fix issue #292

bmoscon:


bmoscon:
take a look @jamesblackburn

arctic/chunkstore/chunkstore.py
```diff
@@ -283,10 +284,13 @@ def write(self, symbol, item, chunker=DateChunker(), **kwargs):
             doc[METADATA] = {'columns': data[METADATA][COLUMNS] if COLUMNS in data[METADATA] else ''}
```
bmoscon: I could do that. It would make it more robust going forward, but would not be backwards compatible. Might be worth thinking about. A breaking change now would be feasible, but months down the road will not be.

PR #295, opened 2016-11-29
Move chunkstores per-chunk metadata to its own collection

bmoscon:
This change will help future proof the storage of metadata, and allow us to store per symbol user defined metadata (not implemented yet). Also prevents a vert large sized metadata from drastically reducing the chunking size of every part of a chunk for chunks over 16MB.

### DIFF
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -65,6 +65,10 @@ class ChunkStore(object):
                                        (START, pymongo.ASCENDING),
                                        (SEGMENT, pymongo.ASCENDING)],
                                       unique=True, background=True)
+        self._mdata.create_index([(SYMBOL, pymongo.ASCENDING),
+                                  (START, pymongo.ASCENDING),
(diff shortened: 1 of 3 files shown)


bmoscon:
@jamesblackburn

arctic/chunkstore/chunkstore.py
```diff
@@ -65,6 +65,10 @@ def _ensure_index(self):
                                        (START, pymongo.ASCENDING),
                                        (SEGMENT, pymongo.ASCENDING)],
                                       unique=True, background=True)
+        self._mdata.create_index([(SYMBOL, pymongo.ASCENDING),
+                                  (START, pymongo.ASCENDING),
+                                  (END, pymongo.ASCENDING)],
```
jamesblackburn: How come you have the `START` `END` on the metadata, doesn't the metadata apply to the whole symbol, rather than a range?

bmoscon: no, its per chunk

PR #297, opened 2016-11-29
Decompression fails with empty string

bmoscon:
Bug in the Cython code considered a return code of 0 to mean the decompression failed. 0 is a valid code when the size of the data is zero (i.e. emptry string).


bmoscon:
@jamesblackburn

PR #298, opened 2016-12-01
Tweak the enable_sharding command

jamesblackburn:
- hashed based sharding by default
- be less specific about the database already sharded exception from
mongo

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,9 @@
 ## Changelog
 
+### 1.36
+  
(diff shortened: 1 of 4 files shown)


bmoscon:
looks like some unit tests need to be updated

bmoscon:
also not sure why, but circle is failing because its still using pytest-dbfixtures. i cant find any reference to it in the setup.py, nor in the circle config, so not sure why its installing that .....

PR #303, opened 2016-12-19
Cache for mktz

mckelvin:
My app use `ms_to_datetime` frequently, after profiling using [pyflame](https://github.com/uber/pyflame), I found the function takes `35.62%` of the total CPU time, and the bottleneck is in `mktz`(`31.87%`). I think it's better to cache results for the `mktz` function.

### DIFF
--- a/arctic/date/_mktz.py
+++ b/arctic/date/_mktz.py
@@ -3,6 +3,10 @@ import os
 import dateutil
 import tzlocal
 import six
+import weakref
+
+
+_TZ_CACHE = weakref.WeakValueDictionary()
 
 
 class TimezoneError(Exception):
@@ -34,6 +38,11 @@ def mktz(zone=None):
     if zone is None:
         zone = tzlocal.get_localzone().zone
     zone = six.u(zone)
+
+    cached = _TZ_CACHE.get(zone)
+    if cached is not None:
+        return cached
+
     tz = dateutil.tz.gettz(zone)
     if not tz:
         raise TimezoneError('Timezone "%s" can not be read' % (zone))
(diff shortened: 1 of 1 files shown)


bmoscon:
looks good and passes tests (circle just looks like its broken because coveralls wont work on a fork). A few questions:

1. Whats the performance improvement from using this? 
2. Any way we can add new unit tests to test the new functionality?

bmoscon:
@jamesblackburn

mckelvin:
> 1. Whats the performance improvement from using this?

I don't profile in production yet, but there're some result in ipython:

```python
In [1]: from arctic.date import mktz

In [2]: timeit mktz()
10000 loops, best of 3: 125 µs per loop

In [4]: from weakref import WeakValueDictionary

In [5]: cache_dict = WeakValueDictionary()

In [6]: cache_dict["Asia/Shanghai"] = mktz()

In [7]: timeit cache_dict.get("Asia/Shanghai")
1000000 loops, best of 3: 1.1 µs per loop
```

I followed the `mktz()` using strace and found it will read files from disk:`/etc/localtime`, `/usr/share/zoneinfo/Asia/Shanghai`, which should be blame for the slow performance.

> 2. Any way we can add new unit tests to test the new functionality?

I'm not sure if test case like this is enough, I'll amend in the PR if that's enough:
```python
def test_mktz_cache():
    with patch('arctic.date._mktz._TZ_CACHE') as tz_cache:
        tz_cache.clear()
        assert len(tz_cache) == 0
        r1 = mktz("UTC")
        assert len(tz_cache) == 1
        r2 = mktz("UTC")
        assert len(tz_cache) == 1
        assert r1 == r2
```

jamesblackburn:
Looks good.  We `@memoize` this internally. Definitely worth caching.

jamesblackburn:
@mckelvin worth turning this into a bug so we can track it

PR #304, opened 2016-12-20
Update travis.yml to use container build infrastructure again

bmoscon:
With most recent updates to the server test fixtures (pytest-server-fixtures) we
should be able to run with sudo: false again, thus enabling containers again. This
will result in a small but noticeable speedup in build times


bmoscon:
will squash the commits

bmoscon:
@jamesblackburn

PR #307, opened 2016-12-22
Revert "Cache for mktz #303"

jamesblackburn:
Sharing tzfile objects doesn't appear to be safe particularly on DST
change:
https://github.com/dateutil/dateutil/issues/338

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,7 +2,6 @@
 
 ### 1.37
   * Bugfix:  #300 to_datetime deprecated in pandas, use to_pydatetime instead
-  * Feature: #303 Cache timezones in mktz for improved performance
 
 ### 1.36 (2016-12-13)
   
--- a/arctic/date/_mktz.py
+++ b/arctic/date/_mktz.py
@@ -3,10 +3,6 @@ import os
 import dateutil
 import tzlocal
 import six
-import weakref
-
-
-_TZ_CACHE = weakref.WeakValueDictionary()
 
 
 class TimezoneError(Exception):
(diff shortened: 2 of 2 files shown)


jamesblackburn:
@bmoscon

bmoscon:
looks good, how much of a perf hit is the copy? If this any better/slower than it was originally (before caching) ?

bmoscon:
I also assume you ran the unit tests locally since this only seems to break internally and not on circle/travis? :D

jamesblackburn:
Hrm I see this caching making almost no difference:

Before
```
%timeit mktz()
The slowest run took 6.03 times longer than the fastest. This could mean that an intermediate result is being cached 
1000 loops, best of 3: 178 µs per loop
%timeit mktz('Europe/London')
The slowest run took 7.54 times longer than the fastest. This could mean that an intermediate result is being cached 
1000 loops, best of 3: 176 µs per loop
```

After:

```
%timeit mktz('Europe/London')
1000 loops, best of 3: 192 µs per loop
%timeit mktz()
The slowest run took 4.62 times longer than the fastest. This could mean that an intermediate result is being cached 
1000 loops, best of 3: 198 µs per loop
```

Also:

```
%timeit copy.copy(tz)
The slowest run took 6.09 times longer than the fastest. This could mean that an intermediate result is being cached 
1000 loops, best of 3: 175 µs per loop
%timeit gettz('Europe/London')
The slowest run took 6.39 times longer than the fastest. This could mean that an intermediate result is being cached 
1000 loops, best of 3: 171 µs per loop
```

bmoscon:
so, @jamesblackburn think we should just back it out? Or do we only need to make the copy when storing the original tzobject into the cache?

jamesblackburn:
I think we have no option but to backout.

jamesblackburn:
@mckelvin we're reverting #303 as this issue causes strange behaviour around DST:
dateutil/dateutil#338

Issue #313, opened 2017-01-12
set and read user defined metadata


bmoscon:
#### Arctic Version

```
# latest
```

#### Arctic Store

```
# ChunkStore
```

Need to be able to set and read per symbol metadata in Chunkstore

bmoscon:
fixed

PR #314, opened 2017-01-12
Issue #313 User Defined Metadata

bmoscon:
Set and read user defined per-symbol metadata

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -3,6 +3,7 @@
 ### 1.37
   * Bugfix:  #300 to_datetime deprecated in pandas, use to_pydatetime instead
   * Bugfix:  #309 formatting change for DateRange ```__str__```
+  * Feature: #313 set and read user specified metadata in chunkstore
 
 ### 1.36 (2016-12-13)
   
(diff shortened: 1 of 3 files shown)


bmoscon:
@jamesblackburn take a look

bmoscon:
@jlou2u

jlou2u:
worth adding a test that hits the segmenting code to make sure it works as we'd like?

jlou2u:
What about adding a test with unserializable data in the user metadata.  The intent would be to make sure the cause of exception is clear and understandable for a user.

jamesblackburn:
Do you really need metadata per chunk rather than metadata per symbol?  In VersionStore (and in general) we store metadata per symbol - I'm not sure how useful it is to have this per chunk...

bmoscon:
@jlou2u metadata collection is per chunk, not per segment, so no reason to test that. would be good to put in some clearer exceptions around bad metadata.

bmoscon:
@jamesblackburn i suppose we really want a way to do both, but currently only have a need for per chunk

bmoscon:
@jamesblackburn spoke with justin, we're going to change this around a bit, so I'll let you know when its ready

PR #319, opened 2017-01-26
Support for audit logs in ChunkStore

bmoscon:
Enable audit logs on operations with the audit keyword arg. Writes supplied dict along with additional information into the audit collection


bmoscon:
see #320

bmoscon:
@jamesblackburn

jamesblackburn:
Auditing in the version store provides the ability to preserve the data before and after the change.  It looks like here there's an audit log, but no actual history of the change being made?

bmoscon:
Version store is a different case - the data is never "gone" we mainly want to see what operations took place and what the state was at that time (chunks, appended rows, etc)

Issue #216, opened 2016-09-10
Tickstore can't save dataframes if the index column has a name.


jonasleotech:
When reading data using read_csv (or other scenarios when we use an existing named columns as the index) and using index_col to assign the index from one of the columns in the data set, the index column will have a name assigned.

The tickstore currently does not support it as once it converts the dataframe to records it always assumes the index will be converted to a dict key "index".

The code below reproduces the issue.

``` python
from arctic import Arctic
import pandas as pd
import numpy as np
import pytz

store = Arctic("127.0.0.1")
store.initialize_library("ticktest","TickStoreV3")
ticktest = store["ticktest"]

rows = 10
df = pd.DataFrame({
        'floats':np.random.random(rows)},
                  index=pd.date_range(start='2011-01-01', freq="S", periods=rows, tz=pytz.utc))
df.index.names= ["price_timestamp"]
print(df.info())

ticktest.delete("SYMBOL2")

# This will raise the error ValueError: no field of name index
ticktest.write("SYMBOL2", df)
```

bmoscon:
fixed

PR #321, opened 2017-01-26
Fix issue 216

bmoscon:
Fixes issues #216. This issue already addressed by PR #217, but author has not responded to requests for unit tests, so I've added one to this PR


bmoscon:
take a look @jamesblackburn

PR #344, opened 2017-04-12
Make concurrent appends safe

richardbounds:
Make concurrent appends safe(r) by introduce a 'dirty_append' kwarg to the append() method. 

Any time an append may not be completely sequential with the previous version, it is flagged as 'dirty' and we make it multiprocess safe by concatenating and rewriting any previously appended chunks.

### DIFF
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -211,7 +211,7 @@ class NdarrayStore(object):
         rtn = np.dtype(rtn, metadata=dict(dtype.metadata or {}))
         return rtn
 
-    def append(self, arctic_lib, version, symbol, item, previous_version, dtype=None):
+    def append(self, arctic_lib, version, symbol, item, previous_version, dtype=None, dirty_append=True):
         collection = arctic_lib.get_top_level_collection()
         if previous_version.get('shape', [-1]) != [-1, ] + list(item.shape)[1:]:
             raise UnhandledDtypeException()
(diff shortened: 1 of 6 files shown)


tests/integration/test_concurrent_append.py
```diff
@@ -0,0 +1,81 @@
...
+    def run_append(end):
+        app_1 = Appender(mongo_host, library_name, sem, 0, end)
+        proc = Process(target=app_1.run)
+        proc.start()
+        sem.release()
+        return proc
+
+    def check_written():
+        sym = library.read('symbol')
+        print "Checking written %d" % len(sym.data)
```
bmoscon: can we make this python3 compatible?

tests/integration/test_concurrent_append.py
```diff
@@ -0,0 +1,81 @@
...
+
+    # time how long it takes to do an append operation
+    start = datetime.now()
+    proc = run_append(1)
+    proc.join()
+    check_written()
+    time_taken = (datetime.now() - start).total_seconds()
+
+    for i in xrange(100):
+        print "Loop %d" % i
```
bmoscon: if this is needed, same as above

arctic/store/_ndarray_store.py
```diff
@@ -404,8 +405,9 @@ def write(self, arctic_lib, version, symbol, item, previous_version, dtype=None)
         if previous_version:
             if 'sha' in previous_version \
                     and self.checksum(item[:previous_version['up_to']]) == previous_version['sha']:
-                #The first n rows are identical to the previous version, so just append.
-                self._do_append(collection, version, symbol, item[previous_version['up_to']:], previous_version)
+                # The first n rows are identical to the previous version, so just append.
+                # Do a 'dirty' append (i.e. concat & start from a new base version) for safety
+                self._do_append(collection, version, symbol, item[previous_version['up_to']:], previous_version, dirty_append=True)
```
jamesblackburn: How is the `concat_and_rewrite` this indirectly calls different to a plain `write`?  Should we just delete lines 405-411

richardbounds: Yup, that doesn't seem unreasonable. I think the only difference is that it just concats any previous appends, rather than the whole thing, but I've got no evidence that is worthwhile optimisation.

Issue #346, opened 2017-04-25
VersionStore date_range logic fails to return chunks beyond first instance of end_date


TomTaylorLondon:
When inserting large dataframes with duplicate datetimes (e.g. MultiIndex panels) date_range may fail to return the correct data.

Issue is with: https://github.com/manahl/arctic/blob/master/arctic/store/_pandas_ndarray_store.py#L87

Should be:
```
idxend = min(np.searchsorted(dts, end, side='right'), len(dts) - 1)
```

bmoscon:
@TomTaylorLondon do you want me to fix this, or are you going to push your own fix?

bmoscon:
also, @TomTaylorLondon it would be good if you could give me an example of the dataframe that causes the issue so I can add a unit test :)

bmoscon:
FYI Tom I have a branch with the fix. I'll open a PR tomorrow.

bmoscon:
fixed

PR #347, opened 2017-04-26
Fix issue #346

bmoscon:
fixes issue #346 

@TomTaylorLondon let me know what you think. Would still like to have an example dataframe for a unit test :)


TomTaylorLondon:
Great thanks! I think a large one with dups in the index should suffice

```
index = [dt(2017,1,1)]*10000 + [dt(2017,1,2)]*10000
data = np.random.random((20000, 10))
pd.DataFrame(index=index, data=data)
```

bmoscon:
thanks @TomTaylorLondon. Had to tweak the sizes, but confirmed that without fix the issue is very glaring, and with fix it works as expected. Thanks! Will merge once unit tests pass

bmoscon:
@TomTaylorLondon unit tests pass on 2.7 and a 3.x build, so I'm merging

PR #349, opened 2017-05-02
If cython lz4 cannot import, disable parallel compression

bmoscon:
If we can't import the parallel/cython lz4, we should disable the PARALLEL flag.


bmoscon:
@jamesblackburn

PR #351, opened 2017-05-10
$size queries can't use indexes, use alternative queries.

anglerud:
We've got queries that take over 200 seconds, this makes them much
faster (milliseconds - 5 order of magnitude speedup, yay!).


bmoscon:
also, ignore those errors in travis, they are from pandas 0.20.0

pablojim:
LGTM

PR #353, opened 2017-05-12
Fix tslib issue with new version of pandas

bmoscon:
New versions of pandas have relocated tslib to _libs. This change supports previous versions and new versions of pandas.

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -343,7 +343,7 @@ class TickStore(object):
             rtn = rtn.sort_index(kind='mergesort')
         if date_range:
             # FIXME: support DateRange.interval...
-            rtn = rtn[date_range.start:date_range.end]
+            rtn = rtn.loc[date_range.start:date_range.end]
         return rtn
 
     def _pad_and_fix_dtypes(self, cols, column_dtypes):
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -343,7 +343,7 @@ class TickStore(object):
             rtn = rtn.sort_index(kind='mergesort')
         if date_range:
             # FIXME: support DateRange.interval...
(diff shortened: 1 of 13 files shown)


bmoscon:
hold off on this one, apparently pandas 0.20.x deprecated way more than I realized, so a lot of code will need to be updated :(

bmoscon:
@jamesblackburn ok take a look now. Its not ideal, but some things can only be worked around in so many ways. Panels, ix, sort(), etc are all deprecated. Some are easy to replace, some are not.

arctic/tickstore/tickstore.py
```diff
@@ -343,7 +343,7 @@ def read(self, symbol, date_range=None, columns=None, include_images=False, allo
             rtn = rtn.sort_index(kind='mergesort')
         if date_range:
             # FIXME: support DateRange.interval...
-            rtn = rtn.ix[date_range.start:date_range.end]
+            rtn = rtn[date_range.start:date_range.end]
```
jamesblackburn: Should this be `.iloc` or `.loc`?

bmoscon: it could be iloc, but start and end are dates (datetime, timestamp, doesnt matter which) so slicing works as well. Note I also made the same change in arctic_copy_data.py in scripts. Happy to change it to iloc if you think thats preferable or if there is some problem with datetime slicing I'm not aware of.

jamesblackburn: As pandas slicing is quite overloaded was wondering if it should be more specific to the particular index.

Not opinionated though, as long as it works and perf is good

bmoscon:
@jamesblackburn i changed the indexer to loc. Tests pass except on master because of the broken lz4 stuff (grrr). I'll probably revert that tonight and then merge this in once master is fixed

bmoscon:
@jamesblackburn going to merge with the datetime slicing changed to loc indexing. Let me know if you decide you want something different!

PR #355, opened 2017-05-13
Revert lz4 pinning

bmoscon:
When version 0.9.0 of lz4 was released, it broke arctic, so I pinned arctic to 0.8.2. This broke things internally so It was changed to <= 0.8.2. Now that lz4 latest works with arctic, we should move back to unpinning the version (as it originally was).


bmoscon:
@jamesblackburn

PR #358, opened 2017-05-18
Revert LZ4 pin

bmoscon:
revert pin plus fix an instance where a numpy array was not being converted to string/bytes (not sure how it worked before, but it doesnt work with the new version of lz4 since it checks types)


bmoscon:
take a look @jamesblackburn

Issue #360, opened 2017-05-24
VersionStore append function using version object instead of integer in versionnedItem


karimaIch:
#### Arctic Store

```
VersionStore
```

Hi Guys, apparently there might be an error in your VersionStore append function code: 
at line 68, your are returning a versionnedItem with as a version's attribute a version object and not a version number. I guess there must be version=previous_version["version"] instead? 

I would appreciate your clarification guys, because I was assuming that version attribute in any versionnedItem is an int. 
thanks in advance.

bmoscon:
@karimaIch this is line 68:

https://github.com/manahl/arctic/blob/master/arctic/store/version_store.py#L68

Thats not what you meant, so please either link to the line, or give me a filename as well

karimaIch:
@bmoscon sorry it was a mistake when i typed the line number, i meant line 468 , here is the link: https://github.com/manahl/arctic/blob/master/arctic/store/version_store.py#L468 
thank you in advance.

bmoscon:
fix is in PR now, thanks for raising this

bmoscon:
merged

PR #361, opened 2017-05-25
Fix issue #360

bmoscon:
On empty appends the data returned has the incorrect version, this fixes that issue


bmoscon:
take a look @jamesblackburn

PR #364, opened 2017-05-26
Remove dependency on Python-LZ4 and use internal Arctic LZ4 Library

bmoscon:


bmoscon:
will squash the merges

TomTaylorLondon:
LGMT. It's good to have a single LZ4 implementation, a shame though that we have to include it in the source (and require the user to compile it).

bmoscon:
I assume the benefits of parallelization outweigh the cons of requiring the compile. I have been building and uploading eggs to pypi...so that should help!

TomTaylorLondon:
Sounds good!

setup.py
```diff
@@ -109,6 +108,7 @@ def run_tests(self):
                    "pytest-server-fixtures",
                    "pytest-timeout",
                    "pytest-xdist",
+                   "lz4"
```
jamesblackburn: Presumably this is now just a test_requires?

tests/unit/test_compression.py
```diff
@@ -51,26 +50,6 @@ def test_compression_equal_regardless_parallel_mode():
     assert serial == parallel
 
 
-def test_compress_decompress_no_parallel():
```
jamesblackburn: If these tests are removed, do we still have an external lz4 dependency?

setup.py
bmoscon: yes, there is a test that compares base LZ4 compression vs our compression. Its a performance test. I assume we still want this? If not I can excise it and remove LZ4. We should have it as a separate test that you'd have to manually run, not one that runs every time?

tests/unit/test_compression.py
bmoscon: yes, see previous comment

bmoscon:
@jamesblackburn only added one tostring(), it was just missed before when we converted to Python3. Somehow it didnt break lz4, but i think thats just because we got lucky and the underlying datastructure in numpy was just a byte array. We could also change all the tostrings() to tobytes() which might be slightly faster, but since its a bytearray already, converting between strings and bytes should be very low overhead.

bmoscon:
@jamesblackburn i checked the numpy source code, to strings just calls tobytes, so I can change all the tostrings to tobytes() to save that single function call overhead. Let me know, otherwise I'll merge this later tonight

tests/unit/test_compression.py
bmoscon: here are the tests that still use it:

https://github.com/manahl/arctic/blob/master/tests/unit/test_compress.py

https://github.com/manahl/arctic/blob/master/tests/integration/test_compress_integration.py

I think there is value in keeping these somewhere, but we could move them outside the 'test' directory if desired.

jamesblackburn:
I think it's fine @bmoscon

PR #368, opened 2017-06-01
Arctic needs six as a dependency

Frankkkkk:


bmoscon:
@jamesblackburn I'll probably merge this on Monday unless you can think of a reason not to.

PR #371, opened 2017-06-08
Expose more calls in the BSONStore

achamayou:
I'd like to use a few more calls for the project I'm currently working on. I've also aligned the names of arguments with the pymongo API for better consistency.

### DIFF
--- a/arctic/store/bson_store.py
+++ b/arctic/store/bson_store.py
@@ -71,66 +71,65 @@ class BSONStore(object):
         self._arctic_lib.check_quota()
         return self._collection.insert_many(documents, **kwargs)
 
-    @mongo_retry
-    def delete_one(self, filter_, **kwargs):
+    def delete_one(self, filter, **kwargs):
         """
         See http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.delete_one
         """
-        return self._collection.delete_one(filter_, **kwargs)
+        return self._collection.delete_one(filter, **kwargs)
 
     @mongo_retry
-    def delete_many(self, filter_, **kwargs):
+    def delete_many(self, filter, **kwargs):
         """
         See http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.delete_many
         """
-        return self._collection.delete_many(filter_, **kwargs)
+        return self._collection.delete_many(filter, **kwargs)
 
     @mongo_retry
-    def update_one(self, filter_, update, **kwargs):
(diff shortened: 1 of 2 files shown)


Wilfred:
Tests are failing on Python 3.6 on Travis. It shows `tests/integration/tickstore/test_ts_write.py::test_ts_write_pandas[tickstore] FAILED` then later the test hangs and Travis kills it after 10 minutes without output.

achamayou:
Yes I saw, but there are no details, and it doesn't seem likely to be related to the change. Is there a way to restart the tests?

arctic/store/bson_store.py
```diff
@@ -56,36 +56,90 @@ def find(self, *args, **kwargs):
...
         """
         See http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.insert_many
         """
         self._arctic_lib.check_quota()
-        return self._collection.insert_many(values)
+        return self._collection.insert_many(documents, **kwargs)
 
     @mongo_retry
-    def delete_one(self, query):
+    def delete_one(self, filter_, **kwargs):
```
richardbounds: Not your change, but do we want `@mongo_retry` here? Isn't it possible that we might accidentally delete more than one if we retry?

achamayou: That's a good point, I will look into the details of mongo_retry. I would hope it only catches exceptions that mean the request didn't go through...

jamesblackburn: It will do the wrong thing if multiple documents match the query filter. The exceptions can be raised on network issues if the write isn't acknowledged, but is still processed by mongo, you can end up deleting multiple docs.

jamesblackburn: Just `filter` to match the underlying API? Otherwise you can end up in a funny state whereby the kwargs contains filter but you're massing it in as the first arg

achamayou: Ok, I will remove the retry on delete_one. I'd used filter_ to avoid shadowing the built-in, but you're right, it's bad idea with kwargs present, I will fix that too.

bmoscon:
@achamayou @Wilfred yes something about travis and pytest server fixtures causes intermittent hangs. I've never ever had them when running locally. you can restart by going into travis and restarting the selected run

achamayou:
@bmoscon yes, thank you, I've found the restart button and I can get them green after a few (mongo_)retries.

bmoscon:
@achamayou they dont all need to be green. If they're passing on circle and one of the 2.7s is passing and one of the 3.X's is passing, its usually fine to merge.

PR #374, opened 2017-06-12
Allow sharding on keys other than symbol, default BSONStore to shard on _id

achamayou:
BSONStore instances generally don't have symbols (the idea is that they should be free-form), but we still want to be able to shard them when they get big enough. Ideally there would be a user-specified way to pick a key, _id is a good default at any rate.

### DIFF
--- a/arctic/_util.py
+++ b/arctic/_util.py
@@ -23,7 +23,7 @@ def are_equals(o1, o2, **kwargs):
         return False
 
 
-def enable_sharding(arctic, library_name, hashed=True):
+def enable_sharding(arctic, library_name, hashed=True, key='symbol'):
     c = arctic._conn
     lib = arctic[library_name]._arctic_lib
     dbname = lib._db.name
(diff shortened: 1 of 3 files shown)


arctic/_util.py
```diff
@@ -23,7 +23,7 @@ def are_equals(o1, o2, **kwargs):
         return False
 
 
-def enable_sharding(arctic, library_name, hashed=True):
+def enable_sharding(arctic, library_name, hashed=True, key='symbol'):
```
bmoscon: are other keys possible? are they all user defined? are they all internally defined? I can't tell. Might be nice to add a docstring to this

achamayou: I will add a docstring, 'symbol' is the only key that was used until now, but for BSONStore we would like other options.

achamayou: Updated now.

jamesblackburn: This does preclude multi-key sharding.  But perhaps this is a rare case anyway...

arctic/store/bson_store.py
```diff
@@ -30,7 +30,7 @@ def __init__(self, arctic_lib):
     def initialize_library(cls, arctic_lib, hashed=True, **kwargs):
         logger.info("Trying to enable sharding...")
         try:
-            enable_sharding(arctic_lib.arctic, arctic_lib.get_name(), hashed=hashed)
+            enable_sharding(arctic_lib.arctic, arctic_lib.get_name(), hashed=hashed, key='_id')
```
jamesblackburn: If you're sharding by `_id` I think you always want `hashed=True` (as `ObjectId`s are ordered, so range based sharding will cause them to move about, and all writes will happen to a single shard).

arctic/_util.py
achamayou: I'm happy to change to a list (keys), or we can add support for it later if/when it becomes useful.

arctic/store/bson_store.py
achamayou: Ok, but I think I need to keep hashed to comply with the interface. I propose ignoring hashed=False with a warning in the log that it's been ignored, does that sound ok?

jamesblackburn: More a mention FYI.  We set hashed=True on everything as there's little downside.

arctic/_util.py
jamesblackburn: Agreed, probably leave it for now if you're happy.

arctic/store/bson_store.py
achamayou: Ok, I added the warning anyway, since I don't think it can hurt.

arctic/_util.py
achamayou: Ok!

bmoscon:
also @achamayou don't forget to update changes.md!

achamayou:
Will do! I tend to do that from the Github UI directly.

PR #373, opened 2017-06-09
Revert "Merge pull request #363 from manahl/add-concat-flag"

achamayou:
This reverts commit e6695da8770ab87bdbb74c22f48960bc938a9298, reversing
changes made to 5621ce470f73b73cdf71d2ecefcb67133b701f36.

This is causing ordering issues when appending with overlap, it probably needs more work.

PR #363, opened 2017-05-26
Raise if appending out-of-order items; Add concat flag

AdrianTeng:
This function exists since 2014 internally (originally by @TomTaylorLondon). Porting this to open source world.

PR #376, opened 2017-06-19
Concat flag

AdrianTeng:
@achamayou undo-ed #363 in #373, due to breakage in AHL internal test. Turns out the broken test is just copies of https://github.com/manahl/arctic/blob/master/tests/integration/store/test_version_store.py#L261, so I just need to apply the same test change.


arctic/store/_pandas_ndarray_store.py
```diff
@@ -116,6 +116,24 @@ def get_info(self, version):
...
+    def read_segment_last_dt(self, version):
+        if 'segment_index' in version:
+            index = np.fromstring(decompress(version['segment_index']), dtype=INDEX_DTYPE)
+            dt_index = self._datetime64_index(index)
+            if dt_index:
+                return index[dt_index][-1]
+        return None
+
+    def slice_overlap_item_or_raise(self, item, previous_version, concat):
+        """If new item has overlap dt with previous version, keep only new bits if concat=True; raise if concat=False"""
```
achamayou: Maybe mention the type that's raised?

arctic/store/_pandas_ndarray_store.py
```diff
@@ -116,6 +116,24 @@ def get_info(self, version):
         ret['dtype'] = ast.literal_eval(version['dtype'])
         return ret
 
+    def read_segment_last_dt(self, version):
```
achamayou: Since this is public, best to give it a docstring?

achamayou:
Strange, it's the same transient error I've seen before as well:

`>       raise AssertionError(msg)
E       AssertionError: DataFrame.columns are different
E       
E       DataFrame.columns values are different (66.66667 %)
E       [left]:  Index(['a', 'c', 'b'], dtype='object')
E       [right]: Index(['a', 'b', 'c'], dtype='object')`

This only seems to happen on 3.5, and not every time.

bmoscon:
@achamayou yes its transient, only on 3.5 and only on travis.

Issue #343, opened 2017-04-10
Library rename fails with libraries using namespaces.


arturomichel:
#### Arctic Version
```
1.37
```

#### Arctic Store
```
VersionStore, TickStore
```

#### Platform and version
Ubuntu 14.04.5 LTS
Python 3.5
Mongo 3.2.9

#### Description of problem and/or code sample that reproduces the issue
The rename feature throws an error when using namespaces. There are a couple of other scenarios under which this fails. I can only get it to work with libraries with no namespace. 

```
import arctic as a
store = a.Arctic('localhost')
store.initialize_library('test.TEST_LIBRARY')
store.rename_library('test.TEST_LIBRARY', 'prod.TEST_LIBRARY')
---------------------------------------------------------------------------
OperationFailure                          Traceback (most recent call last)
<ipython-input-14-ca192b90bd53> in <module>()
----> 1 lib =get._store.rename_library('test.TEST_LIBRARY', 'prod.TEST_LIBRARY')

/opt/anaconda3/envs/dev1/lib/python3.5/site-packages/arctic/arctic.py in rename_library(self, from_lib, to_lib)
    320         for coll in l._db.collection_names():
    321             if coll.startswith(colname + '.'):
--> 322                 l._db[coll].rename(coll.replace(from_lib, to_lib))
    323         if from_lib in self._library_cache:
    324             del self._library_cache[from_lib]

/opt/anaconda3/envs/dev1/lib/python3.5/site-packages/pymongo/collection.py in rename(self, new_name, **kwargs)
   1773         cmd.update(kwargs)
   1774         with self._socket_for_writes() as sock_info:
-> 1775             sock_info.command('admin', cmd)
   1776 
   1777     def distinct(self, key, filter=None, **kwargs):

/opt/anaconda3/envs/dev1/lib/python3.5/site-packages/pymongo/pool.py in command(self, dbname, spec, slave_ok, read_preference, codec_options, check, allowable_errors, check_keys, read_concern)
    237                            check, allowable_errors, self.address,
    238                            check_keys, self.listeners, self.max_bson_size,
--> 239                            read_concern)
    240         except OperationFailure:
    241             raise

/opt/anaconda3/envs/dev1/lib/python3.5/site-packages/pymongo/network.py in command(sock, dbname, spec, slave_ok, is_mongos, read_preference, codec_options, check, allowable_errors, address, check_keys, listeners, max_bson_size, read_concern)
    100         response_doc = unpacked['data'][0]
[...]

PR #389, opened 2017-07-13
Renaming libraries within a database/namespace

bmoscon:
Previously this did not work, you could only rename libraries named "test", a library named "test.test" was not renameable. 

see #343


bmoscon:
@jamesblackburn

PR #390, opened 2017-07-17
Metadata point-in-time

kraphtuos:
Add a metadata collection for metadata point-in-time query

### DIFF
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -44,10 +44,6 @@ class VersionStore(object):
         if '%s.changes' % c.name not in mongo_retry(c.database.collection_names)():
             # 32MB buffer for change notifications
             mongo_retry(c.database.create_collection)('%s.changes' % c.name, capped=True, size=32 * 1024 * 1024)
-        # TODO: This should not be here
-        if '%s.metadata' % c.name not in mongo_retry(c.database.collection_names)():
-            # metadata entries
-            mongo_retry(c.database.create_collection)('%s.metadata' % c.name)
 
         for th in _TYPE_HANDLERS:
             th.initialize_library(arctic_lib, **kwargs)
@@ -128,6 +124,9 @@ class VersionStore(object):
             Return the symbols available under the snapshot.
         regex : `str`
             filter symbols by the passed in regular expression
+        timed_metadata_only : `bool`
+            search through .metadata if True
+            Default: False
         kwargs :
             kwarg keys are used as fields to query for symbols with metadata matching
             the kwargs query
@@ -139,10 +138,9 @@ class VersionStore(object):
         query = {}
         if regex is not None:
             query['symbol'] = {'$regex': regex}
-        if kwargs.pop('timed_metadata_only', False):
-            coll = self._metadata
-        else:
-            coll = self._versions
+
+        coll = self._metadata if kwargs.pop('timed_metadata_only', False) else self._versions
+
         if kwargs:
             for k, v in six.iteritems(kwargs):
                 query['metadata.' + k] = v
@@ -194,6 +192,10 @@ class VersionStore(object):
             `int` : specific version number
             `str` : snapshot name which contains the version
             `datetime.datetime` : the version of the data that existed as_of the requested point in time
+        kwargs :
+            timed_metadata_only : `bool`
+                search through ._metadata if True
+                Default: False
         """
 
         if kwargs.pop('timed_metadata_only', False):
@@ -606,38 +608,38 @@ class VersionStore(object):
 
     def write_metadata_history(self, symbol, entries, times, **kwargs):
         """
-        Manually write metadata entries in ._metadata collection
+        Manually overwrite entire metadata history for `symbol`
 
         Parameters
         ----------
         symbol : `str`
             symbol name for the item
-        entries : `list of dicts`
+        entries : `list of dict`
             list of metadata to be written
-        times : `list of datetime.datetimes`
+        times : `list of datetime.datetime`
             time stamps of metadata
         kwargs :
             passed through to the write handler
         """
-        if not (isinstance(entries, list) and isinstance(times, list)):
-            raise ValueError('`entries` and `times` must be lists')
+        if isinstance(entries, dict):
+            raise TypeError('`entries` must be a list')
+        entries = list(entries)
+        times = list(times)
 
-        if len(times) == len(entries):
-            times.append(None)
-        if len(times) < len(entries) + 1:
-            raise ValueError('Insufficient time stamps')
-        if len(times) > len(entries) + 1:
-            raise ValueError('Too many time stamps')
+        if len(times) != len(entries):
+            raise ValueError('Number of entries and number of time stamps do not match')
 
-        for i in range(len(entries) - 1):
+        self._delete_timed_metadata(symbol)
+        times.append(None)
+        for i in range(len(entries)):
             self._write_metadata_entry(symbol, entries[i], times[i], times[i + 1])
 
         if not self.has_symbol(symbol):
-            self.write_data(symbol, data=None, metadata=entries[-1])
+            self._write_data(symbol, data=None, metadata=entries[-1])
 
 
     @mongo_retry
-    def _append_metadata_history(self, symbol, metadata, metadata_policy=lambda old, new: new, **kwargs):
+    def _append_metadata_history(self, symbol, metadata, start_time, metadata_policy=lambda old, new: new, **kwargs):
         """
         Update .metadata entry for `symbol`
 
@@ -647,6 +649,8 @@ class VersionStore(object):
             symbol name for the item
         metadata : `dict`
             to be persisted
+        start_time : `datetime.datetime`
+            when metadata becomes effective
         metadata_policy :
             function(old, new):
                 return final metadata to write
@@ -662,21 +666,25 @@ class VersionStore(object):
         if not metadata:
             return None
         elif self.has_symbol(symbol):
-            old_metadata = self.read_metadata(symbol).metadata
-            new_metadata = metadata_policy(old_metadata, metadata)
-            if not new_metadata or old_metadata == new_metadata:
-                logger.info('No change to metadata.')
-                return None
-            else:
-                metadata = new_metadata
-        now = dt.now()
-        self._metadata.find_one_and_update({'symbol': symbol}, {'$set': {'end_time': now}},
+            old_metadata = self.read_metadata(symbol, metadata_history=True).metadata
+            if len(old_metadata) > 0:
+                if old_metadata[0]['start_time'] >= start_time:
+                    raise ValueError('start_time is earlier than the last metadata')
+                if old_metadata[0]['metadata']:
+                    new_metadata = metadata_policy(old_metadata[0]['metadata'], metadata)
+                    if not new_metadata or old_metadata[0]['metadata'] == new_metadata:
+                        logger.info('No change to metadata.')
+                        return None
+                    else:
+                        metadata = new_metadata
+
+        self._metadata.find_one_and_update({'symbol': symbol}, {'$set': {'end_time': start_time}},
                                             sort=[('start_time', pymongo.DESCENDING)])
-        self._write_metadata_entry(symbol, metadata, now, **kwargs)
+        self._write_metadata_entry(symbol, metadata, start_time, **kwargs)
         return metadata
 
     @mongo_retry
-    def update_metadata(self, symbol, metadata, metadata_policy=lambda old, new: new, **kwargs):
+    def update_metadata(self, symbol, metadata, start_time=None, metadata_policy=lambda old, new: new, **kwargs):
         """
         Update .metadata entry for `symbol` without changing the data
 
@@ -686,6 +694,8 @@ class VersionStore(object):
             symbol name for the item
         metadata : `dict`
             to be persisted
+        start_time : `datetime.datetime`
+            when metadata becomes effective
         metadata_policy :
             function(old, new):
                 return final metadata to write
@@ -700,10 +710,12 @@ class VersionStore(object):
         """
         if not metadata:
             raise ValueError('Metadata not specified.')
-        new_metadata = self._append_metadata_history(symbol, metadata, metadata_policy, **kwargs)
+        if start_time is None:
+            start_time = dt.now()
+        new_metadata = self._append_metadata_history(symbol, metadata, start_time, metadata_policy, **kwargs)
         if new_metadata:
             new = self._versions.find_one_and_update({'symbol': symbol}, {'$set': {'metadata': new_metadata}},
-                                                     sort=[('version', pymongo.DESCENDING)])
+                                                     sort=[('version', pymongo.DESCENDING)], new=True)
             return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(),
                                  version=new['version'], metadata=new.pop('metadata', None), data=None)
         logger.warning('No changes detected. Returning current entry.')
@@ -712,9 +724,9 @@ class VersionStore(object):
                              version=old.version, metadata=old.metadata, data=None)
 
     @mongo_retry
-    def write_data(self, symbol, data, metadata=None, prune_previous_version=True, **kwargs):
+    def _write_data(self, symbol, data, metadata=None, prune_previous_version=True, **kwargs):
         """
-        Write `data` under the specified `symbol` name to ._versions.
+        Write `data` under the specified `symbol` name in ._versions
 
         Parameters
         ----------
@@ -796,14 +808,14 @@ class VersionStore(object):
         """
         self._arctic_lib.check_quota()
 
-        new_metadata = self._append_metadata_history(symbol, metadata, metadata_policy, **kwargs)
+        new_metadata = self._append_metadata_history(symbol, metadata, dt.now(), metadata_policy, **kwargs)
         if new_metadata:
             metadata = new_metadata
         elif self.has_symbol(symbol):
             metadata = self.read_metadata(symbol).metadata
         else:
             metadata = None
-        return self.write_data(symbol, data, metadata, prune_previous_version=prune_previous_version, **kwargs)
+        return self._write_data(symbol, data, metadata, prune_previous_version=prune_previous_version, **kwargs)
 
     def _prune_previous_versions(self, symbol, keep_mins=120):
         """
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -70,9 +70,9 @@ class VersionStore(object):
         collection.versions.create_index([('symbol', pymongo.ASCENDING), ('version', pymongo.DESCENDING)], unique=True,
                                          background=True)
         collection.version_nums.create_index('symbol', unique=True, background=True)
-        collection.metadata.create_index([('symbol', pymongo.ASCENDING), ('_start_time', pymongo.DESCENDING)],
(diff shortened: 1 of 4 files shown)


arctic/store/version_store.py
```diff
@@ -107,7 +115,7 @@ def _read_preference(self, allow_secondary):
         return ReadPreference.NEAREST if allow_secondary else ReadPreference.PRIMARY
 
     @mongo_retry
-    def list_symbols(self, all_symbols=False, snapshot=None, regex=None, **kwargs):
+    def list_symbols(self, all_symbols=False, snapshot=None, regex=None, metadata=False, **kwargs):
```
vodaniel: because the option to list symbols only with point in time metadata is going to be mostly used by/for dev:
1) consider renaming the parameter to timed_metadata_only, name=metadata is ambiguous
2) consider hiding it in kwargs ? ( = advanced dev usage only )

As Yoav said this should probably behave like a meta(...) search

arctic/store/version_store.py
```diff
@@ -162,11 +177,11 @@ def list_symbols(self, all_symbols=False, snapshot=None, regex=None, **kwargs):
                                       }
                          }])
 
-        results = self._versions.aggregate(pipeline)
+        results = coll.aggregate(pipeline)
         return sorted([x['symbol'] for x in results])
 
     @mongo_retry
-    def has_symbol(self, symbol, as_of=None):
+    def has_symbol(self, symbol, as_of=None, metadata=False):
```
vodaniel: same as above, parameter name=metadata is ambig

arctic/store/version_store.py
```diff
@@ -293,7 +313,7 @@ def _write_handler(self, version, symbol, data, **kwargs):
             handler = self._bson_handler
         return handler
 
-    def read(self, symbol, as_of=None, date_range=None, from_version=None, allow_secondary=None, **kwargs):
+    def read(self, symbol, as_of=None, date_range=None, from_version=None, allow_secondary=None, all_metadata=False, **kwargs):
```
vodaniel: naming of the parameter, all_metadata:
consider:
 include_meta_history_in_time
 metadata_history
 metadata_in_time
etc.

arctic/store/version_store.py
```diff
@@ -435,10 +466,13 @@ def _read_metadata(self, symbol, as_of=None, read_preference=None):
...
             raise NoDataFoundException("No data found for %s in library %s" % (symbol, self._arctic_lib.get_name()))
 
+        if all_metadata:
+            _version['metadata'] = list(self._metadata.find({'symbol': symbol}, sort=[('_start_time', -1)]))
+
         return _version
 
     @mongo_retry
-    def append(self, symbol, data, metadata=None, prune_previous_version=True, upsert=True, **kwargs):
+    def append(self, symbol, data, metadata=None, prune_previous_version=True, metadata_policy=lambda old, new: new, upsert=True, **kwargs):
```
vodaniel: consider metadata_update_policy=lambda old, new: new or old

kraphtuos: If the metadata_policy returns False, no change to metadata is made.

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+
+        Returns
+        -------
+        VersionedItem containing .metadata
+        """
+        if not metadata:
+            raise ValueError('Metadata not specified.')
+        new_metadata = self._update_metadata_history(symbol, metadata, metadata_policy, **kwargs)
+        if new_metadata:
+            data = self.read(symbol).data
```
vodaniel: discussed offline:
 we will not re-write the data: just update the metadata key in _versions and add a new document in the timed _metadata collection

arctic/store/version_store.py
```diff
@@ -630,6 +822,21 @@ def _prune_previous_versions(self, symbol, keep_mins=120):
         cleanup(self._arctic_lib, symbol, version_ids)
 
     @mongo_retry
+    def delete_metadata(self, symbol):
```
vodaniel: should we consider including an option to delete from versions too ?
Is the case where we delete only the timed meta the dominant one ?

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+        mongo_retry(self._metadata.insert_one)(version)
+
+        logger.debug('Finished writing metadata for %s', symbol)
+
+        self._publish_change(symbol, version)
+
+        return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=0, data=None, metadata=metadata)
+
+    @mongo_retry
+    def _update_metadata_history(self, symbol, metadata, metadata_policy=lambda old, new: new, **kwargs):
```
vodaniel: naming: consider something like _append_to_metadata_history or _merge_to_meta_history 
not sure, we might need to think about naming here and everywhere

bmoscon:
@kraphtuos consider adding more documentation to the docstrings. As someone who is unaware what these changes are for, I have no idea what you'd use these new metadata APIs for. When the comments/docs are unclear, users open tickets to ask how to use it. Since we don't currently have comprehensive documentation, the code comments are critical :)

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+            Default: None
+        kwargs :
+            passed through to the write handler
+
+        Returns
+        -------
+        VersionedItem containing .metadata
+        """
+        self._arctic_lib.check_quota()
+        version = {'_id': bson.ObjectId()}
```
vodaniel: consider renaming this local variable: it suggests this is a _versions document

arctic/store/version_store.py
```diff
@@ -44,6 +44,10 @@ def initialize_library(cls, arctic_lib, hashed=True, **kwargs):
         if '%s.changes' % c.name not in mongo_retry(c.database.collection_names)():
             # 32MB buffer for change notifications
             mongo_retry(c.database.create_collection)('%s.changes' % c.name, capped=True, size=32 * 1024 * 1024)
+        # TODO: This should not be here
```
vodaniel: if anyone has a suggestion about a better place to do this please let us know

kraphtuos:
@bmoscon Thank you for your comment. Once we have finalised all the api details I will expand the doc strings to include usage and examples.

arctic/store/version_store.py
```diff
@@ -44,6 +44,10 @@ def initialize_library(cls, arctic_lib, hashed=True, **kwargs):
...
+        if '%s.metadata' % c.name not in mongo_retry(c.database.collection_names)():
```
jamesblackburn: In mongo you don't need to specifically create collections (unless you want to change the parameters on how it's setup).  So just remove this code.

arctic/store/version_store.py
```diff
@@ -66,6 +70,10 @@ def _ensure_index(self):
         collection.versions.create_index([('symbol', pymongo.ASCENDING), ('version', pymongo.DESCENDING)], unique=True,
                                          background=True)
         collection.version_nums.create_index('symbol', unique=True, background=True)
+        collection.metadata.create_index([('symbol', pymongo.ASCENDING), ('_start_time', pymongo.DESCENDING)],
```
jamesblackburn: Should start_time be ascending rather than descending?

arctic/store/version_store.py
```diff
@@ -131,6 +142,10 @@ def list_symbols(self, all_symbols=False, snapshot=None, regex=None, **kwargs):
         query = {}
         if regex is not None:
             query['symbol'] = {'$regex': regex}
+        if metadata:
```
jamesblackburn: Presumably this will only be used for backwards compatibility - so, if possible, don't expose this to users. (Anything on the API has to exist forever)

arctic/store/version_store.py
```diff
@@ -451,11 +485,19 @@ def append(self, symbol, data, metadata=None, prune_previous_version=True, upser
             to be persisted
         metadata : `dict`
             an optional dictionary of metadata to persist along with the symbol.
+        metadata_policy :
+            function(old, new):
```
jamesblackburn: What does this mean?

arctic/store/version_store.py
jamesblackburn: This definitely needs a big comment...

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+        -------
+        VersionedItem containing .metadata
+        """
+        self._arctic_lib.check_quota()
+        version = {'_id': bson.ObjectId()}
+        version['symbol'] = symbol
+        version['metadata'] = metadata
+        version['_start_time'] = start_time
+        if end_time is not None:
+            version['_end_time'] = end_time
```
jamesblackburn: Does anything check whether there's a previous metadata document that has no `end_time`?

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+
+        Returns
+        -------
+        VersionedItem containing .metadata
+        """
+        self._arctic_lib.check_quota()
+        version = {'_id': bson.ObjectId()}
+        version['symbol'] = symbol
+        version['metadata'] = metadata
+        version['_start_time'] = start_time
```
jamesblackburn: Not a fan of `_` prefixed variables that are API. Given metadata is stored in a sub-document. Can you remove the leading underscore and just call this `start` and `end`?

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+
+        handler = self._write_handler(version, symbol, metadata, **kwargs)
+        mongo_retry(handler.write)(self._arctic_lib, version, symbol, None, None, **kwargs)
+
+        # Insert the new version into the metadata DB
+        mongo_retry(self._metadata.insert_one)(version)
+
+        logger.debug('Finished writing metadata for %s', symbol)
+
+        self._publish_change(symbol, version)
```
jamesblackburn: Don't think this is needed

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+
+        return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=0, data=None, metadata=metadata)
```
jamesblackburn: `version=0` ?

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
             mongo_retry(self._changes.insert_one)(version)
 
     @mongo_retry
-    def write(self, symbol, data, metadata=None, prune_previous_version=True, **kwargs):
+    def _write_metadata_history(self, symbol, metadata, start_time, end_time=None, **kwargs):
```
jamesblackburn: `metadata_history` makes it sound like it's writting multiple metadata entries.  Perhaps `_write_metdata_entry`?

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+
+        Returns
+        -------
+        `metadata` if metadata is written, None if not.
+        """
+        if not metadata:
+            return None
+        elif self.has_symbol(symbol, metadata=True):
+            old_metadata = self.read_metadata(symbol).metadata
+            new_metadata = metadata_policy(old_metadata, metadata)
```
jamesblackburn: It would be interesting to see how `metadata_policy` is used.  Does it need to be a function?

arctic/store/version_store.py
jamesblackburn: Can you add a unique constraint on `(symbol, start)`? This would prevent multiple entries being stored for the same symbol because of a retry or similar.

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+            return self.write_data(symbol, data, new_metadata, **kwargs)
+        logger.warning('No changes detected. Returning current entry.')
+        old = self.read_metadata(symbol)
+        return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(),
+                                 version=old.version, metadata=old.metadata, data=None)
+
+    @mongo_retry
+    def write_data(self, symbol, data, metadata=None, prune_previous_version=True, **kwargs):
+        """
+        Write `data` under the specified `symbol` name to ._versions.
```
jamesblackburn: Data is written to the chunks collection, only the version information lives in `versions`

arctic/store/version_store.py
```diff
@@ -524,9 +568,123 @@ def _publish_change(self, symbol, version):
...
+        if new_metadata:
+            data = self.read(symbol).data
+            return self.write_data(symbol, data, new_metadata, **kwargs)
+        logger.warning('No changes detected. Returning current entry.')
+        old = self.read_metadata(symbol)
+        return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(),
+                                 version=old.version, metadata=old.metadata, data=None)
+
+    @mongo_retry
+    def write_data(self, symbol, data, metadata=None, prune_previous_version=True, **kwargs):
```
jamesblackburn: Why do we need a specific `write_data` call?

arctic/store/version_store.py
```diff
@@ -576,6 +733,41 @@ def write(self, symbol, data, metadata=None, prune_previous_version=True, **kwar
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=version['version'],
                              metadata=version.pop('metadata', None), data=None)
 
+
+    @mongo_retry
+    def write(self, symbol, data, metadata=None, prune_previous_version=True, metadata_policy=lambda old, new: new, **kwargs):
```
jamesblackburn: I'd keep the single `write` method if we can

arctic/store/version_store.py
jamesblackburn: `delete_version` isn't safe with `append` currently

arctic/store/version_store.py
kraphtuos: .metadata collection doesn't exist at the moment. Do I need to move this code to somewhere else (eg when the mongo library is created)?

arctic/store/version_store.py
kraphtuos: The idea is that documents in .metadata collection do not need to be versioned since user should never manually edit any document.

arctic/store/version_store.py
kraphtuos: ``write_data`` used to modify documents in .versions without making changes to .metadata where ``write`` also updates .metadata. This is so that legacy code using ``write`` will also update .metadata.

arctic/store/version_store.py
kraphtuos: ``metadata_policy`` is used to determine the new metadata to be written given the previous and the passed metadata. ``overwrite`` is used as default in arctic as it simulates previous behaviour. In Mongoose we will use tree_update instead.

arctic/store/version_store.py
```diff
@@ -197,12 +194,10 @@ def has_symbol(self, symbol, as_of=None, metadata=False):
             `int` : specific version number
```
vodaniel: document hidden params

arctic/store/version_store.py
vodaniel: this is not a versioned item and there is no req to return one, we could just return named tuple or the document itself

arctic/store/version_store.py
```diff
@@ -591,27 +587,57 @@ def _write_metadata_history(self, symbol, metadata, start_time, end_time=None, *
...
+        symbol : `str`
+            symbol name for the item
+        entries : `list of dicts`
+            list of metadata to be written
+        times : `list of datetime.datetimes`
+            time stamps of metadata
+        kwargs :
+            passed through to the write handler
+        """
+        if not (isinstance(entries, list) and isinstance(times, list)):
```
vodaniel: suggestion: accept any list-like object and create a local copy before you append, you do not want to modify the input here

arctic/store/version_store.py
vodaniel: to be deleted

arctic/store/version_store.py
```diff
@@ -131,6 +139,10 @@ def list_symbols(self, all_symbols=False, snapshot=None, regex=None, **kwargs):
         query = {}
         if regex is not None:
             query['symbol'] = {'$regex': regex}
+        if kwargs.pop('timed_metadata_only', False):
+            coll = self._metadata
+        else:
+            coll = self._versions
```
vodaniel: [nit][maybe]More pythonic:
```python
coll = self._metadata if kwargs.pop('timed_metadata_only', False) else self._versions
```

arctic/store/version_store.py
```diff
@@ -647,6 +649,8 @@ def _append_metadata_history(self, symbol, metadata, metadata_policy=lambda old,
             symbol name for the item
         metadata : `dict`
             to be persisted
+        start_time : `datetime.datetime`
+            when metadata becomes effective
         metadata_policy :
```
vodaniel: can we add more documentation for policy, please

arctic/store/version_store.py
```diff
@@ -686,6 +694,8 @@ def update_metadata(self, symbol, metadata, metadata_policy=lambda old, new: new
...
```
vodaniel: probably this is where we want to document the metadata policy

arctic/store/version_store.py
```diff
@@ -66,6 +66,10 @@ def _ensure_index(self):
         collection.versions.create_index([('symbol', pymongo.ASCENDING), ('version', pymongo.DESCENDING)], unique=True,
```
vodaniel: @jamesblackburn , James are you happy with the changes, merge OK  ?  cc: @kraphtuos

richardbounds:
Can we have an example of how this is intended to be used?

arctic/store/version_store.py
```diff
@@ -524,31 +570,124 @@ def _publish_change(self, symbol, version):
...
+        -------
+        VersionedItem containing .metadata
+        """
+        self._arctic_lib.check_quota()
+        document = {'_id': bson.ObjectId()}
+        document['symbol'] = symbol
+        document['metadata'] = metadata
+        document['start_time'] = start_time
+
+        handler = self._write_handler(document, symbol, metadata, **kwargs)
```
richardbounds: At this point, don't we know that metadata must be bson?

arctic/store/version_store.py
```diff
@@ -524,31 +570,124 @@ def _publish_change(self, symbol, version):
...
+            when entry becomes effective
+        kwargs :
+            passed through to the write handler
+
+        Returns
+        -------
+        VersionedItem containing .metadata
+        """
+        self._arctic_lib.check_quota()
+        document = {'_id': bson.ObjectId()}
```
richardbounds: Is document just dummy data? Seems like the tail is wagging the dog here.

richardbounds:
If I'm understanding this right (which I may not be...), this breaks the link between versions and metadata? Previously,  metadata was associated 1:1 with a version of the data - is that not the case any more?

I agree with @bmoscon - we should have more complete docstrings and examples before we merge this, as it seems like quite a fundamental change.

kraphtuos:
We assume metadata is not changing rapidly and is worth catalogued by time.
So this branch decouples metadata from data and stores them with time stamps in another collection.
This allows fast by temporal queries of metadata. A use case is presented in ahl.marketdata.meta (see PR292).
So far all the changes should be backward compatible and user will be migrated on write.

kraphtuos:
Changes moved to a new store.
See PR392.

PR #392, opened 2017-07-25
MetadataStore

kraphtuos:
This stores metadata with timestamps to allow temporal queries.
This is migrated from PR390.

### DIFF
--- a/arctic/store/metadata_store.py
+++ b/arctic/store/metadata_store.py
@@ -1,0 +1,211 @@
+from datetime import datetime as dt
+import logging
+
+import pandas as pd
+
+import bson
+import pymongo
+from ..decorators import mongo_retry
+from ..exceptions import NoDataFoundException
+
+from .bson_store import BSONStore
+
+logger = logging.getLogger(__name__)
+
+METADATA_STORE_TYPE = 'MetadataStore'
+
+class MetadataStore(BSONStore):
+    """
+    Metadata Store. This stores metadata with timestamp to allow temporal queries.
+    """
+
+    @classmethod
+    def initialize_library(cls, arctic_lib, hashed=True, **kwargs):
+        MetadataStore(arctic_lib)._ensure_index()
+        BSONStore.initialize_library(arctic_lib, hashed, **kwargs)
+
+    @mongo_retry
+    def _ensure_index(self):
+        self.create_index([('symbol', pymongo.ASCENDING), ('start_time', pymongo.DESCENDING)],
+                          unique=True, background=True)
+
+    def __init__(self, arctic_lib):
+        self._arctic_lib = arctic_lib
+        self._collection = self._arctic_lib.get_top_level_collection()
+
+    @mongo_retry
+    def find_one(self, *args, **kwargs):
+        """
+        See http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.find_one
+        """
+        return self._collection.find_one(*args, **kwargs)
+
+    @mongo_retry
+    def find_one_and_update(self, filter, update, **kwargs):
+        """
+        See http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.find_one_and_update
+        """
+        self._arctic_lib.check_quota()
+        return self._collection.find_one_and_update(filter, update, **kwargs)
+
+    def find_one_and_delete(self, filter, **kwargs):
+        """
+        See http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.find_one_and_delete
+        """
+        return self._collection.find_one_and_delete(filter, **kwargs)
+
+    def list_symbols(self):
+        return self.distinct('symbol')
+
+    def has_symbol(self, symbol):
+        return self.find_one({'symbol': symbol}) is not None
+
+    def read_metadata(self, symbol, history=False):
+        """
+        Return the metadata saved for a symbol.
+        
+        Parameters
+        ----------
+        symbol : `str`
+            symbol name for the item
+        history: `bool`
+            False: Returns current metadata (Default)
+            True: Returns all metadata entries
+            
+        Returns
+        -------
+        metadata document if history=False
+        pandas.DateFrame containing timestamps and metadata entries if history=True
+        """
+        if not history:
+            return self.find_one({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])['metadata']
+        find = self.find({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+        return pd.DataFrame(*zip(*[(entry['metadata'], entry['start_time']) for entry in find]), columns=['time', 'metadata'])
+
+    def _write_metadata_entry(self, symbol, metadata, start_time):
+        """
+        Create a new metadata entry
+
+        Parameters
+        ----------
+        symbol : `str`
+            symbol name for the item
+        metadata : `dict`
+            to be persisted
+        start_time : `datetime.datetime`
+            when entry becomes effective
+
+        Returns
+        -------
+        Document written
+        """
+        document = {'_id': bson.ObjectId()}
+        document['symbol'] = symbol
+        document['metadata'] = metadata
+        document['start_time'] = start_time
+
+        self.insert_one(document)
+
+        logger.debug('Finished writing metadata for %s', symbol)
+
+        return document
+
+    def write_metadata_history(self, collection):
+        """
+        Manually overwrite entire metadata history for `symbol`
+
+        Parameters
+        ----------
+        collection : `dict`
+            {symbol: (list of metadata, list of timestamps)}
+            symbol : `str`
+                symbol name for the item
+            metadata : `dict`
+                to be persisted
+            timestamp : `datetime.datetime`
+                start_time of the corresponding metadata
+            Example:
+                {'example': ([{}], [datetime.utcnow()])
+        """
+        if not isinstance(collection, dict):
+            raise TypeError('collection must be a dictionary.')
+        requests = []
+        for symbol, (entries, times) in collection.items():
+            if len(entries) != len(times):
+                raise ValueError('Number of entries and number of time stamps do not match.')
+            if self.has_symbol(symbol):
+                self.delete(symbol)
+            # TODO: Add checks to ensure metadata do change across any timestamp
+            for metadata, start_time, end_time in zip(entries, times, times[1:] + [None]):
+                doc = {'symbol': symbol, 'metadata': metadata, 'start_time': start_time}
+                if end_time is not None:
+                    doc['end_time'] = end_time
+                requests.append(doc)
+
+        self.bulk_write(requests)
+
+    def update_metadata(self, symbol, metadata, start_time=None):
+        """
+        Update metadata entry for `symbol` without changing the data
+
+        Parameters
+        ----------
+        symbol : `str`
+            symbol name for the item
+        metadata : `dict`
+            to be persisted
+        start_time : `datetime.datetime`
+            when metadata becomes effective
+            Default: datetime.utcnow()
+        """
+        if start_time is None:
+            start_time = dt.utcnow()
+        old_metadata = self.find_one({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+        if old_metadata is not None:
+            if old_metadata['start_time'] >= start_time:
+                raise ValueError('start_time is earlier than the last metadata')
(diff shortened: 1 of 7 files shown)


arctic/store/metadata_store.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        return self._collection.find_one_and_delete(filter, **kwargs)
+
+    def list_symbols(self):
+        return self.distinct('symbol')
+
+    def has_symbol(self, symbol):
+        return self.find_one({'symbol': symbol}) is not None
+
+    def read_metadata(self, symbol, history=False):
```
AdrianTeng: I would've implement this as a dataframe store, and return the last entry if `history=False`.
But this works as well :)

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,217 @@
...
+        if last_metadata is None:
+            raise NoDataFoundException('No metadata found')
+
+        self.find_one_and_delete({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+        mongo_retry(self.find_one_and_update)({'symbol': symbol}, {'$unset': {'end_time': ''}},
+                                              sort=[('start_time', pymongo.DESCENDING)])
+
+        return last_metadata
+
+    def delete_metadata(self, symbol):
```
AdrianTeng: May be `delete_all_metadata`? `delete_metadata` sounds quite innocent for what it actually does (chop everything)

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,217 @@
...
+        symbol : `str`
+            symbol name to delete
+
+        Returns
+        -------
+        Deleted metadata
+        """
+        last_metadata = self.find_one({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+        if last_metadata is None:
+            raise NoDataFoundException('No metadata found')
```
AdrianTeng:
```
raise NoDataFoundException('No metadata found for symbol {}'.format(symbol))
```

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,217 @@
...
+        start_time : `datetime.datetime`
+            when metadata becomes effective
+            Default: datetime.utcnow()
+        """
+        if start_time is None:
+            start_time = dt.utcnow()
+        old_metadata = self.find_one({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+        if old_metadata is not None:
+            if old_metadata['start_time'] >= start_time:
+                raise ValueError('start_time is earlier than the last metadata')
```
AdrianTeng: Add `old_metadata['start_time']` and `start_time` into exception message?

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,217 @@
...
+        """
+        if not isinstance(collection, dict):
+            raise TypeError('collection must be a dictionary.')
+        requests = []
+        for symbol, (entries, times) in collection.items():
+            if len(entries) != len(times):
+                raise ValueError('Number of entries and number of time stamps do not match.')
+            if self.has_symbol(symbol):
+                self.delete(symbol)
+            # TODO: Add checks to ensure metadata do change across any timestamp
```
AdrianTeng: :)

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,217 @@
...
+
+from .bson_store import BSONStore
+
+logger = logging.getLogger(__name__)
+
+METADATA_STORE_TYPE = 'MetadataStore'
+
+class MetadataStore(BSONStore):
+    """
+    Metadata Store. This stores metadata with timestamps to allow temporal queries.
```
bmoscon: more detail or an example or use case here might be useful

tests/integration/store/test_metadata_store.py
```diff
@@ -0,0 +1,69 @@
...
+
+
+def test_read_metadata(ms_lib):
+    assert ms_lib.read_metadata(symbol1) == None
+
+    ms_lib.write_metadata(symbol1, metadata1, start_time1)
+    assert ms_lib.read_metadata(symbol1) == metadata1
+    assert_frame_equal(ms_lib.read_metadata(symbol1, history=True), dataframe1)
+
+
```
AdrianTeng: Can we have 1 more test:

```
def test_repeat_write_metadata(ms_lib):
    ms_lib.write_metadata(symbol1, {'k1', 'v1'})
    assert ms_lib.read_metadata(symbol1) == {'k1', 'v1'}
    
    ms_lib.write_metadata(symbol1, {'k1', 'v2'})
    assert ms_lib.read_metadata(symbol1) == {'k1', 'v2'}
```

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+        self._collection = self._arctic_lib.get_top_level_collection().metadata
+
+    @mongo_retry
+    def find_one(self, *args, **kwargs):
+        """
+        See http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.find_one
+        """
+        return self._collection.find_one(*args, **kwargs)
+
+    @mongo_retry
```
TomTaylorLondon: Can you push find_one, find_one_and_update, find_one_and_delete down into BSON store? Quite useful generally.

arctic/store/metadata_store.py
TomTaylorLondon: +1

arctic/store/metadata_store.py
jamesblackburn: Are you using this in real code, or just replicating the Collection API?

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+            True: Returns all metadata entries
+
+        Returns
+        -------
+        metadata document if history=False
+        pandas.DateFrame containing timestamps and metadata entries if history=True
+        """
+        if not history:
+            res = self.find_one({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+            return res and res['metadata']
```
jamesblackburn: This looks magical - naive me would assume this returns a boolean?

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+        find = self.find({'symbol': symbol}, sort=[('start_time', pymongo.ASCENDING)])
+        times = []
+        entries = []
+        for item in find:
+            times.append(item['start_time'])
+            entries.append(item['metadata'])
+        return pd.DataFrame({'metadata': entries}, times)
+
+    def _write_metadata_entry(self, symbol, metadata, start_time):
```
jamesblackburn: Does this method add any value? It's only called from one place?

arctic/store/metadata_store.py
kraphtuos: I agree. This assumes res can never be False but not None. Fixed now.

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+                {'example': ([{}], [datetime.utcnow()])
+        """
+        if not isinstance(collection, dict):
+            raise TypeError('collection must be a dictionary.')
+        documents = []
+        for symbol, (entries, times) in collection.items():
+            if len(entries) != len(times):
+                raise ValueError('Number of entries and number of time stamps do not match.')
+            if self.has_symbol(symbol):
+                self.delete_all_metadata(symbol)
```
jamesblackburn: Potentially a dangerous default?  Deleting all metadata is someone is doing a bulk insert of history?

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+            doc = {'symbol': symbol, 'metadata': entries[0], 'start_time': times[0]}
+            for metadata, start_time in zip(entries[1:], times[1:]):
+                if metadata == doc['metadata']:
+                    continue
+                doc['end_time'] = start_time
+                documents.append(doc)
+                doc = {'symbol': symbol, 'metadata': metadata, 'start_time': start_time}
+            else:
```
jamesblackburn: `else` not needed? You always want to add the last `doc`

arctic/store/metadata_store.py
kraphtuos: My assumption is that if someone's trying to bulk insert history of a symbol then any existing metadata for that symbol is useless. And bulk insert should only happen when they initialise the history for that symbol for the first time.

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+            start_time = dt.utcnow()
+        old_metadata = self.find_one({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+        if old_metadata is not None:
+            if old_metadata['start_time'] >= start_time:
+                raise ValueError('start_time={} is earlier than the last metadata @{}'.format(start_time,
+                                                                                              old_metadata['start_time']))
+            if old_metadata['metadata'] == metadata:
+                logger.warning('No change to metadata')
+                return metadata
+        elif metadata is None:
```
jamesblackburn: Move this check to the top of the method?

arctic/store/metadata_store.py
kraphtuos: I agree. This was to ensure `doc` is added only when for loop exits properly which is unnecessary in this implementation.

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+            return
+
+        self.find_one_and_update({'symbol': symbol}, {'$set': {'end_time': start_time}},
```
jamesblackburn: Should this only be set where `end_time` is empty?  Can you for example have instruments where metadata doesn't exist for a period (maybe a de-listing type event)?

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+        Returns
+        -------
+        Deleted metadata
+        """
+        last_metadata = self.find_one({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+        if last_metadata is None:
+            raise NoDataFoundException('No metadata found for symbol {}'.format(symbol))
+
+        self.find_one_and_delete({'symbol': symbol}, sort=[('start_time', pymongo.DESCENDING)])
+        mongo_retry(self.find_one_and_update)({'symbol': symbol}, {'$unset': {'end_time': ''}},
```
jamesblackburn: Is it clear that `end_time` should be unset during deletion?

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,229 @@
...
+        Delete all metadata of `symbol`
+
+        Parameters
+        ----------
+        symbol : `str`
+            symbol name to delete
+        """
+        logger.warning("Deleting entire metadata history for %r from %r" % (symbol, self._arctic_lib.get_name()))
+        self.delete_many({'symbol': symbol})
+        assert not self.has_symbol(symbol)
```
jamesblackburn: I don't think this `assert` adds anything (and should be a real exception if this could go wrong).  I'd just remove this line.

arctic/store/metadata_store.py
kraphtuos: Pass `metadata=None` will actually set `metadata` to be `None` which is intended as some user may find it useful to unset metadata for a period of time. This will also explain the next two comments.

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,186 @@
...
+            raise TypeError('collection must be a dictionary.')
+        documents = []
+        for symbol, (entries, times) in collection.items():
+            if len(entries) != len(times):
+                raise ValueError('Number of entries and number of time stamps do not match.')
+            if self.has_symbol(symbol):
+                self.purge(symbol)
+            doc = {'symbol': symbol, 'metadata': entries[0], 'start_time': times[0]}
+            for metadata, start_time in zip(entries[1:], times[1:]):
+                if metadata == doc['metadata']:
```
gityoav: if metadata is None or ...?

arctic/store/metadata_store.py
```diff
@@ -0,0 +1,186 @@
...
+        find = self.find({'symbol': symbol}, sort=[('start_time', pymongo.ASCENDING)])
+        times = []
+        entries = []
+        for item in find:
+            times.append(item['start_time'])
+            entries.append(item['metadata'])
+        return pd.DataFrame({'metadata': entries}, times)
+
+    def write_history(self, collection):
+        """
```
gityoav: add the ability to write history from a pd.DataFrame assuming same format as the output from read()

kraphtuos:
If we are happy I will merge this tomorrow.

Issue #394, opened 2017-08-01
Enhancement request - multi-symbol reads


RichardWarfield:
Great project.  We're looking into using VersionStore for EOD data where we have a large number of symbols stored and often need to retrieve series for some arbitrary fraction of those.  So for example, 5000 stored symbols where we often need to read 500 at a time.  

If I understand correctly, we currently need to query each symbol with a separate call to store.read() and thus a separate Mongo query. This isn't ideal from a performance perspective and I was wondering whether some sort of multi-symbol read function had been considered, which could use an $in query to fetch all the symbols in one query. If there isn't a good reason for NOT having this I could potentially work on incorporating it in VersionStore.

bmoscon:
a symbol in arctic does not have to be the same as a stock symbol. Its just a name you've given to the data. You can store 5000 securities in a single data frame and write it as a single arctic symbol. 

That being said, I'm not sure if a multi symbol query is something we'd want or not, so I'll let @jamesblackburn  opine

RichardWarfield:
I get that a symbol can contain multiple series.  But in our case there isn't really a natural way to group the series in data frames for efficient concurrent access, short of putting everything in one frame -- which while I haven't benchmarked it, seems inefficient and would make versioning and metadata awkward.

bmoscon:
@RichardWarfield I'm going to get this done for you next week some time, so i'll tag you on the PR and you can tell me if its what you had in mind or not

RichardWarfield:
Thanks!

bmoscon:
done for chunkstore, ill do it for version store now

bmoscon:
I'm going to open a new ticket for it though so i have something to link against that is different

PR #401, opened 2017-08-19
Multi Symbol Read for ChunkStore

bmoscon:
@jamesblackburn @RichardWarfield please take a look. This addresses #394 

Its likely it doesnt 100% work correctly yet - I just wanted to code up a proof of concept and a simple unit test to test out base functionality. I haven't updated doc strings, or written up more complicated test cases with things like date ranges, symbols that do not exist, symbols outside of a date range, etc. 

I will add those tests and update the documentation assuming that this is in the right direction.

I also have not updated / changed tickstore or version store. I assume this is something we'd want in VersionStore but not Tickstore. Let me know @jamesblackburn

### DIFF
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -242,15 +242,17 @@ class ChunkStore(object):
         -------
         DataFrame or Series
         """
+        if not isinstance(symbol, list):
+            symbol = [symbol]
+
         sym = self._get_symbol_info(symbol)
         if not sym:
             raise NoDataFoundException('No data found for %s' % (symbol))
 
-        # special handling based on a single symbol or multi symbol read
-        spec = {SYMBOL: {'$in': symbol}} if isinstance(symbol, list) else {SYMBOL: symbol}
-        chunker = CHUNKER_MAP[sym[0][CHUNKER]] if isinstance(symbol, list) else CHUNKER_MAP[sym[CHUNKER]] 
-        deser = SER_MAP[sym[0][SERIALIZER]].deserialize if isinstance(symbol, list) else SER_MAP[sym[SERIALIZER]].deserialize 
-
+        
+        spec = {SYMBOL: {'$in': symbol}}
+        chunker = CHUNKER_MAP[sym[0][CHUNKER]]
+        deser = SER_MAP[sym[0][SERIALIZER]].deserialize
 
         if chunk_range is not None:
             spec.update(chunker.to_mongo(chunk_range))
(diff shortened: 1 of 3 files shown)


arctic/chunkstore/chunkstore.py
```diff
@@ -239,37 +242,43 @@ def read(self, symbol, chunk_range=None, filter_data=True, **kwargs):
...
         DataFrame or Series
         """
-
         sym = self._get_symbol_info(symbol)
         if not sym:
             raise NoDataFoundException('No data found for %s' % (symbol))
 
-        spec = {SYMBOL: symbol,
-               }
+        # special handling based on a single symbol or multi symbol read
```
pablojim: I wonder if it would cleaner to treat all cases as a list - e.g. place a single symbol in a one element list and then have a special case for the return value.

arctic/chunkstore/chunkstore.py
```diff
@@ -239,37 +242,43 @@ def read(self, symbol, chunk_range=None, filter_data=True, **kwargs):
...
+                            (START, pymongo.ASCENDING),
                             (SEGMENT, pymongo.ASCENDING)]
         segment_cursor = self._collection.find(spec, sort=by_start_segment)
 
-        chunks = []
+        chunks = defaultdict(list)
         for _, segments in groupby(segment_cursor, key=lambda x: x[START]):
             segments = list(segments)
-            mdata = self._mdata.find_one({SYMBOL: symbol, START: segments[0][START], END: segments[0][END]})
+            mdata = self._mdata.find_one({SYMBOL: segments[0][SYMBOL],
```
pablojim: Is there a possible issue here if segments for multiple symbols have the same start date?
Possibly one for a different PR but I wonder if this grouping could be done in Mongo also?

arctic/chunkstore/chunkstore.py
bmoscon: yeah's a good idea - it would clean the code up a lot. I'll make that change and update this

arctic/chunkstore/chunkstore.py
bmoscon: yeah thats a case I didn't test out. There might be an issue, but if this isn't a feature we want, I don't want to spend the time to work out all the kinks right away

pablojim: I think it would be a very useful feature. Especially in equities where the universe is very large but on a given day we trade a sub set of this universe. These leads to a tension between all equity data in a huge single symbol versus a symbol per equity. With an efficient multi symbol read the symbol per equity becomes more attractive.

I think the multiple symbols with the same start data can be solved with:
` for _, segments in groupby(segment_cursor, key=lambda x: (x[START], x[SYMBOL])):`
(untested!)

bmoscon: it appears to work as-is. perhaps because of the way the collection find is already sorted?

pablojim: Yeah reading the docs on itertools groupby this will mostly work. I think there is still an edge case. I'll try make a test case for it.

PR #415, opened 2017-09-18
Fix #414

### LINKED ISSUE #414: VersionStore snapshot failure
#### Arctic Version

```
1.43.0
```

#### Arctic Store

```
VersionStore
```

#### Platform and version

Using Python 3.5.2

I tried to snapshot a collection using function snapshot inside version_store. It produces following error message:

> <class 'pymongo.errors.OperationFailure'> list_symbols [Exceeded memory limit for $group, but didn't allow external sort. Pass allowDiskUse:true to opt in.], retrying 3

My understanding is that artic is trying to find all the symbols inside this collection, but since this is a large collections with over one million of symbols, artic function list_symbols causes above pymongo error. As the error message suggests, we should pass allowDiskUse:true instead. Is there any place in the arctic source code I can add this optional argument in? I tried to add a line 

`pipeline.extend({'allowDiskUse': True})`

inside function list_symbols in version_store.py, but it is not working.

Alternatively, is there any other way to solve this issue?

Many thanks.

kraphtuos:


arctic/store/version_store.py
```diff
@@ -162,7 +162,7 @@ def list_symbols(self, all_symbols=False, snapshot=None, regex=None, **kwargs):
                                       }
                          }])
 
-        results = self._versions.aggregate(pipeline)
+        results = self._versions.aggregate(pipeline, allowDiskUse=True)
```
jamesblackburn: Tbh I think the external sort is the real issue here - and we shouldn't be doing it for efficiency reasons.  Can we fix the query (or index) so that we don't require a full sort?

kraphtuos: I don't think any sorting is happening here. Just that if memory spills, it will be able to write to disk.
This is a new feature in v2.6 while v2.4 is used in circleci. (And no such problem in travis-ci)

jamesblackburn: It's spilling the temporary results to disk because the results of the aggregation pipeline don't fit in memory.  This is quite expensive (compared with an indexed read). 

Can the code / indexes be modified instead to not require external disk?

jamesblackburn:
Please make the commit comment and the PR self-standing (i.e. don't just reference the github issue).  The template should be something like:
https://chris.beams.io/posts/git-commit/

to keep the `git --oneline` shortlog nice and readable

arctic/store/version_store.py
kraphtuos: If the size of the returned list of symbols exceeds 16MB cap, spilling is inevitable when using aggregate.

jamesblackburn:
I'm going to close this as:
 - needs refactoring / rebasing
 - should, if possible, not use an external sort... 
 - it needs a simple unit / integration test

PR #416, opened 2017-09-18
Bugfix/metadata store append inconsistent return

kraphtuos:


arctic/store/metadata_store.py
```diff
@@ -2,12 +2,12 @@
 import logging
 
 import pandas as pd
-
 import bson
 import pymongo
+
+from .._util import indent
```
jamesblackburn: _minor_ lexicographic order

kraphtuos: But '_' comes before (lower case) letters. And this is consistent with VersionStore ordering.

PR #419, opened 2017-09-21
remove python 3.5 from travis.yml

bmoscon:
Arctic has no python3 exclusive features, much less anything exclusive to 3.6 over 3.5. No reason have increased build times by running on two 3.X versions, especially now that 3.5 is bugfix only

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,6 @@ sudo: false
 language: python
 python:
   - "2.7"
-  - "3.5"
   - "3.6"
 
 addons:


bmoscon:
@kraphtuos no idea. I can't reproduce it locally (and I set the tests to run a loop continually overnight with no issues) so I think it must be some issue with travis or docker :(

dag4202:
Whenever I attempt to pip install arctic 1.51 in a python:3.5 docker container, pip hangs after attempting to download arctic.  This is not an issue with arctic 1.49.

bmoscon:
3.5 is still supported, we just dont waste time running tests on it anymore since we have no python3 specific code in arctic. what OS are you on?

dag4202:
I'm on windows 10 professional, running docker for windows on hyper-V. I should probably open a new issue.

bmoscon:
@dag4202 no windows wheel was built for 1.51 which is probably the issue. 1.52 should be coming out soon, I'll make sure a wheel i built for that on Windows.

Issue #420, opened 2017-09-22
DateChunker doesn't respect DateRange interval


Zvezdin:
#### Arctic Version

1.52.0

#### Arctic Store

ChunkStore

#### Platform and version

Python 64 bit, Win 10.

#### Description of problem and/or code sample that reproduces the issue

`chunkStore.read(key, chunk_range = DateRange(startDate, endDate, CLOSED_OPEN))` returns as if the interval was CLOSED_CLOSED. In the DateChunker's source code, there is no place that checks the DateRange's interval

bmoscon:
from arctic's DateRange:

```
a      b     CLOSED_CLOSED                 date >= a and date <= b
a      b     CLOSED_OPEN                   date >= a and date < b
```

so the only difference between closed_closed and closed_open is that the 'end' date on the interval is non inclusive.


You are correct - this is not respected in chunkstore currently. It always defaults to closed_closed if an open and an end date are provided


```
    date_range = pd.date_range(start=datetime(2017, 5, 1), periods=8, freq='D')

    df = DataFrame(data={'data': range(8)},
                   index=DatetimeIndex(date_range, name='date'))


    chunkstore_lib.write('test', df, chunk_size='D')

    ret = chunkstore_lib.read('test', chunk_range=DateRange(datetime(2017,5,2), datetime(2017,5,5), CLOSED_OPEN))
    print(ret)
    ret = chunkstore_lib.read('test', chunk_range=DateRange(datetime(2017,5,2), datetime(2017,5,5), CLOSED_CLOSED))
    print(ret)
    ret = chunkstore_lib.read('test', chunk_range=DateRange(datetime(2017,5,2), None, CLOSED_OPEN))
    print(ret)
```


output:


```
                        data
date            
2017-05-02     1
2017-05-03     2
2017-05-04     3
2017-05-05     4
                        data
date            
2017-05-02     1
2017-05-03     2
2017-05-04     3
2017-05-05     4
                         data
date            
2017-05-02     1
2017-05-03     2
2017-05-04     3
2017-05-05     4
2017-05-06     5
2017-05-07     6
2017-05-08     7
```

bmoscon:
I'll get this fixed this weekend.

Zvezdin:
Thanks for the quick reply. I wanted to use a sliding window to load data from my chunkStore, and that problem occured.

Btw, can we just admire the issue's id?

bmoscon:
fixed

PR #426, opened 2017-09-26
fix date chunker so it respects intervals

bmoscon:
date chunker in chunkstore does not currently respect open intervals from date range when filtering data frames by user supplied chunk_range


bmoscon:
@jamesblackburn

bmoscon:
this fixes #420

Issue #427, opened 2017-09-26
chunkstore overwrite behavior not correct when index is removed


bmoscon:
#### Arctic Version

```
# Latest
```

#### Arctic Store

```
# Chunkstore
```

Write a dataframe to chunkstore with an index. Write the same dataframe to the same symbol without an index. When you read it back, it still has the index. The metadata is probably not being updated correctly.

bmoscon:
fixed

PR #429, opened 2017-09-27
Correctly set index in metadata and replace existing metadata

bmoscon:
Dataframes without an index have no "index" field in their metadata,
    so updating existing metadata does not remove the index field.
    Also, metadata was incorrectly being stored. symbol, start, and end
    are the index for the collection, but metadata was being
    stored without an index (so it was defaulting to null,null,null)


bmoscon:
fixes issue #427

bmoscon:
@jamesblackburn

Issue #425, opened 2017-09-26
Metadata for Tickstore


SnowWalkerJ:
Hi, I see from the documents that metadata can be saved in a VersionStore. However, I'm using TickStore and now I can only save metadata of each symbol in a seperate mongo library, which is very inconvenient.

Is there a way to save metadata in a TickStore?

bmoscon:
There is no way. What sorts of things are you looking to store? simple key: value type things (python dict)? Dataframes? Something else?

Also, what sort of granularity? One piece of metadata per symbol would be extremely easy to add. (that one piece could be as large or small as you want)

SnowWalkerJ:
A python dict is good enough to store something like the creator, privilege, or the source of the data.

It should not be difficult to add this kind of things into mongodb. I just think it would be convenient if a metadata argument is added to the read/write methods.

bmoscon:
fixed

PR #433, opened 2017-10-01
User defined metadata for tickstore

bmoscon:
this was requested in #425

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -354,6 +354,18 @@ class TickStore(object):
         return rtn
 
     def read_metadata(self, symbol):
+        """
+        Read metadata for the specified symbol
+
+        Parameters
+        ----------
+        symbol : `str`
+            symbol name for the item
+
+        Returns
+        -------
+        dict
+        """
         return self._metadata.find_one({SYMBOL: symbol})[META]
 
     def _pad_and_fix_dtypes(self, cols, column_dtypes):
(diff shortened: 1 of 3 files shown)


bmoscon:
@jamesblackburn et al. Not 100% complete (needs a lot more tests). But I don't want to write the rest of them up until I get some input

bmoscon:
@jamesblackburn still waiting on this. I'd rather not make any changes to tickstore that you haven't looked over.

arctic/tickstore/tickstore.py
```diff
@@ -238,7 +246,7 @@ def _read_preference(self, allow_secondary):
         return ReadPreference.NEAREST if allow_secondary else ReadPreference.PRIMARY
 
     def read(self, symbol, date_range=None, columns=None, include_images=False, allow_secondary=None,
-             _target_tick_count=0):
+             _target_tick_count=0, metadata=False):
```
jamesblackburn: This could well be in a `read_metadata` method (like in `VersionStore`) rather than overloading the return types on `read`.

bmoscon: yeah that makes sense. I'll change it to its own method

arctic/tickstore/tickstore.py
```diff
@@ -261,7 +269,7 @@ def read(self, symbol, date_range=None, columns=None, include_images=False, allo
 
         Returns
         -------
-        pandas.DataFrame of data
+        pandas.DataFrame of data, or tuple of (data, metadata) if metadata is `True`
```
jamesblackburn: Worth changing this back

bmoscon: good eye!

PR #438, opened 2017-10-06
Updating version number

jonbannister:


eeaston:
This had to be rolled as there was a deleted release on pypi.org, and you cannot replace it.

PR #441, opened 2017-10-17
Fix read of empty MultiIndex+tz Series

### LINKED ISSUE #440: Save/load fails on empty MultiIndex + timezone
This affects the `PandasStore` and anything using it. 
The test below fails

```
def test_save_read_pandas_empty_series_with_datetime_multiindex_with_timezone(library):
    empty_index = pd.MultiIndex(levels=(pd.DatetimeIndex([], tz="America/Chicago"), pd.Index([])), labels=([], []))
    df = Series(data=[], index=empty_index)
    library.write('pandas', df)
    saved_df = library.read('pandas').data
    assert empty_index.equal_levels(saved_df.index), "Index timezone information should be maintained, even when empty"
```

vmereuta:
It's impossible to determine the dtype of a saved empty index from the saved information alone and pandas has the bad habit of defaulting to either float or int. We can work out whether a MultIndex level is meant to be a DatetimeIndex based on the presence of the separate timezone information. Based on this, we can cast back empty indices to DatetimeIndex where appropriate.

This is only a shim - a good fix would have us saving down all the dtypes for both indices and data. This should ensure that any arbitrary type empty series written out can be read back in as it was.

Fixes #440

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### 1.54 (2017-10-18)
+  * Bugfix:  #440 Fix read empty MultiIndex+tz Series
+
 ### 1.53 (2017-10-06)
   * Perf:    #408 Improve memory performance of version store's serializer
   * Bugfix   #394 Multi symbol read in chunkstore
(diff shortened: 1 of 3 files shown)


bmoscon:
@vlad please update changes.md

vmereuta:
`Changes.md` updated.

Issue #442, opened 2017-10-17
Chunkstore segment scheme breaks incompatible with previously written data


bmoscon:
see comments at end of https://github.com/manahl/arctic/issues/407

basically update/append operations are not correctly cleaning up old segments (which start at -1).

bmoscon:
fixed

PR #452, opened 2017-11-12
fix legacy segments - issue-442

bmoscon:
previous changes fixed a bug where segments labels started with -1 and we not properly handled when the chunk size increased to more than 1 segment. This introduced a new bug where legacy data with -1 segments would not properly be removed and it would cause decompression errors during a read.

This fixes that by converting all -1 segments to 0 segments at library object instantiation.


bmoscon:
see #442

bmoscon:
@miketkelly

bmoscon:
I also added some code to alert the user if data has already been corrupted since the defect was present in the code for almost a year before it was fixed.

Issue #457, opened 2017-12-02
Helper tool to try and fix bad chunks in chunkstore


bmoscon:
single chunks with invalid segments can easily be fixed but multi segment chunks with invalid segment ids need a special tool to fix (that should be invoked manually)

bmoscon:
fixed

PR #458, opened 2017-12-03
Chunkstore Tools

bmoscon:
New tools module for chunkstore. includes first tool - the segment id repair
tool. This tool will scan all symbols (or user specified symbols) and
check for bad segment mappings and fix them.


bmoscon:
this is a fix for both #442 and #457

Issue #385, opened 2017-06-29
ZeroDivisionError: float division by zero


davidwneary:
#### Arctic Version

```
1.42.0
```

#### Arctic Store

```
# ChunkStore
```

#### Platform and version

Python 3.5. Mongo is hosted on the Azure CosmosDB.

#### Description of problem and/or code sample that reproduces the issue

An error that has only arisen since deploying the db to an Azure CosmosDB:

When I try to append to the database, I get this error:

```
  File "/usr/local/lib/python3.5/site-packages/arctic/chunkstore/chunkstore.py", line 506, in append
    self.__update(sym, item, metadata=metadata, combine_method=SER_MAP[sym[SERIALIZER]].combine, audit=audit)
  File "/usr/local/lib/python3.5/site-packages/arctic/chunkstore/chunkstore.py", line 400, in __update
    self._arctic_lib.check_quota()
  File "/usr/local/lib/python3.5/site-packages/arctic/arctic.py", line 488, in check_quota
    remaining_count = remaining / avg_size
ZeroDivisionError: float division by zero
```

avg_size is 0 because 'size' property of all of the relevant stats is 0.

There is definitely data in there - library.read shows me that.

Any idea what could be causing this issue?

Cheers
Dave

bmoscon:
Definitely not an expert, but I'm guessing the issue is with MongoDB. It may be wire compatible, but clearly something isn't 100% correct in terms of some of the non-wire operations that mongo supports.

I'd recommend contacting Azure support or CosmosDB support and seeing if they support anything like https://docs.mongodb.com/v3.2/reference/method/db.collection.stats/ for their database.

bmoscon:
Adding @jamesblackburn to see if he has any insight

davidwneary:
I agree it's almost certainly a CosmosDB issue, thanks for the link - I'll take it up with Azure.

bmoscon:
@davidwneary I think I'm going to put some exception handling in around check quota. The quota stuff encountering an error shouldnt kill a critical operation (read/write/etc).

bmoscon:
fixed - sorry for the delay!

PR #459, opened 2017-12-03
Prevent exceptions in check_quota from killing a write

bmoscon:
Since the quota has already been 'checked' we shouldn't letthe generation of informational statistics kill the write


bmoscon:
this fixes issue #385

Issue #250, opened 2016-10-02
Issues with multiprocessor/threading and later PyMongo versions


saeedamen:
First off, Arctic is a great library, thanks for open sourcing it and I'm going to add a wrapper for Arctic to be used in my own open source financial data libraries findatapy and finmarketpy! 

I found a bit of an issue with using multiprocessor/threading in Python and PyMongo, which might affect other users. Every so often MongoDB appears to deadlock, when I make calls from multiple threads via Arctic. It turns out this is an issue with later versions of PyMongo (mine is 3.2). There is however a workaround by using connect=False (see http://api.mongodb.com/python/current/faq.html#pymongo-fork-safe) - like the below:

```
c = pymongo.MongoClient('127.0.0.1', connect=False)
from arctic import Arctic
store = Arctic(c)
```

Other people have fixed the problem by downgrading to an older version e.g. 2.8 (see https://github.com/dcrosta/flask-pymongo/issues/66 and http://stackoverflow.com/questions/31030307/why-is-pymongo-3-giving-serverselectiontimeouterror/31194981#3119498)

bmoscon:
We can expose that as an option to arctic, but I don't think thats something we'd want to make default. 

thoughts @jamesblackburn ?

jamesblackburn:
This is weird and unexpected.  We make efforts not to connect eagerly in the Arctic constructor, but this doesn't seem to be working right in the raw native Arctic.  Needs investigating.

dimosped:
This might help to resolve this issue:
[https://github.com/manahl/arctic/pull/460](https://github.com/manahl/arctic/pull/460)

dimosped:
Using connect=False doesn't prevent socket being initialized in the parent process before forking (using MongoEngine):
https://github.com/MongoEngine/mongoengine/issues/1234#issuecomment-216769597

dimosped:
@bmoscon @jamesblackburn  do we think this is still an issue?
With multi-threading I doubt we have this issue.
With multiprocessing, creating MongoClient and then fork is known to cause deadlocks, but I expect this change to help mitigate the issue:
https://github.com/manahl/arctic/pull/460/files#diff-fb0e751696a7d90f9c0481d9544e1c53R118

bmoscon:
I think we should close this

PR #460, opened 2017-12-06
Optimize auth (fix for #464 , also potential fix for #250 )

dimosped:
This is a fix for #464 

Currently we go through the authentication every time the property "_db" in the ArcticLibraryBinding is accessed.

Getting the auth credentials multiple times with every read call, causes disk reads  (not cached, at least for the auth_hook implementation we use) and re-auth with Mongo.

This induces a large performance hit, and with this fix we get ~50% speed improvement (test with 200x reads from VersionStore).

**With this change**auth is called:
 - at instantiation of the Artict top level lib, or 
 - when/if Arctic is reset (explicitly or afer fork), for all cached stores, or
 - when a store instance is initialized

We need to make sure that this change is safe in a multiprocessing setup, wrt PyMongo locks.

Also **added safety for multiprocessing** given that MongoClient is not meant to be instantiated before fork, but users of Arctic may do so any time:
[http://api.mongodb.com/python/current/faq.html#using-pymongo-with-multiprocessing](http://api.mongodb.com/python/current/faq.html#using-pymongo-with-multiprocessing)

This might be a potential fix for issue #250 

Using connect=False doesn't prevent socket being initialized in the parent process before forking (using MongoEngine):
MongoEngine/mongoengine#1234 (comment)

### DIFF
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -90,7 +90,6 @@ class ChunkStore(object):
 
     @mongo_retry
     def _reset(self):
-        self._arctic_lib.reset_auth()
         # The default collection
         self._collection = self._arctic_lib.get_top_level_collection()
         self._symbols = self._collection.symbols
--- a/arctic/chunkstore/chunkstore.py
+++ b/arctic/chunkstore/chunkstore.py
@@ -90,6 +90,7 @@ class ChunkStore(object):
 
     @mongo_retry
     def _reset(self):
+        self._arctic_lib.reset_auth()
         # The default collection
         self._collection = self._arctic_lib.get_top_level_collection()
         self._symbols = self._collection.symbols
--- a/tests/integration/test_arctic.py
+++ b/tests/integration/test_arctic.py
@@ -1,9 +1,12 @@
-from datetime import datetime as dt
-from mock import patch
-from pandas.util.testing import assert_frame_equal
+import os
 import pytest
 import time
 
+from datetime import datetime as dt
+from mock import patch
+from multiprocessing import Process
+from pandas.util.testing import assert_frame_equal
+from random import random
 
 from arctic.arctic import Arctic, VERSION_STORE
 from arctic.exceptions import LibraryNotFoundException, QuotaExceededException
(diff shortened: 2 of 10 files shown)


bmoscon:
circle ci passes, but the failures on travis look real.

dimosped:
Thanks @bmoscon , I would appreciate it if you could fix travis. 
I just pushed another incremental commit for better integration test coverage.

bmoscon:
@dimosped my fix has been merged in. rebase with master and you should be good.

dimosped:
@bmoscon done, rebased and force pushed to my branch

bmoscon:
the travis 3.6 error is:  `print "\nDone pid={}".format(my_pid)`

might want to clean up some of those tests :D

tests/integration/test_arctic.py
```diff
@@ -190,3 +194,83 @@ def test_lib_rename_namespace(arctic):
...
+        time.sleep(random() * 0.2)
+    for i in range(total_writes):
+        if i % 20 == 0:  # add some randomisation, make sure that processes are multiplexed across time
+            time.sleep(random())
+        key = "{}_{}".format(my_pid, i)
+        vstore.write(key, data + [key])
+    for i in range(total_writes):
+        key = "{}_{}".format(my_pid, i)
+        assert vstore.read(key).data == data + [key]
+    print "\nDone pid={}".format(my_pid)
```
bmoscon: dont think we need the 'print' do we? if we do, make it a function please

dimosped: removed the print statement

bmoscon:
looks good to merge now!

arctic/chunkstore/chunkstore.py
```diff
@@ -90,6 +90,7 @@ def _check_invalid_segment(self):
 
     @mongo_retry
     def _reset(self):
+        self._arctic_lib.reset_auth()
```
jamesblackburn: I think what's a little bit awkward is requiring every implementation to `_reset_auth` on the LibraryBinding - rather than have the top-level Arctic object poke the bindings directly?  Otherwise we risk adding a new library type and forgetting to implement this reset functionality?

dimosped: unfortunatelly the Store implementations do not have a parent class so it is not possible to e.g. call super _reset() or do something similar.

The _reset() is being called only by Arcitc's reset() and form the init() of each Store implementation.

The auth details are stored/handled in ArcticLibraryBinding, so I could potentially:
 1. always trigger auth when get_top_level_collection() is called for first time 
 2. trigger auth in ArcticLibraryBinding whenever the underlying connection of Arctic has changed, i.e. Arctic's reset() has been called.

I'll push quickly a fix for this.

dimosped: The change is now pushed, please have a look

dimosped: added test for this exact scenario

tests/integration/test_arctic.py
```diff
@@ -190,3 +194,101 @@ def test_lib_rename_namespace(arctic):
 def test_lib_type(arctic):
     arctic.initialize_library('test')
     assert(arctic.get_library_type('test') == VERSION_STORE)
+
+
+MY_ARCTIC = None  # module-level Arctic singleton
```
jamesblackburn: might be worth splitting into a test_arctic_multithread.py

dimosped: sure, splitted

dimosped:
@bmoscon  this is now good to merge

PR #463, opened 2017-12-07
Fix integration tests with pymongo 3.6

bmoscon:
New version of pymongo has new default arg that mocks in the integration tests are not accounting for


bmoscon:
@dimosped this should fix the issue in travis

Issue #465, opened 2017-12-12
Feature Request: Add fast operations for write only metadata and restore symbol to a given version


dimosped:
#### Arctic Version
```
1.55
```

#### Arctic Store

```
# VersionStore
```

#### Description of problem and/or code sample that reproduces the issue

Currently there is no easy way to update the metadata only for a symbol without reading/re-writing the data for the given symbol.

Also, the users have no option to restore a given symbol to a version, without reading/re-writing the data/metadata form the reference version.

We need the Arctric's VersionStore API to be extended to support those two operations, whose runtime must be constant, regardless of the data size.

bmoscon:
@dimosped we can close this right?

shashank88:
Closing this, feel free to reopen if needed.

PR #466, opened 2017-12-12
added API for write only metadata and restore symbol to version. Fast operations.

dimosped:
No unit/integration tests added, this is the initial PR to discuss and reach a consensus on the API.

This implements feature: #465


arctic/store/version_store.py
```diff
@@ -576,6 +575,135 @@ def write(self, symbol, data, metadata=None, prune_previous_version=True, **kwar
...
+            self._prune_previous_versions(symbol)
+
+        logger.debug('Finished updating versions with new metadata for %s', symbol)
+
+        self._publish_change(symbol, new_version)
+
+        return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=new_version['version'],
+                             metadata=new_version.pop('metadata', None), data=None)
+
+    def write_metadata(self, symbol, metadata, prune_previous_version=True, **kwargs):
```
richardbounds: I still think `update_metadata(...)` would be clearer, but happy to go with the consensus on this.

jamesblackburn: (and fwiw I think `write_` is more consistent with this and the other APIs...)

arctic/store/version_store.py
```diff
@@ -576,6 +575,135 @@ def write(self, symbol, data, metadata=None, prune_previous_version=True, **kwar
         return VersionedItem(symbol=symbol, library=self._arctic_lib.get_name(), version=version['version'],
                              metadata=version.pop('metadata', None), data=None)
 
+    def _add_new_version_using_reference(self, symbol, new_version, reference_version, prune_previous_version):
+        if not new_version or \
+                not reference_version or \
+                new_version['symbol'] != reference_version['symbol'] or \
+                new_version['_id'] == reference_version['_id'] or \
+                not new_version['base_version_id']:
+            raise ArcticException("Illegal or non-matching nwe_version or reference_version were supplied: %s and %s",
```
jamesblackburn: Belt and braces, but this could never happen?

arctic/store/version_store.py
```diff
@@ -576,6 +575,135 @@ def write(self, symbol, data, metadata=None, prune_previous_version=True, **kwar
...
+                                  new_version, reference_version)
+
+        # Check if in the meanwhile the reference version (based on which we updated incrementally) has been removed
+        last_look = self._versions.find_one({'_id': reference_version['_id']})
+        if last_look is None or last_look.get('deleted'):
+            raise pymongo.errors.OperationFailure("Failed to write metadata for symbol %s. "
+                                                  "The previous version (%s, %d) has been removed during the update" %
+                                                  (symbol, str(reference_version['_id']), reference_version['version']))
+
+        # There is always a small risk here another process in between these two calls (above/below)
```
jamesblackburn: Couldn't you check that the `reference_version` exists after inserting the `new_version`?

arctic/store/version_store.py
```diff
@@ -576,6 +575,135 @@ def write(self, symbol, data, metadata=None, prune_previous_version=True, **kwar
...
+            symbol name for the item
+        as_of : `str` or `int` or `datetime.datetime`
+            Return the data as it was as_of the point in time.
+            `int` : specific version number
+            `str` : snapshot name which contains the version
+            `datetime.datetime` : the version of the data that existed as_of the requested point in time
+        prune_previous_version : `bool`
+            Removes previous (non-snapshotted) versions from the database.
+            Default: True
+        allow_secondary : `bool` or `None`
```
jamesblackburn: I don't think `allow_secondary` should be on this API. It's only there for scaling out reads

arctic/store/version_store.py
```diff
@@ -576,6 +575,135 @@ def write(self, symbol, data, metadata=None, prune_previous_version=True, **kwar
...
+        # Reaching here means that data and/or metadata exist and we are all set to update the metadata
+        new_version_num = self._version_nums.find_one_and_update({'symbol': symbol},
+                                                             {'$inc': {'version': 1}},
+                                                             upsert=True, new=True)['version']
+
+        # Create a new version entry, cloning the one supplied from the user
+        version = {k: version_to_restore[k] for k in version_to_restore.keys()}
+        version['_id'] = bson.ObjectId()
+        version['version'] = new_version_num
+        version['base_version_id'] = version_to_restore.get('base_version_id', version_to_restore['_id'])
```
jamesblackburn: In terms of testing - I think it would be worth testing the GC at this point... Up till now there's only ever been one `base_version_id` in flight as it's only used during `append`.  After this change version history can 'fork'

arctic/store/version_store.py
dimosped: removed

arctic/store/version_store.py
dimosped: not in our case (and almost unlikely we get the same uuid from bson's rng).
Added so that it can be re-used safely in the future.

arctic/store/version_store.py
dimosped: sure, I can check after and if missing I 'll remove the new version and raise an exception.

There will always be some extreme corner case scenarios (*) (since these actions are not atomic, or part of a transaction) however,  this check after insertion will certainly improve things.

process_B remains block just before the command which deletes the reference version, process_A finishes, process_B deletes the reference version

arctic/store/version_store.py
dimosped: I'd also vote for write_* simply for consistency, while I like more the update semantically.

arctic/store/version_store.py
dimosped: Check is now pushed

arctic/store/version_store.py
dimosped: yes, very good point, I actually tested this manually and will add integration test.
The GC will preserve both the base_verserion_id and the new version document which we created and references it.

I chose this for performance, so that we don't have to update at all the segments by adding a new parent id in the parent set.

jamesblackburn: Yep - the implementation looks good. Just wanted to check you'd thought about this :)

bmoscon:
looks fine but as james mentions - needs tests :)

dimosped:
@jamesblackburn @bmoscon @richardbounds  Unit and Integration tests are now added

arctic/store/version_store.py
jamesblackburn: I'd tend to user `assert` for constraints that can't be violated (apart from by bugs) and exceptions for things that are user API facing.

tests/integration/store/test_version_store.py
```diff
@@ -973,3 +973,212 @@ def test_append_after_empty(library):
...
+def test_write_metadata(library):
+    symbol = 'FTL'
+    mydf_a, mydf_b = _rnd_df(10, 5), _rnd_df(10, 5)
+    with patch('arctic.arctic.logger.info') as info:
+        library.write(symbol, data=mydf_a, metadata={'field_a': 1})  # creates version 1
+        library.write(symbol, data=mydf_b, metadata={'field_a': 2})  # creates version 2
+        library.write_metadata(symbol, metadata={'field_b': 1})  # creates version 3 (only metadata)
+        v = library.read(symbol)
+        assert_frame_equal(v.data, mydf_b)
+        assert v.metadata == {'field_b': 1}
```
jamesblackburn: could assert that `field_a` no longer exists.

arctic/store/version_store.py
dimosped: changed to using assertion here

PR #468, opened 2017-12-14
Re-adding compatibility for pandas 0.20.x

jonbannister:
The get_timezone library was moved around in 0.20 and 0.21. We are still using it even though it was deprecated, so adding some missing version compatibilities.

Related to #470

### DIFF
--- a/arctic/serialization/numpy_records.py
+++ b/arctic/serialization/numpy_records.py
@@ -2,11 +2,14 @@ import logging
 import numpy as np
 
 from pandas import DataFrame, MultiIndex, Series, DatetimeIndex, Index
-try:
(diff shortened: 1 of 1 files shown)


bmoscon:
I don’t think this sets a good precedent. Pandas is always doing things like this. I think we should support latest + 0.16.2. If you try and support every point release between it’s going to be ugly. Just my opinion!

bmoscon:
@jamesblackburn as long as its simple things like a few extra imports, I dont think its a big issue.

bmoscon:
@jonbannister are you going to merge this in before the 1.56 release?

jonbannister:
I'd like to, but I don't seem to have merge permission!

PR #471, opened 2017-12-16
support for clang 5.0 on OSX

bmoscon:
Clang 5.0 (easily installed with homebrew) seamlessly supports openMP


bmoscon:
@jamesblackburn I tested on 10.12 and 10.13. Works with clang on both (brew provided clang)

PR #472, opened 2017-12-20
Support for strings in tickstore columns

bmoscon:
if the column is an object type, assume its a string. If its not an exception will be raised.

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -636,6 +636,12 @@ class TickStore(object):
             array = array.astype('<f8')
         elif (array.dtype.kind) in ('U', 'S'):
             array = array.astype(np.unicode_)
+        elif (array.dtype.kind) == 'O':
+            try:
+                array = np.array([s.encode('ascii') for s in array])
+                array = array.astype(np.unicode_)
+            except:
+                raise UnhandledDtypeException("Unsupported dtype '%s' - only int64, float64 and U are supported" % array.dtype)
         else:
             raise UnhandledDtypeException("Unsupported dtype '%s' - only int64, float64 and U are supported" % array.dtype)
         # Everything is little endian in tickstore
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -638,10 +638,10 @@ class TickStore(object):
             array = array.astype(np.unicode_)
         elif (array.dtype.kind) == 'O':
             try:
-                array = np.array([s.encode('ascii') for s in array])
+                array = np.array([s.encode('utf-8') for s in array])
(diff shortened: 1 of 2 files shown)


bmoscon:
@jamesblackburn feel free to add any other reviewers you like. This is such a frequently requested feature I figured I should add it. Its definitely not the fastest, but if you aren't using columns with strings you'll never hit the performance impact. Numpy does have very fast ways to encode strings, but the array has to already be of `S` or `U` type for the function in `numpy.char` to work on the array. This is the only work around I could find since simply casting the array to unicode with astype will convert objects (like user defined class) to a string representation of the class signature

arctic/tickstore/tickstore.py
```diff
@@ -636,6 +636,12 @@ def _ensure_supported_dtypes(array):
             array = array.astype('<f8')
         elif (array.dtype.kind) in ('U', 'S'):
             array = array.astype(np.unicode_)
+        elif (array.dtype.kind) == 'O':
+            try:
+                array = np.array([s.encode('ascii') for s in array])
```
bashtage: Is real unicode support available?  Would it be possible to use UTF-8?  Not essential but might as well have it given that it is easy to slip into UTF8 sometimes.

arctic/tickstore/tickstore.py
```diff
@@ -636,6 +636,12 @@ def _ensure_supported_dtypes(array):
...
+                array = array.astype(np.unicode_)
+            except:
+                raise UnhandledDtypeException("Unsupported dtype '%s' - only int64, float64 and U are supported" % array.dtype)
```
bashtage: Maybe add to the error that object columns must be coercible to strings?  This can be a little confusing for pandas users since 'U' can't actually be set cast to anymore.

arctic/tickstore/tickstore.py
bmoscon: yeah thats easy enough to change

arctic/tickstore/tickstore.py
bmoscon: sure I can make the error message more specific for this specific failure

bashtage:
This will be very helpful for me.  In general, what types of data structures can be stored to a TICK_STORE?  DataFrames with some restrictions are OK.  I've been using List[dict] to get around the string issue for now.  Are dicts of arrays OK as long as one has the name 'index'?  This latter case would mostly fix the performance problems of converting DataFrames to List[dict] since I could use something like 

```
recs = df.to_records(index=False)
to_write = {'index':recs['index'],'Price':recs['Price']}
```

bmoscon:
@bashtage I don't have a lot of experience with storing `dict`s in tickstore, but i think you are correct. The unit/integration tests have a lot of examples

PR #474, opened 2017-12-20
Fixing issue where authentication is reset/failed (randomly) 

dimosped:
We have observed spurious MongoDB OperationFailure exceptions upon common operations like find:

`OperationFailure: not authorized on XXXXXX to execute command { find: "index", filter: { alias: "YYYY" }, limit: 1, singleBatch: true }`

Fixes:
  - Upon creation/initialization of ArcticLibraryBinding we call "get_top_level_collection()" to make sure that Arctic's cache has the library and therefore, if connection is reset, library's "_reset()" is called.
  - When using \@mongo_retry handle OperationFailure exception and try to re-authenticate. This will help successfully issue the mongo command if failure was due to broken/reset authentication.

#475

### DIFF
--- a/arctic/arctic.py
+++ b/arctic/arctic.py
@@ -408,7 +408,8 @@ class ArcticLibraryBinding(object):
         database_name, library = self._parse_db_lib(library)
         self.library = library
         self.database_name = database_name
-        self._auth(self.arctic._conn[self.database_name])
+        self._auth(self._curr_conn[self.database_name])  # authenticate
+        self.get_top_level_collection()  # allow early caching of the library in Arctic (assures _reset is called)
 
     @property
     def _db(self):
--- a/arctic/arctic.py
+++ b/arctic/arctic.py
@@ -415,7 +415,7 @@ class ArcticLibraryBinding(object):
     def _db(self):
         with self._lock:
             arctic_conn = self.arctic._conn
-            if arctic_conn is self._curr_conn:
+            if arctic_conn is not self._curr_conn:
                 self._auth(arctic_conn[self.database_name])  # trigger re-authentication if Arctic has been reset
                 self._curr_conn = arctic_conn
         return self.arctic._conn[self.database_name]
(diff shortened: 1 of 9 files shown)


dimosped:
@bmoscon @jamesblackburn 
We need to make a new Arctic release with the Auth retry fix ASAP.

I implemented the logic which fixes this and unblocks us, but I  was not sure for the items marked as TODO.

If you/James are happy with them, I can implement them with an incremental commit and include in this release (v.1.56.0).

arctic/arctic.py
```diff
@@ -408,7 +408,8 @@ def __init__(self, arctic, library):
         database_name, library = self._parse_db_lib(library)
         self.library = library
         self.database_name = database_name
-        self._auth(self.arctic._conn[self.database_name])
+        self._auth(self._curr_conn[self.database_name])  # authenticate
+        self.get_top_level_collection()  # allow early caching of the library in Arctic (assures _reset is called)
```
jamesblackburn: Not sure this can make a difference as the various library `__init__` methods we have call `get_top_level_collection()` implicitly.  So I suspect something else is going on...

dimosped: certainly no effect here, just updated for sake of simplicity to avoid an extra unnecessary call, since arctic._conn is a property wrapper

dimosped:
@jamesblackburn @bmoscon 
I suggest I withdraw this PR, #476  applies the intended fix

dimosped:
The only extra feature it adds is to retry authentication withitn mongo_retry().
This is triggered only when OperationFailure is raised (can make it more specific to also look for code 13).

Since the Arctic's code has changed so that we only trigger auth when Arctic's connection is reset,
if for any reason now the auth fails (mongodb-related issue, I believe we've seen sometimes this happening) , or database credentials change during runtime, it won't recover.

It is not necessary to merge this feature for now, as we haven't observed/verified the actual issue it addresses.
It 
may worth though to keep the branch and have the fix available in the future.

PR #476, opened 2017-12-21
Fix multiprocessing auth

jamesblackburn:


arctic/arctic.py
```diff
@@ -144,12 +144,14 @@ def _conn(self):
             return self.__conn
 
     def reset(self):
+        logger.debug("Arctic.reset()")
```
bmoscon: can't tell if this is an actual issue or a github formatting issue, but this line looks to be indented incorrectly.

jamesblackburn: Looks ok to me...

bmoscon: ok, on my monitoir it had one space more than the line below it. :)

dimosped:
there is also:
https://github.com/manahl/arctic/pull/477/files

arctic/arctic.py
```diff
@@ -447,6 +449,7 @@ def _auth(self, database):
             authenticate(database, auth.user, auth.password)
 
     def reset_auth(self):
+        logger.debug("reset_auth() %s" % self)
```
dimosped: we'd better move it in _auth() 
the reset_auth is not being used (much).

dimosped: since we no longer spam the auth with every call, it won't add much of a perf hit

jamesblackburn: I'm already getting logs for calls into `_auth` as `authenticate` does a debug log on each call

PR #477, opened 2017-12-21
integration test for auth fix with multiprocessing

dimosped:
Didn't want to use Mocks in an integration test, used an actual auth hook

### DIFF
--- a/tests/integration/test_arctic_multithreading.py
+++ b/tests/integration/test_arctic_multithreading.py
@@ -41,6 +41,7 @@ def my_auth_hook(host, app_name, database_name):
     global AUTH_COUNT
     AUTH_COUNT += 1
 
+
 @pytest.mark.timeout(300)
 def test_multiprocessing_safety(mongo_host, library_name):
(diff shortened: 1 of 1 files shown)


dimosped:
@jamesblackburn @bmoscon  note this is a PR with Jame's branch as target

tests/integration/test_arctic_multithreading.py
```diff
@@ -6,14 +6,20 @@
 
 from arctic import Arctic, VERSION_STORE
 from arctic.exceptions import LibraryNotFoundException
+from arctic.hooks import register_get_auth_hook
 from arctic.store.version_store import VersionStore
 
 MY_ARCTIC = None  # module-level Arctic singleton
+AUTH_COUNT = 0
 
-
```
jamesblackburn: PEP-8 gets a bit sad if you don't have two blank lines before top-level functions

dimosped: fixed

PR #478, opened 2017-12-21
ENH: Add min date function

bashtage:
Add min date function to match max date function

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -754,6 +754,8 @@ class TickStore(object):
         """
         res = self._collection.find_one({SYMBOL: symbol}, projection={ID: 0, END: 1},
                                         sort=[(START, pymongo.DESCENDING)])
+        if res is None:
+            raise NoDataFoundException("No Data found for {}".format(symbol))
         return utc_dt_to_local_dt(res[END])
 
     def min_date(self, symbol):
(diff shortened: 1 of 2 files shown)


bashtage:
I found this missing when attempting to merge from another datasource wanting to limit queries to match only the range of the tick data I have.  

There didn't seem to be any tests of `max_date`, so I haven't added any here.

bmoscon:
@bashtage thanks! We do like unit/integration tests though, so please do add one for this. All new code should have associated tests.

arctic/tickstore/tickstore.py
```diff
@@ -749,3 +749,16 @@ def max_date(self, symbol):
...
+        """
+        Return the minimum datetime stored for a particular symbol
+
+        Parameters
+        ----------
+        symbol : `str`
+            symbol name for the item
+        """
+        res = self._collection.find_one({SYMBOL: symbol}, projection={ID: 0, START: 1},
+                                        sort=[(START, pymongo.ASCENDING)])
```
dimosped: do we care about handling any of the below, or we prefer to let the exception bubble up?
 - res is None
 - START in res

maybe:
```
if not res:
     raise NoDataFoundException("No Data found for {}".format(symbol))
```

bashtage:
That would be better than the status quo which raises KeyError

bashtage:
LMK know if you think this is a good place for the test.

bashtage:
@dimosped I have also added nicer errors when there are no min/max dates.

arctic/tickstore/tickstore.py
bashtage: I put this in

PR #482, opened 2018-01-03
Code cleanup

bmoscon:
Remove unused imports, trailing whitespace, fix overridden built in functions, etc

### DIFF
--- a/arctic/store/bson_store.py
+++ b/arctic/store/bson_store.py
@@ -86,64 +86,64 @@ class BSONStore(object):
         self._arctic_lib.check_quota()
         return self._collection.insert_many(documents, **kwargs)
 
-    def delete_one(self, filter, **kwargs):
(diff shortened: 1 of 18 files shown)


bmoscon:
@jamesblackburn

arctic/store/bson_store.py
```diff
@@ -86,64 +86,64 @@ def insert_many(self, documents, **kwargs):
         self._arctic_lib.check_quota()
         return self._collection.insert_many(documents, **kwargs)
 
-    def delete_one(self, filter, **kwargs):
+    def delete_one(self, query_filter, **kwargs):
```
jamesblackburn: I think we should keep this as `filter` as this matches the upstream pymongo API, and changing the keyword here is a minor API change.

bmoscon:
@jamesblackburn - ok filter change reverted - will merge once tests pass

Issue #488, opened 2018-01-11
fromstring in numpy deprecated in latest version


bmoscon:
`fromstring` has been deprecated in favor of `frombuffer`

PR #489, opened 2018-01-11
Change numpy.fromstring to numpy.frombuffer due to deprecation

bmoscon:


bmoscon:
@jamesblackburn

Cronan:
Does this change the return value? My reading of the dox would say no, the real change is in the details of the memory allocation, but I just wanted to be sure ...

dimosped:
@bmoscon @Cronan @jamesblackburn 
the info in this SO post still seems accurate (up to today):
https://stackoverflow.com/questions/22236749/numpy-what-is-the-difference-between-frombuffer-and-fromstring

Note the memcopy methods used in fromstring:
https://github.com/numpy/numpy/blob/2bacec9ecb0fc665117a22027b73812509a99f56/numpy/core/src/multiarray/ctors.c#L3667
https://github.com/numpy/numpy/blob/2bacec9ecb0fc665117a22027b73812509a99f56/numpy/core/src/multiarray/ctors.c#L3689

Versus the use of the supplied buffer as the backing storage without copy in frombuffer:
https://github.com/numpy/numpy/blob/2bacec9ecb0fc665117a22027b73812509a99f56/numpy/core/src/multiarray/ctors.c#L3570
https://github.com/numpy/numpy/blob/2bacec9ecb0fc665117a22027b73812509a99f56/numpy/core/src/multiarray/ctors.c#L1078

If no separator in supplied in fromstring, it considers the data as binary, and behaves similarly with frombuffer, however it does the extra memcopy. 
If the buffer is mutated later on and frombuffer is used, the changes will be also visible to the returned numpy multiarray.

dimosped:
Fixes #488

bmoscon:
very basic testing by me revealed that it gives the same results. I can instrument a build that does both and compares the results and re-run the unit tests if you like.

dimosped:
it seems to me it is not really necessary:
"_Deprecated since version 1.14: If this argument is not provided, fromstring falls back on the behaviour of frombuffer after encoding unicode string inputs as either utf-8 (python 3), or the default encoding (python 2)._"
https://docs.scipy.org/doc/numpy-1.14.0/reference/generated/numpy.fromstring.html

Since we don't use the "sep"  argument, it is identical to using frombuffer after 1.14.0. 
Looking in the implementation of fromstring / frombuffer in Numpy, there are no changes since 6 years, so the same must hold for older versions if no sep is used (e.g. for numpy 1.9.2 which is 2-3 years old).

Issue #288, opened 2016-11-21
Out-Of-Core and mapping function  over chunks


FaGuoMa:
Tickstore data gets pretty big very quickly and I wonder if seemless mapping of functions to iterate over chunks is either implemented or in the making. We have looked into [Dask](https://github.com/dask/dask) or developing our own but it seems like a potentially common enough issue for Arctic users working with tick data, so asking before re-inventing the wheel

bmoscon:
we are planning on doing something like this for ChunkStore. Any ideas on this for tickstore or versionstore @jamesblackburn ?

jamesblackburn:
Dask integration would be really good!

jengelman:
@jamesblackburn @bmoscon Did anyone make any progress on this?  Would be super useful for my workflow.

bmoscon:
I can probably get this done by end of week

bmoscon:
@jengelman also, i assume you want something like `map()`. It becomes a little less true to the stdlib version in that 1) we need a symbol and 2) we also need ranges, (or an empty range for all). 3) we also have different types of operations (read, write, etc) so exactly what do you want?

I was thinking about two different methods, a `read_map` and a `write_map` that each took a symbol, date/chunk range, and a func to apply. Was this what you had in mind, or something else?

jengelman:
@bmoscon I've been primarily using tickstore, staring to work with chunkstore, and read_map and write_map sound great.  Ideally, read_map on tickstore would take an optional frequency param, so you could do summary statistics on time slices without doing something like copying everything to chunkstores first.

jengelman:
@bmoscon Did you ever getting around to taking a look at this?

bmoscon:
@jengelman not really. I started looking at it but got caught up in the details of the methods. `read_map` sort of makes sense - give it a normal read (which presumably will return multiple chunks) and rather than returning all the chunks as a single df - it will call the provided function on each chunk. i.e. mapping the func over each chunk in the read. I wasn't sure how the write_map should work. You'd provide a dataframe you wanted to write and you'd call the function on each chunk of the dataframe before it was written? That is doable but if the DF gets modified in such a way that it would affect its chunk mapping that gets problematic. just checking for that would make it quite slow (could opt to not check, but it makes me nervous).

bmoscon:
this is done for chunkstore, but it sounds like the requester also wants it for tickstore?

FaGuoMa:
[...]

PR #499, opened 2018-01-28
Issue-288 - Read/Write function mapping

bmoscon:
see #288 

User requested support for applying functions during read/write operations.

@jengelman please take a look and let me know what you think. Is this what you had in mind? Is it even useful? I'm not sure if the mapping on a read is even useful, or should be in arctic since its trivial to implement in a user application. The mapping of a function to a write seems useful since its chunkstore itself that breaks the DF into chunks before writing.


jengelman:
@bmoscon The write bit looks good.  I think you're right about the read-map for Chunkstores, since that's really easy to do client-side.  Ideally, we could have a read-map for Tickstore, since doing the chunking and computation client-side can get complicated/expensive, but that might be getting too far into Dask/etc territory.

PR #502, opened 2018-02-06
fixed the false positive integrity errors and their handling 

dimosped:


arctic/store/_ndarray_store.py
```diff
@@ -36,6 +36,32 @@ def _promote(type1, type2):
     return np.dtype([(n, _promote(dtype1.fields[n][0], dtype2.fields.get(n, (None,))[0])) for n in dtype1.names])
 
 
+def _attempt_update_unchanged(symbol, unchanged_segment_ids, collection, version, previous_version):
+    if not unchanged_segment_ids or not collection or not version:
+        return
+    # Update the parent set of the unchanged/compressed segments
+    result = collection.update_many({
+                                        'symbol': symbol,  # hit only the right shard
```
jamesblackburn: Could you verify that the `.explain(...` on a query of this shape is sane in the sharded context?  (I presume it will be...)

arctic/store/_ndarray_store.py
```diff
@@ -36,6 +36,32 @@ def _promote(type1, type2):
...
+                                                           # update_many is a broadcast query otherwise
+                                        '_id': {'$in': [x['_id'] for x in unchanged_segment_ids]}
+                                    },
+                                    {'$addToSet': {'parent': version['_id']}})
+    # Fast check for success without extra query
+    if result.matched_count == len(unchanged_segment_ids):
+        return
+    # update_many is tricky sometimes wrt matched_count across replicas when balancer runs. Check based on _id.
+    unchanged_ids = set([x['_id'] for x in unchanged_segment_ids])
+    spec = {'symbol': symbol,
```
jamesblackburn: Why a different query `spec` here? 

Wouldn't using the query from above and `set(...)` below give the same answer?

dimosped: Wanted to avoid the integrity check against the set of the ids we got when we started the change.

Rerunning the query will pickup any (unexpected) changes which may have happened in the meanwhile, from current state of Mongo.
This query is efficient in terms of performance, as uses shard key and index.

On the other hand, we can use the same query with minimal risk, 
maybe I got too obsessed with this one.

arctic/store/_ndarray_store.py
dimosped: Seems like it indeed hits only one shard

>>> expl = m['bloomberg.BLOOMBERG']._collection.find({'_id': ObjectId('55972d552914cf59a30ea0d5')}).explain()
>>> len(expl['queryPlanner']['winningPlan']['shards'])
24
>>> expl2 = m['bloomberg.BLOOMBERG']._collection.find({'symbol': 'BBG002RRJNS5', '_id': ObjectId('55972d552914cf59a30ea0d5')}).explain()
>>> len(expl2['queryPlanner']['winningPlan']['shards'])
1

dimosped: Also:

>>> expl2['queryPlanner']['winningPlan']['stage']
u'SINGLE_SHARD'

>>> expl['queryPlanner']['winningPlan']['stage']
u'SHARD_MERGE'

Issue #508, opened 2018-02-23
VersionStore may retrieve old versions of the data when run in multiple processes


aflag:
#### Arctic Version

```
1.59.0
```

#### Arctic Store

```
VersionStore
```

#### Platform and version

Python 2.7.11

#### Description of problem and/or code sample that reproduces the issue

VersionStore's list_symbols, read and read_metadata (when called with a datetime) may return an old version of the data. This will only happen if multiple instances of VersionStore, using different bson.ObjectId objects (eg. the instances are running in different processes), append or write to the same symbol within the same second.

Although the version property is synchronised between the two different processes, it is not taken into consideration when getting the latest version of a symbol when calling list_symbols, read and read_metadata.

PR #509, opened 2018-02-26
Issue #508 - List symbols returning deleted symbols

aflag:
see #508

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ### 1.61
   * Feature: #288 Mapping reads and writes over chunks in chunkstore
+  * Bugfix: #508 VersionStore: list_symbols and read now always returns latest version
 
 ### 1.60 (2018-2-13)
   * Bugfix: #503 ChunkStore: speedup check for -1 segments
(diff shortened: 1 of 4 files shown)


tests/unit/store/test_version_store.py
```diff
@@ -238,6 +238,22 @@ def test_snapshot():
                                                 call('foo', as_of=None, read_preference=ReadPreference.PRIMARY)])
 
 
+def test_list_symbols_default_pipeline():
+    versions = Mock()
+    vs = create_autospec(VersionStore, _versions=versions)
+    versions.aggregate.return_value = []
+
+    VersionStore.list_symbols(vs)
```
bmoscon: whats the purpose of this? this returns a list right? it doesnt seem like the return value is being captured or used.

CHANGES.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog
 
+### 1.61
+  * Bugfix: #508 VersionStore: list_symbol and read now always returns latest version
```
jamesblackburn: `list_symbols()`

tests/unit/store/test_version_store.py
jamesblackburn: Looks like the unit test is just asserting that pymongo aggregation pipeline is being called correctly.  For this mocked case the `list_symbols` will return the `return_value` from above.

PR #513, opened 2018-02-28
Fix slow list versions

dimosped:
We made unnecessarily multiple calls to _find_snapshots, while we could pre-compute and avoid pushing lots of mongo queries:
https://github.com/manahl/arctic/blob/master/arctic/store/version_store.py#L262

For collections with many symbols, many versions and many snapshots, the improvement is very significant with this fix:

```
Without optimisation:
Total time is: 80.6683588028
Count stats: mean=12.95, std=0.569, min=6, max=13
Time stats: mean=0.083851, std=0.042957, min=0.037374, max=0.484092

With Arctic fix:
Total time is: 11.6488740444
Count stats: mean=12.95, std=0.569, min=6, max=13
Time stats: mean=0.012107, std=0.003715, min=0.008276, max=0.067882
```

### DIFF
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -247,16 +247,13 @@ class VersionStore(object):
                 raise NoDataFoundException('No snapshot %s in library %s' % (snapshot, self._arctic_lib.get_name()))
 
         versions = []
-        snapshots = None
+        snapshots = {ss.get('_id'): ss.get('name') for ss in self._snapshots.find()}
         for symbol in symbols:
             query['symbol'] = symbol
             seen_symbols = set()
             for version in self._versions.find(query, projection=['symbol', 'version', 'parent', 'metadata.deleted'], sort=[('version', -1)]):
                 if latest_only and version['symbol'] in seen_symbols:
                     continue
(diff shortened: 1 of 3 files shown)


arctic/store/version_store.py
```diff
@@ -247,19 +247,23 @@ def list_versions(self, symbol=None, snapshot=None, latest_only=False):
...
+        snapshots = None
         for symbol in symbols:
             query['symbol'] = symbol
             seen_symbols = set()
             for version in self._versions.find(query, projection=['symbol', 'version', 'parent', 'metadata.deleted'], sort=[('version', -1)]):
                 if latest_only and version['symbol'] in seen_symbols:
                     continue
+                if snapshots is None:
+                    # populate only once and if is really going to be used
+                    snapshots = {ss.get('_id'): ss.get('name') for ss in self._snapshots.find()}
```
jamesblackburn: I think this will always be hit - just always run above?

dimosped: sounds right, is an overkill to only avoid overhead on empty collections
moved it up

Issue #519, opened 2018-03-13
list_symbols fails with libraries which have large number of symbols and versions


dimosped:
#### Arctic Version
```
1.61.0
```

#### Arctic Store
```
VersionStore
```

#### Platform and version

Python 2.7.11 and 3.x (MongoDB 3.2.11)

#### Description of problem and/or code sample that reproduces the issue

To reproduce create a library with ten a library with a very large number of versions and symbols:
```
In[7]: l._versions.count()
Out[7]: 104017

In[10]: len(l._versions.distinct(key='symbol'))
Out[10]: 9988
```

and then list symbols:

```
l.list_symbols()

2018-03-08 19:52:57,979 - WARNING PID:  7604 - <class 'pymongo.errors.OperationFailure'> list_symbols [Sort exceeded memory limit of 104857600 bytes, but did not opt in to external sorting. Aborting operation. Pass allowDiskUse:true to opt in.], retrying 3
2018-03-08 19:52:58,905 - WARNING PID:  7604 - <class 'pymongo.errors.OperationFailure'> list_symbols [Sort exceeded memory limit of 104857600 bytes, but did not opt in to external sorting. Aborting operation. Pass allowDiskUse:true to opt in.], retrying 4
2018-03-08 19:53:00,048 - WARNING PID:  7604 - <class 'pymongo.errors.OperationFailure'> list_symbols [Sort exceeded memory limit of 104857600 bytes, but did not opt in to external sorting. Aborting operation. Pass allowDiskUse:true to opt in.], retrying 5
2018-03-08 19:53:01,168 - WARNING PID:  7604 - <class 'pymongo.errors.OperationFailure'> list_symbols [Sort exceeded memory limit of 104857600 bytes, but did not opt in to external sorting. Aborting operation. Pass allowDiskUse:true to opt in.], retrying 6
2018-03-08 19:53:02,281 - WARNING PID:  7604 - <class 'pymongo.errors.OperationFailure'> list_symbols [Sort exceeded memory limit of 104857600 bytes, but did not opt in to external sorting. Aborting operation. Pass allowDiskUse:true to opt in.], retrying 7
2018-03-08 19:53:03,387 - WARNING PID:  7604 - <class 'pymongo.errors.OperationFailure'> list_symbols [Sort exceeded memory limit of 104857600 bytes, but did not opt in to external sorting. Aborting operation. Pass allowDiskUse:true to opt in.], retrying 8
[...]

PR #520, opened 2018-03-13
Fix for list_symbols reaching memory limit of MongoDB

dimosped:
Fix for issue #519 with list_versions memory limit exceeded


arctic/store/version_store.py
```diff
@@ -133,6 +133,9 @@ def list_symbols(self, all_symbols=False, snapshot=None, regex=None, **kwargs):
             query['symbol'] = {'$regex': regex}
         if kwargs:
             for k, v in six.iteritems(kwargs):
+                # TODO: this doesn't work as expected as it ignores the versions with metadata.deleted set
+                #       as a result it will return symbols with matching metadata which have been deleted
+                #       Maybe better add a match step in the pipeline instead of making it part of the query
```
bmoscon: if this doesnt work, should we remove it?

arctic/store/version_store.py
```diff
@@ -148,22 +151,32 @@ def list_symbols(self, all_symbols=False, snapshot=None, regex=None, **kwargs):
...
+                    }
+                },
+            }},
+            # Don't include symbols which are part of some snapshot, but really deleted...
+            {'$match': {'version_custom': {'$mod': [2, 0]}}}
+        ])
+
+        # We may hit the group memory limit (100MB), so use allowDiskUse to circumvent this
+        #  - https://docs.mongodb.com/manual/reference/operator/aggregation/group/#group-memory-limit
+        return sorted([x['_id'] for x in self._versions.aggregate(pipeline, allowDiskUse=True)])
```
bmoscon: how much latency does this add?

arctic/store/version_store.py
dimosped: We had a brief chat on this.

Seems that apart from some integration tests, users don't typically pass metadata args with list_symbols. Is this a valid assertion?

We can either remove this functionality and change the api/behaviour with this PR, or fix it with another one.
IMHO the fix should only touch/modify the pipeline if metadata arguments are passed, i.e. remain unmodified as-is if no metadata args are present. An extra $match stage could do.

arctic/store/version_store.py
dimosped: We only hit this if the sorting/grouping mem footprint at mongo exceeds 100MB.
With the new pipeline, we don't hit it with our largest libs (100K+ version docs).

I noticed a query which was previously failing with "memory exceeded" OperationFailure, 
took 4sec to complete.
The equivalent query without the sorting part, was ~1.5sec faster.
This totally depends on how heavy is the query.

dimosped: The 'allowDiskUse' is there now as a gueard dog, to avoid failure when/if memory is exceeded.

PR #522, opened 2018-03-19
issue 521 - updated setup.py for clang-6.0

dcwtx:
clang-6.0 includes openmp support by default, so the linker flag is no longer required so long as the compiler flag was included.  Therefore in [setup.py](https://github.com/manahl/arctic/blob/master/setup.py) make the default `link_args` empty at line 29, and if gcc is the installed compiler add the `link_arg` `-fopenmp` at line 50.  Also updated the ValueError to reference clang-6.0 (now default install when executing `brew install llvm`).

As I reviewed this before submission, I see that I could have deleted the comment at line 32 also as this is not true. _(fixed in subsequent commit)_

### DIFF
--- a/README.md
+++ b/README.md
@@ -117,10 +117,8 @@ Arctic currently works with:
 
 
 Operating Systems:
-
  * Linux
  * macOS
-
 
 ## Acknowledgements
 
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Arctic storage implementations are **pluggable**.  VersionStore is the default.
 
 ## Requirements
 
-##### Arctic currently works with:
+Arctic currently works with:
 
  * Python 2.7, 3.4, 3.5, 3.6
  * pymongo >= 3.0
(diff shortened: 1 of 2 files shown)


bmoscon:
@dcwtx all the builds failed, so you might want to back out part of the change. I believe its related to the link_arg changes.

dcwtx:
Yep, noticed that and have been trying to track down the problem.  Both gcc and llvm are building _compress.so, but in the llvm version it does not contain the symbol that _compression.py is looking for.  I went back to ground zero and installed clang-5.0 instead, as the ValueError text in setup.py indicated that should work (note current default for homebrew is clang-6.0), but got the same link error.

I'm going to keep working on it but it seems to me that at present clang is not working with openmp, as noted in the comment on line 32 of the original version of the file.

dcwtx:
Ok, going to push a revised version.  Revised setup.py to move `link_args=["-fopenmp"]` to the top, and revised the change to link_args in the case of using clang to `link_args=["-fopenmp=libomp"]`.  Additionally, added a check if using clang to see if `/usr/local/opt/libomp` exists - if not, added a ValueError exception to prompt to install libomp using `brew install libomp`.

Now when I run `python setup.py install` then `python setup.py test` I am getting the same test output regardless of whether gcc or llvm is installed.  Will push the change now to see how the automated test package does.

dcwtx:
I went ahead and updated the README.md to add compilation information to the front page.  I came into this because Rob has included Arctic in pysystemtrade and I was not able to compile it when I pulled the new repo.  If this info had been on the front page it would have saved me a lot of time... but then compilation with llvm wouldn't work...

bmoscon:
thanks - I'll look over this later today. I did post an issue in `pysystemtrade` asking if there is a reason its not using the `arctic` pypi source for the package, since there are wheels in pypi.

bmoscon:
I'm fine with the `setup.py` changes but I'll wait for @jamesblackburn to opine on the changes to the readme. I don't mind them, but I don't think they are necessary. The errors you get from building on macOS should be sufficient, plus we also don't list out all the steps to build on Linux either. If users arent comfortable building source code and working through their platform issues wheels are provided on pypi. So - I'll leave it up to James

dcwtx:
Saw your note to Rob and his response. I think your comments about the readme make sense, and given the availability of the wheels this is probably overkill. It can be figured out from the setup.py. Less is more.

bmoscon:
@dcwtx if you back out the changes to the readme I'll merge it in

PR #525, opened 2018-03-29
Issue 510: Support for multi columns in DataFrames

aflag:


bmoscon:
@aflag you'll need to either pin the pytest version lower or fix the conftest issue

Issue #6, opened 2015-06-09
Cannot compile on Windows


beartastic:
I installed the recommended C++ Compiler for Python 2.7, however I can't seem to run the installer. 

The fatal error I got is:
C1083: Cannot open include file: 'stdint.h': No such file or directory

Is it possible to release a compiled version of this?

beartastic:
Please ignore. I was able to compile and run the code, but had to manually copy all the modules within Arctic to the site-packages directory.

Many thanks!

jamesblackburn:
Oh super! What did you have to do? We don't run this code on Windows here, so anything we can do to make this seamless would be good...

beartastic:
Here are my steps:
1. Install MinGW. Make sure to install the following packages:
Under Basic Setup:
mingw32-base, mingw32-gcc-g++
Under All Packages:
mingw32-pthreads-w32

2. Make sure the mingw bin directory is inside your PATH environment

3. Go into your Python27 directory then under Lib\distults\distutils.cfg, add the following:

[build]
compiler=mingw32

4. Do a pip install for numpy, pandas.

5. Finally do the install for arctic. Fingers crossed you won't get any error messages.

6. Create a simple python script that does an import to Arctic. If you get an error when running the code, which I did get, can try this solution: I had to directly copy the rest of the subdirectories with the .py files from the installation source directly into the site-packages/arctic folder myself. I am not sure if this is a bug with the installer or I am just doing it incorrectly.

Either way it seems to run now!

jamesblackburn:
Would be great to get this automated so we can have some windows eggs / wheels to upload to PyPI

agiap:
I am getting the same error message as beartastic ("fatal error C1083: Cannot open include file: 'stdint.h': No such file or directory". I followed his steps but it didn't help. Any idea? Thanks.

jamesblackburn:
We don't technially require the native bits (they just provide parallel compression), you can run arctic without them as per:
https://github.com/manahl/arctic/issues/52#issuecomment-159011085

NickAltmann:
I had two problems getting up and running on Windows.

The first was with the lz4 dependency.  This was failing because the compile uses gcc specific command line arguments, causing the MSVC compile to fail.  Not sure if I'm missing something, I've posed the question [here](http://stackoverflow.com/questions/35862884/windows-compilation-failure-on-pip-install-due-to-invalid-command-flags).

[...]

PR #528, opened 2018-04-09
BLD: Correct compiler flags for Windows

bashtage:
Pass correct compiler flags for Windows


bashtage:
Should be enough to close #6 -- Master correctly builds on Windows 2.7 and 3.6, x64.

bashtage:
Yes. These only hit the windows build. The compiler flags are unchanged for
Linux is OSX.

setup.py
```diff
@@ -115,7 +119,8 @@ def extensions():
     from Cython.Build import cythonize
     return cythonize(Extension('arctic._compress',
                                sources=["src/_compress.pyx", "src/lz4.c", "src/lz4hc.c"],
-                               extra_compile_args=['-fopenmp', '-fpermissive'], # Avoid compiling error with prange. Similar to http://stackoverflow.com/questions/36577182/unable-to-assign-value-to-array-in-prange
+                               # Avoid compiling error with prange. Similar to http://stackoverflow.com/questions/36577182/unable-to-assign-value-to-array-in-prange
```
jamesblackburn: Presumably this should be hoisted up to line 30 where `-fpermissive` is defined?

jamesblackburn: Otherwise LGTM!

bashtage: Moved the comment to be by the flag.

PR #530, opened 2018-04-10
PERF: Disable boundscheck and wrap around

bashtage:
Disable boundscheck and wraparound in compression
to avoid unnecessary checks that require acquiring the
GIL inside OpenMP blocks


bashtage:
I couldn't get ASV to run but local testing indicated a 2% improvement (running the ASV tests directly in IPython).  Seems like a free lunch.  Might be worth more on systems with more cores since grabbing the GIL created thread contention.

jamesblackburn:
LGTM - @bmoscon happy?

Issue #168, opened 2016-07-13
Arctic converts all values to NaN when writing a single column dataframe with a column name of ‘’


TomTaylorLondon:
```
df = pd.DataFrame(data=range(3), index=range(3))
df  = df * 1.0
df.columns = ['']
print df
> 0  0
> 1  1
> 2  2
lib.write('df', df)
lib.read('df').data

> index     
> 0      NaN
> 1      NaN
> 2      NaN

```

bmoscon:
how do we want to work around this @jamesblackburn @richardbounds ? I assume that the empty name means when we rebuild from records, we have a proper size, but now improper number of cols (due to missing name), so it fills empty col with NaNs in pandas. Off the top of my head we could

1). have a sentinel for an empty column, store with sentinel and rename after reconstructing. 
2). throw an error if column name is empty string
3). handle it like with do with missing index names and give it a default name (col1, col2, col3, etc).


thoughts? suggestions?

bmoscon:
@jamesblackburn just bringing this up again. At a minimum we should disallow columns with an empty string name

bmoscon:
@TomTaylorLondon @jamesblackburn any thoughts about how you want to proceed?

bmoscon:
closed via #548

PR #548, opened 2018-04-21
Raise exception if empty string given as column name (Issue #168)

srf94:


srf94:
Just checked, if you comment out the two lines that raise the exception in numpy_records.py, the test fails as needed.

bmoscon:
@srf94 can you add a line to the changelog indicating which bug this fixes?

PR #549, opened 2018-04-21
Better datetime to ms

srf94:


bmoscon:
The test fails :(

srf94:
After discussion, we will truncate microseconds, not round to nearest millisecond.

PR #550, opened 2018-04-21
MacOS Fix build with conda gcc

jamesblackburn:
Latest conda gcc support libomp on MacOS X.  If gcc is on the PATH then use that.


setup.py
```diff
@@ -23,35 +23,39 @@
...
-    os.environ["CC"] = cc if 'clang' in cc else cc.replace("g++", "gcc")
-    os.environ["CXX"] = cc
+    # if a recent compiler on PATH (e.g. from anaconda) then let's use that
+    gcc_out = subprocess.check_output(['gcc', '-v'], stderr=subprocess.STDOUT)
+    if b'LLVM' in gcc_out:    
+        # clang on macOS does not work with OpenMP
+        ccs = ["/usr/local/bin/g++-5",
+               "/usr/local/bin/g++-6",
+               "/usr/local/bin/g++-7",
+               "/usr/local/opt/llvm/bin/clang"]
```
aflag: Shouldn't anaconda's gcc be added here? Also, if there's LLVM in the output, should we consider it to be clang and pass -fopenmp=libomp paramater as defined below?

jamesblackburn: Essentially we don't enter this code path if the `gcc` is on the PATH (i.e. likely anaconda / brew gcc).  By default `gcc` on Mac OS X is LLVM

PR #555, opened 2018-04-21
Circleci v2

jamesblackburn:
It mostly builds with this.  Issues:
 - slow - is the caching working right?
 - uses large circleci image - the small ones don't have libssl => breaks mongo
 - tests don't pass (that lz4 issue in the other ticket...)


bmoscon:
@jamesblackburn i fixed quite a few issues - it builds and will sometimes complete all the tests. It seems to either timeout on a test and the whole thing ends, or it will have time outs on individual tests and have a few failures. I'm personally all for just dropping circle ci. It seems like we'd need to revamp the pytest fixtures to not create/destroy mongo instances on every single test to fix this (since its been timing out waiting for mongo to start up).

PR #554, opened 2018-04-21
fix failed tests due to the new lz4 api

dimosped:
Fixes issue #553

PR #562, opened 2018-05-11
Add advanced query for list_symbols() similar to VersionStore

egao1980:
Adding missing features:
- regex symbol lookup
- as_of support - show symbols valid starting with the provided date
- metadata field exact match queries

### DIFF
--- a/arctic/store/metadata_store.py
+++ b/arctic/store/metadata_store.py
@@ -77,32 +77,40 @@ class MetadataStore(BSONStore):
          -------
          String list of symbols in the library
         """
-        query = {}
+        index_query = {}
+        data_query = {}
+
         if regex is not None:
-            query['symbol'] = {'$regex': regex}
+            index_query['symbol'] = {'$regex': regex}
 
         if as_of is not None:
-            query['start_time'] = {'$lte': as_of}
+            index_query['start_time'] = {'$lte': as_of}
 
         if kwargs:
             for k, v in six.iteritems(kwargs):
-                query['metadata.' + k] = v
+                data_query['metadata.' + k] = v
 
-        if len(query):
-            pipeline = []
-            if query:
-                # Match based on user criteria first
-                pipeline.append({'$match': query})
-
-            pipeline.extend([
-                {'$sort': {'start_time': pymongo.DESCENDING}},
-                # Group by 'symbol'
-                {'$group': {'_id': '$symbol'}},
-                {'$project': {'_id': 0, 'symbol':  '$_id'}}])
-
-            return sorted(r['symbol'] for r in self.aggregate(pipeline))
-        else:
+        # Skip aggregation pipeline
+        if not(index_query or data_query):
             return self.distinct('symbol')
+
+        pipeline = []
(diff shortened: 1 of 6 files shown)


arctic/store/metadata_store.py
```diff
@@ -57,8 +59,50 @@ def __repr__(self):
...
+            query['symbol'] = {'$regex': regex}
+
+        if as_of is not None:
+            query['start_time'] = {'$lte': as_of}
+
+        if kwargs:
+            for k, v in six.iteritems(kwargs):
+                query['metadata.' + k] = v
+
+        if len(query):
```
aflag: Maybe we can change this to just `if query`

arctic/store/metadata_store.py
```diff
@@ -57,8 +59,50 @@ def __repr__(self):
...
+            pipeline = []
+            if query:
```
aflag: This will always be true

arctic/store/metadata_store.py
```diff
@@ -57,8 +59,50 @@ def __repr__(self):
...
+                # Match based on user criteria first
+                pipeline.append({'$match': query})
+
+            pipeline.extend([
+                {'$sort': {'start_time': pymongo.DESCENDING}},
+                # Group by 'symbol'
+                {'$group': {'_id': '$symbol'}},
```
aflag: Can we use $max on the start_time here instead of sorting? https://docs.mongodb.com/manual/reference/operator/aggregation/max/#grp._S_max

The current sort is not hitting the index (('symbol', 1), ('start_time', -1)) so it will do a full scan, sorting all documents in the collection every time this is called. If we group by start_time we are still going over the entire collection, but at least at a cost of O(N)

arctic/store/metadata_store.py
```diff
@@ -57,8 +59,50 @@ def __repr__(self):
...
+
+         Parameters
+         ----------
+         as_of : `datetime.datetime`
+            filter symbols valid at given time
+         regex : `str`
+             filter symbols by the passed in regular expression
+         kwargs :
+             kwarg keys are used as fields to query for symbols with metadata matching
+             the kwargs query
```
aflag: Maybe add documentation saying that this operation will scan over all documents in the collection if any of these parameters are passed.

I understand that the operation self.distinct(symbol) is not that fast either, but at least it should be a covered query. Therefore, hopefully, using only the index and not loading every document into memory.

dimosped:
@egao1980  can you also do something similar to this for the failing compression tests?
https://github.com/manahl/arctic/pull/554/files

arctic/store/metadata_store.py
egao1980: Will simplify code a bit

arctic/store/metadata_store.py
egao1980: I've updated the aggregation pipeline to avoid full scan as much as possible.

egao1980:
It looks like tz related tests are failing due to changes in the latest pandas version. Fixing those I think is outside of the scope of this change

Issue #566, opened 2018-05-22
pandas 0.23 issues


bmoscon:
pandas 0.23 introduces some changes to timezone handling that break some of the integration tests

bmoscon:
fixed

PR #567, opened 2018-05-23
Fix issues with pandas 0.23

bmoscon:
latest version of pandas introduces some changes that affect how timezones are stored and used in `pandas.Timestamp`. This corrects the issues in the unit tests and in tickstore. See issue #566

### DIFF
--- a/arctic/multi_index.py
+++ b/arctic/multi_index.py
@@ -93,8 +93,6 @@ def groupby_asof(df, as_of=None, dt_col='sample_dt', asof_col='observed_dt'):
     '''
     if as_of:
         if as_of.tzinfo is None and df.index.get_level_values(asof_col).tz is not None:
-            print(as_of.tzinfo)
-            print(df.index.get_level_values(asof_col).tz)
             as_of = as_of.replace(tzinfo=pytz.utc)
     return fancy_group_by(df,
                           grouping_level=dt_col,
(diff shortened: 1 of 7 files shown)


bmoscon:
will squash the commits to a single commit

bmoscon:
@jamesblackburn 
@aflag 

would appreciate a quick review before I merge this in!

arctic/multi_index.py
```diff
@@ -90,6 +91,9 @@ def groupby_asof(df, as_of=None, dt_col='sample_dt', asof_col='observed_dt'):
     asof_col: ``str`` or ``int``
         Name or index of the column in the MultiIndex that is the observed date
     '''
+    if as_of:
+        if as_of.tzinfo is None and df.index.get_level_values(asof_col).tz is not None:
+            as_of = as_of.replace(tzinfo=pytz.utc)
```
dimosped: would something like this be better to align tzones ?
```
df_idx_tz = df.index.get_level_values(asof_col).tz
if as_of.tzinfo is None and df_idx_tz is not None:
   as_of = as_of.replace(tzinfo=df_idx_tz)
```

or 

```
df_idx_tz = df.index.get_level_values(asof_col).tz
if as_of.tzinfo is None and df_idx_tz is not None:
   as_of = as_of.replace(tzinfo=pytz.utc).astimezone(df_idx_tz)
```

jamesblackburn: We generally assume that naive datetimes are 'local' don't we?

bmoscon: good point - other code does just convert/force things to local tz, so I've mirrored that.

Issue #571, opened 2018-06-07
De-cythonize Arctic


dimosped:
#### Arctic Version

```
1.57.0
```

#### Arctic Store

```
All stores
```

#### Description of problem and/or code sample that reproduces the issue

Arctic is using Cython to use true-parallel compression (releasing the Gil).

This however makes the installation of Arctic cumbersome, its build slow, and we don't receive the latest/greatest from LZ4 implementation.

Remove arctic's Cython code, and switch to using vanilla python-lz4, with Threadpool.

PR #572, opened 2018-06-07
Decythonized lz4

dimosped:
Adresses issues #571 #558 

This removes the cython code, and repalces with the latest-greatest LZ4 which is also faster (due to updated implementation).

We now use a threadpool instead  with a configurable fixed size (experimentation suggests that parallelism returns are diminishing after a given point).

The new python-lz4, releases the GIL :-)
https://github.com/python-lz4/python-lz4/blob/master/lz4/block/_block.c#L190,L218

Also, previously we would spawn a huge number of (new) threads, which would cause parallel version slowdowns, due to the increased overhead of frequent context switching.


arctic/_compression.py
```diff
@@ -20,23 +24,27 @@ def enable_parallel_lz4(mode):
...
-    if len(str_list) > LZ4_N_PARALLEL:
-        return clz4.compressarr(str_list)
-    else:
-        return [clz4.compress(s) for s in str_list]
-
+    if not str_list:
+        return str_list
+
+    if not ENABLE_PARALLEL or len(str_list) <= LZ4_N_PARALLEL:
+        # Less than 50 chunks its quicker to compress sequentially..
```
aflag: If we change LZ4_N_PARALLEL this comment will become outdated. Maybe we can say "Parallel compression is only worth it when we have a lot of chunks" or something like that.

dimosped: yes, correct.
Once I find the sweet spot I will also change this comment

bmoscon:
finally - no more cython - no more platform specific wheels . . . :)

tests/unit/test_compression.py
```diff
@@ -20,16 +20,16 @@ def test_compressarr():
 
 def test_compress_array_usesLZ4():
     cfn = Mock()
-    with patch('arctic._compression.clz4.compressarr', cfn):
+    with patch('arctic._compression.lz4_compress', cfn):
         compress_array([b"foo"] * 100)
-        assert cfn.call_count == 1
+        assert len(cfn.call_args_list) == 100  # call_count is not thread safe
```
dimosped: we've been biten by this intermittently failing test.
Mock's call count is not reliable with mutlithreading.
The increment is not a single Python vm instruction/atomic:
https://github.com/python/cpython/blob/master/Lib/unittest/mock.py#L957
and
https://stackoverflow.com/questions/39332139/thread-safe-version-of-mock-call-count

A trivial example:
```
import multiprocessing as mp
from mock import Mock

mymock = Mock()

def Process(n):
    for i in range(n):
        mymock.visit()

n_threads = 5
n_calls = 10000
result=mp.pool.ThreadPool(n_threads).map(Process, [n_calls]*n_threads)
print(mymock.visit.call_count)
assert mymock.visit.call_count == n_threads * n_calls
```

dimosped:
```
>>> import dis
... def f(x):
...     x+=1
...     return x
... dis.dis(f)
  3           0 LOAD_FAST                0 (x)
              3 LOAD_CONST               1 (1)
              6 INPLACE_ADD         
              7 STORE_FAST               0 (x)

  4          10 LOAD_FAST                0 (x)
             13 RETURN_VALUE
```

 "i += 1" is not an atomic python Interpreter instruction
and GIL may be released in between the load and the INPLACE_ADD/STORE_FAST

jamesblackburn:
Do we need to assert a certainly lz4 version in the setup.py now?

dimosped:
@jamesblackburn  @bmoscon  there is no need to pin LZ4 versions, as we are backwards compatible:
https://github.com/dimosped/arctic/blob/decythonized-lz4/arctic/_compression.py#L5,L9

Also this is the new decision logic for parallel compression:
https://github.com/dimosped/arctic/blob/decythonized-lz4/arctic/_compression.py#L93,L94

The experimental methodology to tune the params is here, along with results:
https://github.com/dimosped/arctic/tree/decythonized-lz4/benchmarks/lz4_tuning

arctic/_compression.py
```diff
@@ -1,13 +1,17 @@
...
 
+from lz4.block import compress as lz4_compress, decompress as lz4_decompress
 
 logger = logging.getLogger(__name__)
 
 # switch to parallel LZ4 compress (and potentially other parallel stuff), Default True
 ENABLE_PARALLEL = not os.environ.get('DISABLE_PARALLEL')
-LZ4_N_PARALLEL = 50  # No. of elements to use parellel compression in LZ4 mode
+LZ4_N_PARALLEL = os.environ.get('LZ4_N_PARALLEL', 50)  # No. of elements to use parallel compression in LZ4 mode
+LZ4_WORKERS = os.environ.get('COMPRESSION_WORKERS', 4)
```
jamesblackburn: I thought `2` workers was a reasonable sweet spot for this?

dimosped: it is 2x workers for non-HC, and 4 (or better 8) for HC.

We could either have 2 thread pools (one for default with 2x and one for HC with 8x) which I don't like.
So I chose a middle ground of 4.

Since VersionStore doesn't use HC, the users can specify env:  COMPRESSION_WORKERS=2
For TickStore heavy use, users may set COMPRESSION_WORKERS=8.

I might add this as a comment somewhere

PR #574, opened 2018-06-14
add host to VersionedItem

scriada:
This is useful when working with multiple stores across different hosts, as all the necessary information to reference/load an item is embedded in the `VersionedItem`.


bmoscon:
I'm not sure why you want this - you need a new Arctic instance to reference a new host, no? You couldnt access this data without knowing the host already, right?

richardbounds:
I think the general requirement is: we've got a VersionedItem that came from an Arctic library somewhere and we'd like to be able to see where it came from.

It's a bit clunky having separate library and host attributes - it would be neat to use 'user.library@host' for the library name (as we already use that format for resolving libraries elsewhere), but since it sounds like that would be a breaking change for lots of existing code, this seems like the least worst alternative.

bmoscon:
Why not store this in information in the user defined metadata?

richardbounds:
We could, but it could potentially get out of sync when host names change, or data is copied around.

bmoscon:
Ah sorry, I didnt mean actually write it to the DB, I meant populate the metadata with this value when returning the VersionedItem to the user. Similar to what the PR has now, except you'd just add the host key/value to the metadata rather than add a new field

dimosped:
@scriada @richardbounds  this is neat

I tend to agree with @bmoscon  that we could make this part of the metadata (added at the client as VersionedItem is returned), instead of changing the VersionedItem tuple.
If we go down that route, then we need to also add some code in the write/append path, to strip out the 'arctic_host' key from the metadata and don't write it to mongo. 
E.g. think of a scenario where we read (vstore adds automatically arctic_host in returned metadata), then write back updated data + metadata.

scriada:
We certainly could put `host` in the metadata, but as @dimosped pointed out this would be fiddly to manage (append/write/copy etc). Also, what if someone saved an item with custom metadata that had a `host` attribute? How would we distinguish from this?

In my mind, `metadata` is user-defined and related to the data itself. Things like `library`, `host`, `version` relate to the storage mechanism itself, which is why they are attributes of `VersionedItem` and not part of the metadata.

dimosped:
@scriada  can you please trigger one more build after rebasing from master?
would be great to see one green build

dimosped:
@bmoscon  @jamesblackburn  are we good with this change ?

dimosped:
@scriada  can you please resolve the conflicts and add a line in the CHANGES.md describing the feature ?
Would be good to include this in the next release.

dimosped:
this implements feature #595

Issue #577, opened 2018-06-20
Implement faster and incremental serialization for VersionStore/Pandas


dimosped:
#### Arctic Version

```
1.67
```

#### Arctic Store

```
VersionStore (NdarrayStore, PandasStore)
```

#### Description of problem and/or code sample that reproduces the issue

Arctic VersionStore's serialization is inefficient in terms of speed and memory footprint.

Especially for large data frames it ends up having 3x copies of the data (original data, in recarrays format, in serialized bytes format) which are maintained in memory until all data have been written to MongoDB.

Experiment and prototype incremental serialization, where only a small batch of segments are serialized each time and removed from memory as soon as they are written to MongoDB.

This work will only deliver the functionality of incremental serializer, and not the Mongo Async I/O part.

Tests must be added to verify the binary compatibility of the existing serializer and the incremental.

PR #578, opened 2018-06-20
Incremental serialization initial

dimosped:
This is a draft PR for #577
Please ignore until the implementation is mature enough to be merged.

### DIFF
--- a/tests/unit/serialization/test_incremental.py
+++ b/tests/unit/serialization/test_incremental.py
@@ -92,7 +92,7 @@ def test_serialize_pandas_to_recarray(input_df):
 @pytest.mark.parametrize("input_df", _test_data().keys())
 def test_serialize_incremental_pandas_to_recarray(input_df):
     incr_ser = IncrementalPandasToRecArraySerializer(df_serializer, _test_data()[input_df][0])
-    chunk_bytes = [chunk_b for chunk_b, _ in incr_ser.generator_bytes]
+    chunk_bytes = [chunk_b for chunk_b, _ in incr_ser.generator_bytes()]
 
     matching = _test_data()[input_df][1][0].tostring() == b''.join(chunk_bytes)
     assert matching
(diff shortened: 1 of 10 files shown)


tests/unit/serialization/test_incremental.py
```diff
@@ -0,0 +1,105 @@
...
+from arctic.exceptions import ArcticSerializationException
+from arctic.serialization.incremental import IncrementalDataFrameToRecArraySerializer
+from arctic.serialization.numpy_records import DataFrameSerializer
+from tests.integration.chunkstore.test_utils import create_test_data
+
+from tests.util import get_large_ts
+
+NON_HOMOGENEOUS_DTYPE_PATCH_SIZE_ROWS = 50
+
+onerow_ts = get_large_ts(1)
```
jamesblackburn: Do just turn these into `pytest.fixtures` - as test discovery imports and runs the module you don't want to be doing anything heavyweight on import.

dimosped: sure, hidden these in a lazy initializer, only invoked upon test invocation

dimosped: @jamesblackburn  @bmoscon   I'd like to merge this if no objections (a couple extra commits since the review)

Merging won't have any impact, as the incremental serializer isn't being used in the code, this only adds the implementation.

Once merged, I'll follow up with a subsequent PR which will enable the async internal  serialization/Mongo design, which will actually use the incremental serializer.

dimosped:
dimosped 5 minutes ago  Member
@jamesblackburn  @bmoscon  I'd like to merge this if no objections (a couple extra commits since the review)

Merging won't have any impact, as the incremental serializer isn't being used in the code, this only adds the implementation.

Once merged, I'll follow up with a subsequent PR which will enable the async internal serialization/Mongo design, which will actually use the incremental serializer.

Issue #579, opened 2018-06-27
VersionStore corrupts symbol after a sequence of restore-version and append


dimosped:
#### Arctic Version

```
1.67.0
```

#### Arctic Store

```
VersionStore
```

#### Description of problem and/or code sample that reproduces the issue

We had multiple corruption incidents after version_restore and then doing append:

```
2018-06-21 15:54:59,014 - WARNING PID:  9346 - <class 'pymongo.errors.OperationFailure'> _do_read [Incorrect number of segments returned for px_last_today:163.  Expected: 88, but got 87. ], retrying 3
--
2018-06-21 15:55:00,978 - WARNING PID:  9346 - <class 'pymongo.errors.OperationFailure'> _do_read [Incorrect number of segments returned for px_last_today:163.  Expected: 88, but got 87. ], retrying 4

```

and 

```
 File "build/bdist.linux-x86_64/egg/arctic/decorators.py", line 57, in f_retry
  File "build/bdist.linux-x86_64/egg/arctic/decorators.py", line 50, in f_retry
  File "build/bdist.linux-x86_64/egg/arctic/store/version_store.py", line 386, in _do_read
  File "build/bdist.linux-x86_64/egg/arctic/store/_ndarray_store.py", line 234, in read
  File "build/bdist.linux-x86_64/egg/arctic/store/_ndarray_store.py", line 264, in _do_read
pymongo.errors.OperationFailure: Incorrect number of segments returned for sid:T:44.  Expected: 17, but got 18. 
```

PR #580, opened 2018-06-27
Fix for the corruption issue of restore-appends

dimosped:
See issue #579 for details.

### DIFF
--- a/arctic/store/_version_store_utils.py
+++ b/arctic/store/_version_store_utils.py
@@ -80,6 +80,23 @@ def _define_compat_pickle_load():
 
 
 def analyze_symbol(l, sym, from_ver, to_ver, do_reads=False):
+    """
+    This is a utility function to produce text output with details about the versions of a given symbol.
+    It is useful for debugging corruption issues and to mark corrupted versions.
+    
+    Parameters
+    ----------
+    l : `arctic.store.version_store.VersionStore`
+        The VersionStore instance against which the analysis will be run.
+    sym : `str`
+        The symbol to analyze
+    from_ver : `int` or `None`
+        The lower bound for the version number we wish to analyze. If None then start from the earliest version.
+    to_ver : `int` or `None`
+        The upper bound for the version number we wish to analyze. If None then stop at the latest version.
+    do_reads : `bool`
+        If this flag is set to true, then the corruption check will actually try to read the symbol (slower).
+    """
     logging.info('Analyzing symbol {}. Versions range is [v{}, v{}]'.format(sym, from_ver, to_ver))
     prev_rows = 0
     prev_n = 0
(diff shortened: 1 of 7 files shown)


dimosped:
@jamesblackburn  @bmoscon  this is a high priority fix.

arctic/store/_version_store_utils.py
```diff
@@ -76,6 +79,87 @@ def _define_compat_pickle_load():
     return functools.partial(pickle_compat.load, compat=True)
 
 
+def analyze_symbol(l, sym, from_ver, to_ver, do_reads=False):
```
bmoscon: can we add a docstring of some sort here to describe the purpose and functionality of this function?

dimosped: sure, docstrings added for all new util methods

arctic/store/version_store.py
```diff
@@ -507,7 +507,11 @@ def append(self, symbol, data, metadata=None, prune_previous_version=True, upser
         next_ver = self._version_nums.find_one_and_update({'symbol': symbol, },
                                                           {'$inc': {'version': 1}},
                                                           upsert=False, new=True)['version']
-        if next_ver != previous_version['version'] + 1:
+
+        # Do a dirty append if previous version is the result of a restore version, or
+        # a chain of write_metadata after a restore_version.
+        # This prevents a well known issue of corrupted data.
+        if next_ver != previous_version['version'] + 1 or previous_version.get('restored_version'):
```
jamesblackburn: Should no longer need this check?

arctic/store/version_store.py
```diff
@@ -710,21 +714,28 @@ def restore_version(self, symbol, as_of, prune_previous_version=True):
...
+        #       and only do concat_rewrite on the first append operation, after the restore (i.e. using a flag).
+        #       Keep in mind the history branch induced corruption (see Issue 
+
+        # Don't create a new version if we are trying to restore to the latest
+        # TODO: this is not atomic, and is prone to race conditions if multiple processes touch the same symbol
         version_to_restore = self._read_metadata(symbol, as_of=as_of)
+        try:
+            latest_version = self._read_metadata(symbol)
+            if version_to_restore['version'] == latest_version['version']:
+                return latest_version
```
jamesblackburn: Is this actually doing a restore - it looks like it's just returning the version?

dimosped: this is a noop really, i.e. when we try to restore to a snapshot/version which is the latest (no write/append/delete/write_metadata since then), then do nothing, and don't cause an unnecessary read/write

arctic/store/version_store.py
dimosped: was left from my original implementation :-/ 
the second part of the condition is unnecessary now

setup.py
```diff
@@ -83,7 +83,7 @@ def run_tests(self):
                       "enum34",
                       "mockextras",
                       "pandas",
-                      "pymongo",
+                      "pymongo>=3.6.0",
```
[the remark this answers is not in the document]
dimosped: this gives now a green build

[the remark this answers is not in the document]
bmoscon: 👍

dimosped:
Yes, this is deprecated, closing it

PR #585, opened 2018-07-11
Release restore hotfix merge on master

dimosped:
This will just merge the hotfix release branch changes onto master.

The hotfix release was branched from 1.67.0 , not from master HEAD, hence required rebasing the hotfix branch on master + conflict resolution before we can merge.

### DIFF
--- a/arctic/store/_version_store_utils.py
+++ b/arctic/store/_version_store_utils.py
@@ -9,6 +9,7 @@ import pandas as pd
 import pymongo
 from bson import Binary
 from pandas.compat import pickle_compat
+from pymongo.errors import OperationFailure
 
 
 def _split_arrs(array_2d, slices):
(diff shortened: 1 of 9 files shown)


arctic/store/_version_store_utils.py
```diff
@@ -76,6 +80,219 @@ def _define_compat_pickle_load():
...
+            # Quick check: compare segment count
+            if total_segments != v.get('segment_count', 0):
+                return True  # corrupted, don't proceed with fetching from mongo the first hit
+            # Quick check: Segment counts agree and size is zero
+            if total_segments == 0:
+                return False
+        
+        if check_last_segment:
+            # Quick check: compare the maximum segment's up_to number. It has to verify the version's up_to.
+            max_seg = collection.find_one(spec, {'segment': 1}, sort=[('segment', pymongo.DESCENDING)])
```
jamesblackburn: Out of interest, what's the explain() on this look like?

arctic/store/_version_store_utils.py
```diff
@@ -76,6 +80,219 @@ def _define_compat_pickle_load():
...
+    try:
+        # Not that commands sequence (a) is slower than (b)
+        # (a) curs = collection.find(spec, {'segment': 1}, sort=[('segment', pymongo.DESCENDING)])
+        #     curs.count()
+        #     curs.next()
+        # (b) collection.find(spec, {'segment': 1}).count()
+        #     collection.find_one(spec, {'segment': 1}, sort=[('segment', pymongo.DESCENDING)])
+
+        if check_count:
+            total_segments = collection.find(spec, {'segment': 1}).count()
```
jamesblackburn: What does the explain on this look like? Worry that having the field spec `{'segment'...` causes this to read the docs?

arctic/store/_version_store_utils.py
```diff
@@ -76,6 +80,219 @@ def _define_compat_pickle_load():
...
+    for seg in l._collection.find({'symbol': sym}, sort=[('_id', pymongo.ASCENDING)]):
+        logging.info("{: <32}  {: <7}  {: <10} {: <30}".format(
+            hashlib.md5(seg['sha']).hexdigest(),
+            seg.get('segment'),
+            'compressed' if seg.get('compressed', False) else 'raw',
+            [str(p) for p in seg.get('parent', [])]
+        ))
+
+
+def _fast_check_corruption(collection, sym, v, check_count=True, check_last_segment=True, check_append_safe=False):
```
jamesblackburn: `check_count` and `check_last_segment` always True - remove?

arctic/store/_version_store_utils.py
```diff
@@ -76,6 +80,219 @@ def _define_compat_pickle_load():
...
+        #     curs.count()
+        #     curs.next()
+        # (b) collection.find(spec, {'segment': 1}).count()
+        #     collection.find_one(spec, {'segment': 1}, sort=[('segment', pymongo.DESCENDING)])
+
+        if check_count:
+            total_segments = collection.find(spec, {'segment': 1}).count()
+            # Quick check: compare segment count
+            if total_segments != v.get('segment_count', 0):
+                return True  # corrupted, don't proceed with fetching from mongo the first hit
```
jamesblackburn: This is an unlikely case so maybe this short-circuit isn't necessary?

arctic/store/_version_store_utils.py
```diff
@@ -76,6 +80,219 @@ def _define_compat_pickle_load():
...
+                         "check_count and check_last_segment set to True")
+
+    # If version marked symbol as deleted, it will force writes/appends to start from a new base: non corrupted.
+    if isinstance(v.get('metadata'), dict) and v['metadata'].get('deleted'):
+        return False
+     
+    if check_append_safe:
+        # Check whether appending to the symbol version can potentially corrupt the data (history branch).
+        # Inspect all segments, don't limit to v['up_to']. No newer append segments after v should exist.
+        spec = {'symbol': sym, 'parent': v.get('base_version_id', v['_id'])}
```
jamesblackburn: It's only `write_metadata` that writes `base_version_id` ?

arctic/store/_version_store_utils.py
```diff
@@ -76,6 +80,219 @@ def _define_compat_pickle_load():
...
+        True if the symbol is found corrupted, False otherwise.
+    """
+    input_v = l._versions.find_one({'symbol': sym, 'version': input_v}) if isinstance(input_v, int) else input_v
+    return _fast_check_corruption(l._collection, sym, input_v)
+
+
+def is_corrupted(l, sym, input_v):
+    """
+        This method can be used to check for a corrupted version.
+        Will resort to a regular read (slower) if the internally invoked fast-detection gives hint for a corruption.
```
jamesblackburn: Looks like it does the read if `_fast_check_corruption` returns False. i.e. it's slow in the non-corrupt case?

dimosped:
@jamesblackburn  @bmoscon  I will get some slack in about an hour and will attach the benchmarks and the explain info, to back up the non degraded performance observations.
Then, I'll merge.

arctic/store/_version_store_utils.py
dimosped: Look similar.
Without the projection:
https://gist.github.com/dimosped/b0eea9ce2b7084b2833adb4d63bacbac

With the projection:
https://gist.github.com/dimosped/70115b548a3daef71e9ceb5ef343187a


Also benchmarked and with projection was significantly faster, probably due to fethching more data:
200x rounds.

With projection Stats (seconds): 
	Min=0.000867128372192
	Max=0.0121870040894
	Mean=0.00173362374306
	Stdev=0.00144508307458

Without projection stats (seconds): 
	Min=0.00388884544373
	Max=0.0164110660553
	Mean=0.00772412419319
	Stdev=0.00237581682623

arctic/store/_version_store_utils.py
dimosped: The API doc is misleading/wrong.

The _fast_check_corruption returns true if found corruption, and the method will fall back to full read only if the fast check suggests that there is no corruption:
```python
if not _fast_check_corruption(l._collection, sym, input_v):
```
i.e. we want to fail fast (w/o read) if we find corruption with the fast check, and perform the read only if the data look otherwise good form a count/max-segment perspective.

I will update the doc.

arctic/store/_version_store_utils.py
dimosped: Agreed, I've added this check under a flag to use it on-demand:

```python
if check_count:
......
```

arctic/store/_version_store_utils.py
dimosped: In the ndarray store code it is actually called with F T T:

```python
if _CHECK_CORRUPTION_ON_APPEND and _fast_check_corruption(
                    collection, symbol, previous_version,
                    check_count=False, check_last_segment=True, check_append_safe=True):
```

I guess you suggest to remove altogether the default args, ie. force the caller to be explicit, correct?

PR #586, opened 2018-07-18
POC of Version Store with pluggable backend

pablojim:
First implementation of Version Store with pluggable backends. POC in S3 with read, write, soft delete & snapshots using the existing VersionStore chunking and serialisation mechanisms. Append and hard deletes are not implemented.

This implementation stands alone and has no effect on existing functionality. Contains lots of code duplication form the existing functionality. Has limited error checking and cleanup functionality. This PR is mostly for discussion purposes at this point.

### DIFF
--- a/arctic/pluggable/generic_version_store.py
+++ b/arctic/pluggable/generic_version_store.py
@@ -1,0 +1,436 @@
+import logging
+
+import bson
+import six
+
+from .._util import indent
+from ..exceptions import NoDataFoundException, ArcticException
+from arctic.pluggable._pickle_store import PickleStore
+from arctic.store.versioned_item import VersionedItem
+
+logger = logging.getLogger(__name__)
+
+VERSION_STORE_TYPE = 'VersionStore'
+_TYPE_HANDLERS = []
+
+
+def register_versioned_storage(storageClass, storage_args=tuple(), storage_kwargs=None):
+    storage_kwargs = storage_kwargs or {}
+    existing_instances = [i for i, v in enumerate(_TYPE_HANDLERS) if str(v.__class__) == str(storageClass)]
+    store = storageClass(*storage_args, **storage_kwargs)
+    if existing_instances:
+        for i in existing_instances:
+            _TYPE_HANDLERS[i] = store
+    else:
+        _TYPE_HANDLERS.insert(0, store)
+    return storageClass
+
+
+class GenericVersionStore(object):
+
+    _bson_handler = PickleStore()
+
+    def __init__(self, library_name, backing_store):
+        self.library_name = library_name
+        self._backing_store = backing_store
+
+    def __str__(self):
+        return """<%s at %s>
+%s""" % (self.__class__.__name__, hex(id(self)), indent(str(self.library_name), 4))
+
+    def __repr__(self):
+        return str(self)
+
+    def list_symbols(self, all_symbols=False, snapshot=None, regex=None, **kwargs):
+        """
+        Return the symbols in this library.
+
+        Parameters
+        ----------
+        all_symbols : `bool`
+            If True returns all symbols under all snapshots, even if the symbol has been deleted
+            in the current version (i.e. it exists under a snapshot... Default: False
+        snapshot : `str`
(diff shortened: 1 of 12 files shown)


arctic/pluggable/_parquet_store.py
```diff
@@ -0,0 +1,52 @@
...
+
+    def get_info(self, version):
+        ret = {'type': 'parquet', 'handler': self.__class__.__name__}
+        return ret
+
+    def read(self, backing_store, library_name, version, symbol, **kwargs):
+        segment_keys = version['segment_keys']
+        assert len(segment_keys) == 1, "should only be one segment for parquet"
+        # TODO this is S3 functionality bleeding out of the backing store.
+        # Currently reading a Pandas dataframe from a parquet bytes array fails as it only takes a file path.
```
pablojim: Fixed in https://github.com/dask/fastparquet/pull/330/files

arctic/pluggable/key_value_datastore.py
```diff
@@ -0,0 +1,401 @@
...
+        base_symbols_path = self._make_base_symbols_path(library_name)
+        _, dirs, _ = next(os.walk(base_symbols_path), ([], [], []))
+
+        def mtime(d):
+            return os.path.getmtime(os.path.join(base_symbols_path, d))
+
+        return sorted(dirs, key=mtime)
+
+    def delete_symbol(self, library_name, symbol):
+        """Soft deletes a symbol - no data is removed, snapshots still work.
```
jamesblackburn: Snapshots can be done with hard-links which would make deletions safe

jamesblackburn:
Quite a nice bit of work - any comments on performance of the implementations?

pablojim:
@jamesblackburn 
From some early results for the parquet store - for reading some large objects there are dramatic improvements - 3 seconds vs 90 seconds. These are probably worst case scenarios for arctic. Write performance is not so dramatically affected. I need to test more though.

There would also be large improvements due to being able to load partial frames e.g. only loading selected columns and row groups. This may help cases such as #609.

Still some work and implementation decisions to do though.

arctic/pluggable/generic_version_store.py
```diff
@@ -0,0 +1,436 @@
...
+        VersionedItem named tuple containing the metadata and version number
+        of the written symbol in the store.
+        """
+        _id = bson.ObjectId()
+        version = {'_id': _id, 'symbol': symbol, 'metadata': metadata}
+
+        previous_version = self._backing_store.read_version(self.library_name, symbol)
+
+        handler = self._write_handler(version, symbol, data, **kwargs)
+        handler.write(self._backing_store, self.library_name, version, symbol, data, previous_version, **kwargs)
```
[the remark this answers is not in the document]
pablojim: I thought about this approach. One worry is that it becomes very complicated with an explosion of possible interactions between the different parts which may or may not make sense e.g. mongodb metadata with parquet serialisation with 2Mb chunking on a S3 backend etc. etc.

Each library would then have to be configured with a particular config to work correctly. I think it might be an abstraction too far.

yschimke:
@pablojim Shame to let this bitrot, can we discuss later this week with @willdealtry, @shashank88

shashank88:
> @pablojim Shame to let this bitrot, can we discuss later this week with @willdealtry, @shashank88

Yeah, this seems pretty good, will go through it tonight.

yschimke:
If we don’t think this is prod ready or a feature we want to support long term. Maybe we can segregate it is an example and make sure our API allows this sort of flexibility. 

pablojim:
> If we don’t think this is prod ready or a feature we want to support long term. Maybe we can segregate it is an example and make sure our API allows this sort of flexibility.

Apart from some dependencies it is completely isolated from the rest of Arctic. It's all in the "pluggable" package and duplicates some code from the main APIs.

One option would be to merge it but mark it in the code and documentation as Beta until it is deemed ready for wider use.

shashank88:
Will move to a contrib directory to unblock this PR, without committing this to be used as a production store.

PR #589, opened 2018-07-24
Fix issue whereby appending can silently reorder columns

mehertz:
The latest numpy's astype will happily reorder your dtype column names without reordering the columns themselves. Previously we relied on astype to also reorder the columns.

We already have some code that effectively backports the old implementation so this just extends that.


mehertz:
Tests actually do pass, but coverage seems to fail:

INFO:coveralls:{"message":"Couldn't find a repository matching this job.","error":true}

dimosped:
@mehertz  don't worry about the circleci since all tests are green both on circle and travis, looks good, will merge it.

@jamesblackburn @bmoscon  can you give Matt rights for merging on master ?

Issue #596, opened 2018-08-03
Deprecation in pymongo breaking travis


bmoscon:
```
arctic/arctic.py:215: in initialize_library
    if len(self._conn[l.database_name].collection_names()) > 5000 and check_library_count:
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
self = Database(MongoClient(host=['127.15.206.93:27641'], document_class=dict, tz_aware=False, connect=True), u'arctic_test')
include_system_collections = True, session = None
    def collection_names(self, include_system_collections=True,
                         session=None):
        """**DEPRECATED**: Get a list of all the collection names in this
            database.
    
            :Parameters:
              - `include_system_collections` (optional): if ``False`` list
                will not include system collections (e.g ``system.indexes``)
              - `session` (optional): a
                :class:`~pymongo.client_session.ClientSession`.
    
            .. versionchanged:: 3.7
               Deprecated. Use :meth:`list_collection_names` instead.
    
            .. versionchanged:: 3.6
               Added ``session`` parameter.
            """
        warnings.warn("collection_names is deprecated. Use "
                      "list_collection_names instead.",
>                     DeprecationWarning, stacklevel=2)
E       DeprecationWarning: collection_names is deprecated. Use list_collection_names instead.
```

from build:

https://travis-ci.org/manahl/arctic/jobs/411724126

dimosped:
@bmoscon  I was working on fixing this :-)  
am I duplicating any of your WiP ?

bmoscon:
@dimosped nope, go for it!

dimosped:
Thanks.
Will have to wait the PR branch to get green:
https://github.com/manahl/arctic/pull/597

PR #597, opened 2018-08-03
fix failing integration tests on travis due to list_collection_names deprecation

dimosped:
Fixes issues #596 and #599

list_collection_names was added with pymongo 3.6.0:
https://github.com/mongodb/mongo-python-driver/blob/3.6.0/pymongo/database.py#L590

We switched to requiring pymongo>=3.6.0 recently:
https://github.com/manahl/arctic/blob/master/setup.py#L86

Therefore it is safe to simply replace collection_names with list_collection_names

### DIFF
--- a/README.md
+++ b/README.md
@@ -111,7 +111,7 @@ Arctic storage implementations are **pluggable**.  VersionStore is the default.
 Arctic currently works with:
 
  * Python 2.7, 3.4, 3.5, 3.6
- * pymongo >= 3.0
+ * pymongo >= 3.6
  * Pandas
  * MongoDB >= 2.4.x
 
@@ -137,6 +137,7 @@ It wouldn't be possible without the work of the AHL Data Engineering Team includ
  * [Wilfred Hughes](https://github.com/wilfred)
  * [Edward Easton](https://github.com/eeaston)
  * [Bryant Moscon](https://github.com/bmoscon)
+ * [Dimosthenis Pediaditakis](https://github.com/dimosped)
  * ... and many others ...
 
 Contributions welcome!
--- a/arctic/_util.py
+++ b/arctic/_util.py
@@ -1,9 +1,21 @@
 from pandas import DataFrame
 from pandas.util.testing import assert_frame_equal
-from pymongo.errors import OperationFailure
 import logging
+import pymongo
+
 
 logger = logging.getLogger(__name__)
+
+# Avoid import-time extra logic
+_use_new_count_api = None
+
+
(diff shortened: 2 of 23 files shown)


dimosped:
@bmoscon and @jamesblackburn  I added my self in the list of the contributors if you don't mind.

dimosped:
@bmoscon  oh no!
More deprecation errors, fixing this as well
```
DeprecationWarning: count is deprecated. Use Collection.count_documents instead.
```

dimosped:
instead of calling count() on the cursor, we now have to wall count_documents() on the collection, as a separate query.
http://api.mongodb.com/python/current/api/pymongo/cursor.html#pymongo.cursor.Cursor.count

The count_documents() was introduced in pymongo 3.7 and count() on the cursor was deprecated also at the same time.
http://api.mongodb.com/python/current/api/pymongo/collection.html#pymongo.collection.Collection.count_documents

The options were:
1. Change our codebase to use count_documents(), forcing also setup.py to pymongo>=3.7 Unfortunately this would require a lot of code changes at the moment, as count() is all over the place in our home-brew projects.
2. Change our travis now to 3.6.0 <=pymongo<= 3.6.1 (code will still be compatible with 3.7 but with warning messages). No testing for 3.7, no detected breakages which is not great.
3. Change our setup.py to 3.6.0 <=pymongo<= 3.6.1, which is quite sad.

**Chosen solution**:
For now I have added a compatibility mode which makes things work with all pymongo versions, avoiding also the warning messages.

Later when 3.7+ will become the norm, we can drop support for <=3.6.1 and switch to using exclusively count_documents().

dimosped:
@bmoscon  @jamesblackburn  please review the last commit and see my comment above, as this is blocking us.

bmoscon:
looks good @dimosped

dimosped:
This now fixes also #599, had to slightly modify a few of our fixtures.

Also changed the fix for #597 using a solution which avoids any import-time logic, and on first use of count() will lazily detect the right API.

dimosped:
@bmoscon  feel free to have another look as I had to change quite a few things to make it work for:
 - python 3.
 - pytest 3.7.3
 - pymongo 3.7++
 - maintain backwards cocmpatiblility

I won't merge until we have both python 2+3 green on Travis and Green Circle CI.

bmoscon:
@dimosped looks good, and travis and circle are both green now! (circle only failed on coveralls, since you're on a fork not a branch)

dimosped:
thanks @bmoscon  merging now

PR #606, opened 2018-08-16
Version store documentation

bmoscon:
Create initial first pass of version store documentation. I've tried to cover the most important use cases for the main methods of version store, focusing on the basic usecase (dataframes). Is surely not complete, but I think its much better than what we currently have (nothing) :)

I will squash the commits into a single commit

### DIFF
--- a/docs/versionstore.md
+++ b/docs/versionstore.md
@@ -112,7 +112,7 @@ date       id  data
 
 ```
 
-
+DateRange's only apply to pandas DataFrames, and the dataframe must have a datetime index present. 
 
 
 
--- a/docs/versionstore.md
+++ b/docs/versionstore.md
@@ -140,8 +140,65 @@ VersionedItem(symbol=new,library=arctic.vstore,data=<class 'NoneType'>,version=1
 
 ```
 
+Other Methods
+
+A number of other utility methods are available:
+
+* delete
+* has_symbol
+* list_versions
+* read_metadata
+* write_metadata
+* restore_version
+* snapshot
+* delete_snapshot
+* list_snapshots
+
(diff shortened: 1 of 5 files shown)


bmoscon:
@dimosped @jamesblackburn @TomTaylorLondon please have a look. Feel free to add anyone else to the review that may want to look it over

bmoscon:
its probably easiest to read it over here: https://github.com/manahl/arctic/blob/version-store-docs/docs/versionstore.md

docs/versionstore.md
```diff
@@ -1,3 +1,316 @@
...
\ No newline at end of file
+
+(note: current doc is based on arctic-1.68.0)
+
+VersionStore serializes and stores Pandas objects, numpy arrays as well as other python types in MongoDB. Objects are `versioned` and new versions are created when a `symbol` is modified. 
+
+
+# Reading and Writing Data with VersionStore
+
+```
```
dimosped: not sure if specifying the language makes any difference, checking:

python defined:
```python 
import lz4

def print_lz4_version():
    print(lz4.__version__)
```

vs 

default lang:
```
import lz4

def print_lz4_version():
    print(lz4.__version__)
```

dimosped: basically its colors for syntax highlighting

dimosped:
great @bmoscon , the vstore documentation looks quite complete.
If we get readthedocs integration, we can add links to API documentation.

docs/versionstore.md
bmoscon: the dataframe output and the >>> tend to not render well when you use the `python` formatting

PR #608, opened 2018-08-24
Ensure Arctic performs well with MongoDB 3.6+

rob256:
Mongo 3.6+ no longer uses the index to sort when using multi-key indexes ([SERVER-19402](https://jira.mongodb.org/browse/SERVER-19402)), therefore, we are now using the client to sort.

Using in-memory sorting in MongoDB 3.6+ risks hitting the 32MB memory limit, causing queries to fail.

### DIFF
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -1,6 +1,7 @@
 import hashlib
 import logging
 import os
+from operator import itemgetter
 
 
 from bson.binary import Binary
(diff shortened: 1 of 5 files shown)


jamesblackburn:
From SERVER-19402 it seems this only affects sorting on indexed array fields. As we're sorting on a scalar field (which is part of the multi-key index) we would expect sort to still work. However it seems this requires the index to be rebuilt.

dimosped:
@bmoscon  @jamesblackburn  @rob256  I would like to hold this back and make first 1.68.1 release if you don't mind

bmoscon:
@dimosped go for it!

jamesblackburn:
Yep I'm happy.

Issue #612, opened 2018-08-26
Issue with _compression.py


bmoscon:
The recent changes to `_compression.py` included the addition of a global threadpool. This means that as soon as you import arctic in a program, you can no longer spawn any new child processes. While an arctic object / connection shouldnt be used across forks, the import should not need to be done in each process, you should be able to import once and then create local arctic objects without issues.

```
from arctic._compression import compress_array
from multiprocessing import Process


def main():
    data = []
    for i in range(23):
        with open(str(i) + ".tmp", 'rb') as fp:
            d = fp.read()
            data.append(d)
    print(len(data))
    print(len(data[0]))
    new = compress_array(data)


if __name__ == '__main__':
    p = Process(target=main)
    p.start()
```

The above code will fail every time, and by fail I mean it will just block forever waiting for the threadpool map to complete. (I waited over 4 hours). If the same code is run without the child process being created, it finishes in less than a second. (note: the file loading is just a way to create large enough data sets to trigger the parallel compression. Any large enough data should work).

@dimosped

bmoscon:
fixed

PR #613, opened 2018-08-26
resolve issue 612

bmoscon:
Rather than create a ThreadPool as soon arctic is imported, create it when its first used by Arctic

see issue #612

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,7 @@
 
 ### 1.69
   * Docs: VersionStore documentation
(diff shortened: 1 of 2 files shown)


bmoscon:
@dimosped i'm open to other ways to solve this (i.e. integrating the compression component more tightly into the arctic object, etc)

bmoscon:
circleci seems to be having issues, but since its being turned off in 4 days, not sure its a big deal...

Issue #619, opened 2018-09-01
VersionedItem is not backwards compatible between 1.68 and 1.67.1


scriada:
#### Arctic Version

```
1.68
```

#### Arctic Store

```
VersionStore
```

#### Platform and version

All

#### Description of problem and/or code sample that reproduces the issue

#595 introduced a `host` attribute to `VersionedItem`. It turns out that some people using arctic were creating their own `VersionedItem` objects, making this a breaking change. To make this backwards compatible we should make `host` optional.

PR #620, opened 2018-09-01
Make host optional in VersionedItem (#619)

scriada:
This makes `VersionedItem` backwards compatible, so any third-party
libraries creating their own `VersionedItem`s won't be affected.


bmoscon:
this good to merge @scriada ?

PR #622, opened 2018-09-05
Initial implementation of default handling for pickled frames

willdealtry:
For MDP-3767 - throw when a handler has failed to handle a particular type, rather than falling back to default pickling behaviour

### DIFF
--- a/arctic/scripts/arctic_fsck.py
+++ b/arctic/scripts/arctic_fsck.py
@@ -1,4 +1,3 @@
-
 import logging
 import argparse
 
--- a/arctic/scripts/arctic_fsck.py
+++ b/arctic/scripts/arctic_fsck.py
@@ -1,3 +1,4 @@
+
 import logging
 import argparse
 
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -1,5 +1,6 @@
 from datetime import datetime as dt, timedelta
 import logging
+import os
 
 import bson
 from pymongo import ReadPreference
@@ -21,6 +22,7 @@ logger = logging.getLogger(__name__)
 
 VERSION_STORE_TYPE = 'VersionStore'
 _TYPE_HANDLERS = []
+AVOID_FALLBACK_HANDLERS = bool(os.environ.get('AVOID_FALLBACK_HANDLERS'))
 
 
 def register_versioned_storage(storageClass):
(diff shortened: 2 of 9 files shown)


arctic/scripts/arctic_fsck.py
```diff
@@ -1,3 +1,4 @@
+
```
bmoscon: we can remove this unnecessary whitespace

arctic/store/version_store.py
```diff
@@ -21,6 +22,7 @@
 
 VERSION_STORE_TYPE = 'VersionStore'
 _TYPE_HANDLERS = []
+AVOID_FALLBACK_HANDLERS = bool(os.environ.get('AVOID_FALLBACK_HANDLERS'))
```
pablojim: Is it worth making this a property of the library/store rather than a global property?

arctic/store/version_store.py
```diff
@@ -384,10 +395,24 @@ def get_info(self, symbol, as_of=None):
...
+
+        # If the handler doesn't support interrogation of its read options assume
+        # that it does support this option (i.e. fail-open)
+        return True
+
     def _do_read(self, symbol, version, from_version=None, **kwargs):
         if version.get('deleted'):
             raise NoDataFoundException("No data found for %s in library %s" % (symbol, self._arctic_lib.get_name()))
         handler = self._read_handler(version, symbol)
+        if AVOID_FALLBACK_HANDLERS:
```
pablojim: The body of the if seems only vaguely related to AVOID_FALLBACK_HANDLERS. Would a more general option that toggles more strict kwarg checking be a better approach?

arctic/store/version_store.py
```diff
@@ -301,12 +303,21 @@ def _read_handler(self, version, symbol):
...
+        return False
+
     def _write_handler(self, version, symbol, data, **kwargs):
         handler = None
         for h in _TYPE_HANDLERS:
             if h.can_write(version, symbol, data, **kwargs):
                 handler = h
                 break
+            if AVOID_FALLBACK_HANDLERS and self.handler_can_write_type(h, data):
+                raise ArcticException( "Not falling back to default handler for %s" % symbol )
```
pablojim: This behaviour is slightly different than the exception message suggests. If I have multiple handlers for one type this new code will prevent ones further down the list being used as well as preventing the default handler being used.

dimosped:
@bmoscon  @willdealtry  @pablojim  @jamesblackburn  if no objections, I'd like to touch/finish-up this PR and merge it , to be included in our next release.

pablojim:
LGTM

bmoscon:
looks good @dimosped

dimosped:
@bmoscon  @jamesblackburn  @rob256  I'd like to leave this for 1.71.0 as we already have lots of changes.
Until then, we'll have also the response from Mongo support.
Nonetheless, it is a sane change, and will offload Mongo (assuming it doens't make ops too slow at client for very large data -- I doubt it is noticeable -- ).

PR #634, opened 2018-10-04
Fixed small bug in DateRange.intersection

davidduenas3:
'DateRange.intersection' method returns CLOSED interval when intersecting ranges with same start[,end] and distinct interval values. 
It should return an  OPEN_*(>start) interval as it is subset of the same CLOSED_*(>=start)

### DIFF
--- a/arctic/date/_daterange.py
+++ b/arctic/date/_daterange.py
@@ -85,12 +85,12 @@ class DateRange(GeneralSlice):
                     else self.startopen if other.start is None \
                     else other.startopen if self.start < other.start \
                     else self.startopen if self.start > other.start \
-                    else (self.startopen and other.startopen)
+                    else (self.startopen or other.startopen)
         endopen = other.endopen if self.end is None \
                    else self.endopen if other.end is None \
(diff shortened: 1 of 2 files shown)


bmoscon:
@davidduenas3 thanks for this! Can you add some new unit tests that pertain to this?

bmoscon:
@davidduenas3 thanks! Can you please add some unit tests that pertain to the fix?

davidduenas3:
`oo=DateRange('20110101', '20110102', OPEN_OPEN)
cc=DateRange('20110101', '20110102', CLOSED_CLOSED)
oo.start in oo == False
oo.start in cc == True
oo.start in cc.intersection(oo) == False`
Returns: True, True, True

bmoscon:
@dimosped @jamesblackburn looks good to me, any reason we shouldnt merge this?

dimosped:
@bmoscon  looks good :+1:  thanks

dimosped:
@bmoscon  if no objections, I will add a changelog update and merge and add it in our next (soon to be done) release

Issue #590, opened 2018-07-29
PandasStore implementations can serialise unnecessarily twice certain types of dataframes


dimosped:
#### Arctic Version

```
1.57.1
```

#### Arctic Store

```
VersionStore
```

#### Description of problem and/or code sample that reproduces the issue

The code which decides whether a given write handler can be used, does full-serialization to recarray of a dataframe whenever one or more object columns are detected.
https://github.com/manahl/arctic/blob/master/arctic/store/_pandas_ndarray_store.py#L172
https://github.com/manahl/arctic/blob/master/arctic/serialization/numpy_records.py#L140

This can happen in the following cases:
- Multi-indexed dataframe
- Multi-column dataframe
- any dataframe which container one or more columns of type object (includes string, or any column which might have even one row with a string/None value)

In the above cases, we end up serializing twice the dataframe, once for testing the write handler, and second to actually serialize when we call write/append.

For large data-frames, this slows down writes/appends and explodes further the memory usage.

dimosped:
@bmoscon  @jamesblackburn @aflag  here is the draft PR for this fix.

It is being merged on my own arctic clone (onto the branch for the incremental serialization).
https://github.com/dimosped/arctic/tree/incremental-serialization-initial

After merging it will become automatically an extra change in this PR:
https://github.com/manahl/arctic/pull/578

I've made this optimisation optional (disabled by default) and added as thorough as possible tests.

dimosped:
@aflag  this is interesting, we don't even manage to hit this check for n-dimensional DFs, as the serializer explodes:
https://github.com/manahl/arctic/blob/master/arctic/serialization/numpy_records.py#L151

```python
In[50]: import pandas as pd
   ...: import numpy as np
   ...: from arctic.serialization.numpy_records import DataFrameSerializer
   ...: 
   ...: dims = (2, 2, 2)
   ...: pdf = pd.DataFrame({'a': [np.zeros(dims), np.ones(dims)], 'b': [np.zeros(dims), np.ones(dims)]})
   ...: # recs = pdf.to_records()
   ...: f_serializer = DataFrameSerializer()
   ...: recs = df_serializer.serialize(pdf)
Traceback (most recent call last):
  File ".egg/IPython/core/interactiveshell.py", line 2881, in run_code
    exec(code_obj, self.user_global_ns, self.user_ns)
  File "<ipython-input-50-6833e8b9137b>", line 9, in <module>
    recs = df_serializer.serialize(pdf)
[...]

PR #643, opened 2018-10-26
Complete implementation and tests for the fast is-serializable check

dimosped:
Issue #590

I extracted only the fast check improvement form the incremental serialization branch.

This has already been reviewed/approved:
https://github.com/dimosped/arctic/pull/1

It is turned off by default, and I'd like to include it in the next release

### DIFF
--- a/benchmarks/fast_serializable_check.py
+++ b/benchmarks/fast_serializable_check.py
@@ -1,0 +1,46 @@
+import time
+
+import arctic.serialization.numpy_records as anr
+from tests.unit.serialization.serialization_test_data import _mixed_test_data as input_test_data
+
+df_serializer = anr.DataFrameSerializer()
+
+
+def _bench(rounds, input_df, fast):
+    fast = bool(fast)
+    anr.set_fast_check_df_serializable(fast)
+    start = time.time()
+    for i in range(rounds):
+        df_serializer.can_convert_to_records_without_objects(input_df, 'symA')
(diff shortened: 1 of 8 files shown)


benchmarks/fast_serializable_check.py
```diff
@@ -0,0 +1,46 @@
...
+df_serializer = anr.DataFrameSerializer()
+
+
+def _bench(rounds, input_df, fast):
+    fast = bool(fast)
+    anr.set_fast_check_df_serializable(fast)
+    start = time.time()
+    for i in range(rounds):
+        df_serializer.can_convert_to_records_without_objects(input_df, 'symA')
+    print("Time per iteration (fast={}): {}".format(fast, (time.time() - start)/rounds))
```
yschimke: Curious: does python suffer any of the same issues as java for microbenchmarking?

dimosped: JVM definitely has many more optimisations in place, which are affected by micro-benchmarking, and make it tough to deduce safe results for the real-world performance, of the whole app.

Python in this sense is not suffering from the same issues, but still, micro-benchmarking is used more for runtime trends of small code fragments, and shouldn't be used in a generalized manner for the whole app performance.

dimosped:
@bmoscon I remember you expressed interest in this change,  bear in mind that the flag for this optimisation is off, and you will have to set env variable "ENABLE_FAST_CHECK_DF_SERIALIZABLE" or use "set_fast_check_df_serializable(True)" to turn it on.

PR #649, opened 2018-11-06
Setting default for create_index()'s background to True

jonbannister:
This is to prevent accidental foreground index creation, which can block database operations if it takes a particularly long time.


dimosped:
@jonbannister can you also update the call assertions which fail?

shashank88:
Closing this since index creation is not done on store creation anyways

shashank88:
We will land it with the fixed Unit test

shashank88:
Closing this, will create a new PR later with the unit test fix.

Issue #650, opened 2018-11-07
Implement forward pointers for chunks in VersionStore


dimosped:
#### Arctic Version

```
1.72.0
```

#### Arctic Store

```
VersionStore
```

#### Description of problem and/or code sample that reproduces the issue

Add implementation for supporting forward pointers from version documents to segments, instead of the current reverse pointers.

The benefit of this is that we minimize the number of updated/touched mongoDB documents upon a write/append, only updating the list of ObjectIDs in the version document.

Implement 3 globally configured options:

- FW pointers Disabled: VersionStore operates as it used to
- FW pointers Enabled: VersionStore only updates the version document with the fw pointers to segments.
- FW pointers Hybrid: Compatibility operation. VersionStore updates both the version document with fw-pointers, and the segments with backwards pointersr to parents/versions. Makes it easy to experiment safely with FW pointers, and switch back to original implementation (disabled) without issues.

Issue #547, opened 2018-04-20
Investigate the potential read performance improvement using forward pointers


dimosped:
Currently the read path of ndarray_store, instead of fetching the segments using a query spec that matches against 'parent' (segment document stores parent as an array of ObjectIDs):
https://github.com/manahl/arctic/blob/master/arctic/store/_ndarray_store.py#L243

Investigate the potential performance improvements:
- storing in the version document a list of segment object IDs and query with {'$in': [objID1, objID2....]}
- storing in the version document a list of segment SHAs and query with {'$in': [sha1, sha2....]}

Investigate using a MongoDB 3.4 or better 3.6 as there have been improvements for the queries using $in:  https://jira.mongodb.org/browse/SERVER-30189

When the prototype solution is developed, benchmark and pay attention to the indexes used, as well as understand the query execution via studying the explain() results.

Test against a variety of scenarios: variable document size, variable number of versions and symbols, variable number of shards, consistency of data.

The solution needs to be backwards compatible.

dimosped:
@jamesblackburn  @yschimke  choosing symbol+sha for the query vs symbol+id doesn't make a difference in terms of query execution time, explain gives:
```
In[47]: collection.find({'_id': id, 'symbol': sym}).explain()['queryPlanner']['winningPlan']['shards'][0]['winningPlan']
Out[47]: 
{u'inputStage': {u'filter': {u'symbol': {u'$eq': u'BBG000ZB5GX9'}},
  u'inputStage': {u'direction': u'forward',
   u'indexBounds': {u'_id': [u"[ObjectId('55971d7f2914cf59a3031ebc'), ObjectId('55971d7f2914cf59a3031ebc')]"]},
   u'indexName': u'_id_',
   u'indexVersion': 1,
   u'isMultiKey': False,
   u'isPartial': False,
   u'isSparse': False,
   u'isUnique': True,
   u'keyPattern': {u'_id': 1},
   u'stage': u'IXSCAN'},
  u'stage': u'FETCH'},
 u'stage': u'SHARDING_FILTER'}
In[48]: collection.find({'sha': sha, 'symbol': sym}).explain()['queryPlanner']['winningPlan']['shards'][0]['winningPlan']
Out[48]: 
{u'inputStage': {u'inputStage': {u'direction': u'forward',
[...]

PR #651, opened 2018-11-07
Forward pointers implementation

dimosped:
Solves #650, #547 and #663

Implement 3 globally configured env variables:

- **ARCTIC_FORWARD_POINTERS=Disabled**: VersionStore operates as it used to
- **ARCTIC_FORWARD_POINTERS=Enabled**: VersionStore only updates the version document with the fw pointers to segments.
- **ARCTIC_FORWARD_POINTERS=Hybrid**: Compatibility operation. VersionStore updates both the version document with fw-pointers, and the segments with backwards pointersr to parents/versions. Makes it easy to experiment safely with FW pointers, and switch back to original implementation (disabled) without issues. In this mode, reads prefer the FW pointers (if exist) otherwise falls-back to original backwards pointers based reads.


arctic/store/_ndarray_store.py
```diff
@@ -120,6 +129,21 @@ def set_corruption_check_on_append(enable):
...
 
+def update_fw_pointers(collection, result, symbol, version, previous_version):
+    if result is None:
+        return
+    if previous_version and FW_POINTERS_KEY not in previous_version:
+        chunk_ids = [_id for _id in collection.find({'symbol': symbol, 'parent': version_base_or_id(version)},
+                                                    {'_id': 1})]
+    else:
+        chunk_ids = previous_version[FW_POINTERS_KEY] if previous_version else []
+    upserted_ids = [result.upserted_id] \
```
yschimke: why use the ternary operator and escape newlines instead of a bog standard if?

yschimke: why even assign to upserted_ids?

arctic/store/_ndarray_store.py
```diff
@@ -377,11 +408,20 @@ def _do_append(self, collection, version, symbol, item, previous_version, dirty_
                 segment = {'data': Binary(data), 'compressed': False}
                 segment['segment'] = version['up_to'] - 1
                 try:
-                    collection.update_one({'symbol': symbol,
-                                           'sha': checksum(symbol, segment)},
-                                          {'$set': segment,
-                                           '$addToSet': {'parent': version['base_version_id']}},
-                                          upsert=True)
+                    if ARCTIC_FORWARD_POINTERS is FwPointersCfg.DISABLED:
```
yschimke: Is this block clearer if written as two optional write operations?, the second collection.update_one is rewritten but essentially the same

arctic/store/_ndarray_store.py
```diff
@@ -462,12 +502,17 @@ def check_written(self, collection, symbol, version):
...
+        else:
+            spec = {'symbol': symbol, 'parent': parent_id}
+
+        seen_chunks = mongo_count(collection, filter=spec)
 
         if seen_chunks != version['segment_count']:
-            segments = [x['segment'] for x in collection.find({'symbol': symbol, 'parent': parent_id},
-                                                              projection={'segment': 1},
-                                                              )]
+            segments = [x['segment'] for x in collection.find(spec, projection={'segment': 1})]
```
yschimke: worth stripping _id?

arctic/store/_ndarray_store.py
```diff
@@ -268,14 +292,21 @@ def _do_read(self, collection, version, symbol, index_range=None):
...
-                'parent': version_base_or_id(version),
                 'segment': {'$lt': to_index}
                 }
         if from_index is not None:
             spec['segment']['$gte'] = from_index
         else:
             segment_count = version.get('segment_count', None)
 
+        # We want to use FW pointers to read data if:
+        #  ARCTIC_FORWARD_POINTERS is currently enabled/hybrid AND last version was written with FW pointers,
```
yschimke: Should we consider making enabled use the reverse pointers to read, but then checking that the forward pointers reconcile?  As a way of validating for now?

Just an idea, I'm not against this approach, just asking what we want to test during the hybrid mode period.

yschimke:
Looks great to me, but I don't consider myself qualified to approve

arctic/store/_ndarray_store.py
dimosped: I decided to favor some code duplication in favor of minimizing the regression risks, and keep the existing code path (with disabled fw pointers) as similar as possible.

We can properly refactor this once we build confidence in our implementation.
I am adding a todo with your suggestion though.

arctic/store/_ndarray_store.py
dimosped: The written data are being verified upon write, i.e. we count the segments matching the write spec which have been actually written.
https://github.com/dimosped/arctic/blob/forward_pointers_final/arctic/store/_ndarray_store.py#L505

Tbh I am not against this idea, but maybe as a separate option/flag and add it as part of the verification check:
https://github.com/dimosped/arctic/blob/forward_pointers_final/arctic/store/_ndarray_store.py#L505

It would also mean r/w will be slower with an extra step of  1-query collecting N-document IDs.

arctic/store/_ndarray_store.py
dimosped: should be more readable now, tks

arctic/store/_ndarray_store.py
dimosped: handy debug info indeed, added now

bmoscon:
I dont have an issue with this PR - this is more a general critique - rather than using env variables (we have quite a few of them now, all defined in different files) can we also support a YAML file? the default would be env vars, but if any of the values were also defined in the yaml file, those would take precedence?

dimosped:
I was also thinking to refactor and create a yaml or "arcitc/config.py" module which holds all configurations.
The rest of the modules will simply import variables from this, and user will have all config in a common file.

arctic/store/_ndarray_store.py
dimosped: I actually went ahead and implemented this, but during testing of forcing a failed check, a nasty bug was uncovered.

If mongo_retry operation fails once, the successfully written segments always remain (it is Arctic's design decision which won't change now). The upserted ids though in subsequent retried, don't include them, breaking the model. 

Working on it and will add explicit tests.

dimosped:
maybe we could even have instead in config.py a python dict with default values.

I always prefer plain python dicts over yaml when possible, and given all configs are exposed via env variables, they don't need code re-release with updated values.

@yschimke  @bmoscon  @jamesblackburn  @willdealtry  any objections with the above?

yschimke:
@dimosped no objections, it might be best done as a separate PR and move everything over for consistency.

dimosped:
@yschimke  agreed, I will add even more config options in the next PR for the internal async work, so it would be best to refactor then, with the next PR.

dimosped:
@yschimke  @bmoscon @jamesblackburn this was a nasty one over with partial segment writes between mongo_retries:
https://github.com/dimosped/arctic/commit/f2bb75248f7fe0039e4370aa88e117c7e6e40636

Need to also correct/update the versoins/segments garbage collection, to update it to also work for FW robustly

arctic/_util.py
```diff
@@ -15,6 +17,24 @@
...
 
+class FwPointersCfg(Enum):
+    ENABLED = 0
+    DISABLED = 1
+    HYBRID = 2
+
+
+# Forward pointers configuration
+FW_POINTERS_KEY = 'CHUNK_IDS'
+ARCTIC_FORWARD_POINTERS_RECONCILE = bool(os.environ.get('ARCTIC_FORWARD_POINTERS_RECONCILE'))
```
jamesblackburn: Worth documenting these environment variables. Not sure what `...RECONCILE` does without reading the code

arctic/store/_ndarray_store.py
```diff
@@ -120,6 +121,28 @@ def set_corruption_check_on_append(enable):
     _CHECK_CORRUPTION_ON_APPEND = bool(enable)
 
 
+def update_fw_pointers(collection, symbol, version, previous_version, ids_of_updated_segments=None, result=None):
+    # If this is the first use of fw-pointers, query for the full list of segment IDs,
+    # or continue by extending the IDs from the previous version
+    if previous_version and FW_POINTERS_KEY not in previous_version:
+        segment_ids = {_id for _id in collection.find({'symbol': symbol, 'parent': version_base_or_id(version)},
```
jamesblackburn: Worth asserting here the segment id count is right (as is done in the read path)?

arctic/store/_ndarray_store.py
```diff
@@ -268,14 +293,21 @@ def _do_read(self, collection, version, symbol, index_range=None):
...
                 }
         if from_index is not None:
             spec['segment']['$gte'] = from_index
         else:
             segment_count = version.get('segment_count', None)
 
+        # We want to use FW pointers to read data if:
+        #  ARCTIC_FORWARD_POINTERS is currently enabled/hybrid AND last version was written with FW pointers,
+        #  otherwise fall back to original implementation
+        if ARCTIC_FORWARD_POINTERS is not FwPointersCfg.DISABLED and FW_POINTERS_KEY in version:
```
jamesblackburn: I think this `if` should always read via fwd pointers if the field is there, as:
- intent is to move to this anyway
- if someone doesn’t set the environment variable, then reads to fwd pointer only libs will fail. 

Reads should be forwards compatible in new versions of arctic. Users shouldn’t need to set variables to read data.

arctic/store/_ndarray_store.py
```diff
@@ -510,13 +572,21 @@ def _do_write(self, collection, version, symbol, item, previous_version, segment
...
+        if ARCTIC_FORWARD_POINTERS is FwPointersCfg.DISABLED:
+            if previous_version:
+                previous_shas = set(Binary(x['sha']) for x in
+                                    collection.find({'symbol': symbol}, projection={'sha': 1, '_id': 0}))
+        else:
+            # For forward pointers we always need to run the following query (not only when previous version exists).
+            # Reason is that if only some of the segments might get written, and we re-execute write() via mongo_retry,
+            # the upserted_id(s) won't be a complete list of all segment ObjectIDs
+            shas_to_ids = {x['sha']: x['_id']
+                           for x in collection.find({'symbol': symbol}, projection={'sha': 1, '_id': 1})}
```
jamesblackburn: This won’t be covered by an index, will it? Also looks quite expensive as you’re iterating over every chunk for a symbol - and mongo will thrash the WT cache as it loads each into memory...

Should we be using shas instead?

arctic/store/_ndarray_store.py
```diff
@@ -510,13 +572,21 @@ def _do_write(self, collection, version, symbol, item, previous_version, segment
...
+        shas_to_ids, matched_segment_ids = set(), set()
+
         previous_shas = []
-        if previous_version:
-            previous_shas = set([Binary(x['sha']) for x in
-                                 collection.find({'symbol': symbol},
-                                                 projection={'sha': 1, '_id': 0},
-                                                 )
-                                 ])
+        if ARCTIC_FORWARD_POINTERS is FwPointersCfg.DISABLED:
```
jamesblackburn: Again shouldn’t this be based on precious_version, not on the environment variable?

I’m worried that people / code writing to different libraries with different env variables set will leave the versions in a mess?

arctic/store/version_store.py
```diff
@@ -837,33 +837,35 @@ def _find_base_version_ids(self, symbol, version_ids):
                                       '_id': {'$nin': version_ids},
                                       'base_version_id': {'$exists': True},
                                       },
-                                     projection=['base_version_id'],
+                                     projection={'base_version_id': 1},
                                      )
         return [version["base_version_id"] for version in cursor]
 
-    def _prune_previous_versions(self, symbol, keep_mins=120, keep_version=None):
+    def _prune_previous_versions(self, symbol, keep_mins=0.01, keep_version=None):
```
jamesblackburn: Why are you turning keep_mins down? 

This is / was needed to handle replication lags etc. Need to ensure this is larger than the maximum lag, or secondary reads can fail.

tests/integration/store/test_version_store.py
```diff
@@ -1607,3 +1624,74 @@ def test_can_write_tz_aware_data_series(library):
...
+    (FwPointersCfg.HYBRID, FwPointersCfg.HYBRID, FwPointersCfg.HYBRID, FwPointersCfg.HYBRID),
+
+    (FwPointersCfg.HYBRID, FwPointersCfg.DISABLED, FwPointersCfg.HYBRID, FwPointersCfg.DISABLED),
+    (FwPointersCfg.HYBRID, FwPointersCfg.ENABLED, FwPointersCfg.HYBRID, FwPointersCfg.ENABLED),
+
+    (FwPointersCfg.ENABLED, FwPointersCfg.HYBRID, FwPointersCfg.ENABLED, FwPointersCfg.HYBRID),
+    (FwPointersCfg.DISABLED, FwPointersCfg.HYBRID, FwPointersCfg.DISABLED, FwPointersCfg.HYBRID),
+
+    (FwPointersCfg.DISABLED, FwPointersCfg.ENABLED, FwPointersCfg.DISABLED, FwPointersCfg.ENABLED),
+    # This throws exception, and ENABLED --> DISABLED fw-pointers config is not allowed/supported
```
jamesblackburn: This will almost certainly happen in real life though if disabled is the default

dimosped:
@jamesblackburn  @yschimke  my last commit made this change fully forward/backwards compatible between enabled-fw-pointers/legacy-parent-pointer/hybrid.

Tomorrow I will finalize the pruning, with respect to the last change for compatibility.

arctic/store/version_store.py
dimosped: ignore this, I have set it for testing the pruning, when the PR reaches its final form, I will revert

arctic/store/_ndarray_store.py
dimosped: I have already updated in my last commit this, be based whenever makes sense on the previous version configuration.

arctic/store/_ndarray_store.py
dimosped: I will check this tomorrow with an explain() and see what happens because _id is quite special, and we have already an index for sha.

If this proves indeed to be expensive, I will consider the switch to shas.

arctic/_util.py
dimosped: yep, adding comments to make it clear.

In the next PR for the async, will move all env variable configs to a separte arctic/config.py module, because they have now become too many and are spread in many modules.

tests/integration/store/test_version_store.py
dimosped: my last commit solved all compatibility issues for all possible transitions.
I will add even more thorough tests tomorrow, but the tests which were already failing now are green.

arctic/store/_ndarray_store.py
dimosped: plase check the new implementation: _spec_fw_pointers_aware()
we no longer rely on env/global variables.

I had already this change in flight, it seems unfortunate that you took the time to review the outdated code.

arctic/store/version_store.py
dimosped: I will resolve this conversation once I switch it back.

dimosped:
@jamesblackburn  @yschimke forgot to mention that an extra aim was to make this also possible/supported:
  - user tries out FW pointers in hybrid mode (e.g. with arctic v1.73.0), writes symbolX
  - then another application which is still in e.g.  v1.67,1, can read/write symbolX as we are fully backwards compatible
  - arcitc v1.73.0 can re-read/re-write symbolX touched by v.167.1
  - user switches to FW ENABLED (only forward pointers) and now v1.67.0 can of course no longer read/write
  - user switches back to FW HYBRID or FW DISABLED and touches symbolX again. 
  - v1.67.0 can read/write symbolX again

yschimke:
:shipit:

arctic/_util.py
dimosped: @yschimke  @jamesblackburn  @bmoscon  I have now added comments with explanation for variables and the fw pointer helper methods.

I will mark this specific conversation now as resolved.

arctic/store/_ndarray_store.py
dimosped: I am resolving this now, the new implementation if code detects fw-pointers info on the version document, it is just using this, without dependency on the environment vars, except FW are explicitly disabled by user.

arctic/store/_ndarray_store.py
dimosped: Before sharing the explain() results, some benchmarks vs this:
https://github.com/manahl/arctic/blob/master/arctic/store/_ndarray_store.py#L515-L519

image

I assume the extra time is only for the transfer of the ObjectIDs and execution is acceptably fast.
The +6ms IMHO in almost a second, can be considered negligible.

Will share the explain() findings shortly

dimosped: Interesting, @jamesblackburn  look how the projection changes the index used between the first and the last queries.

This raises some suspicion that WT cache might indeed get polluted.

image

dimosped: Ah...
image

dimosped: Just noticed,
```
eqlib._collection.find({'symbol': symbol}, projection={'sha': 1, '_id': 1}).explain()
```
has a FETCH stage, 
while this doesn't 
```
eqlib._collection.find({'symbol': symbol}, projection={'sha': 1, '_id': 0}).explain()
```

It is quite obvious that @jamesblackburn  intuition is correct, we would be polluting the cache, an d it is a good idea to convert to using SHAs instead, or add an index (symbol, sha, _id).

I'd rather use SHA though for obvious reasons

yschimke:
LGTM - since this is optional, I suggest landing this and adding the tests as next PR.  It's a big PR already.

yschimke:
🥳

Issue #653, opened 2018-11-09
Add version meta-info in arctic module


dimosped:
#### Arctic Version

```
1.72.0
```

#### Arctic Store

```
VersionStore, TickStore, or ChunkStore
```

#### Description of problem and/or code sample that reproduces the issue

Currently we don't have version information in arctic top level module.

PR #654, opened 2018-11-09
Add version to arctic module

dimosped:
Implements feature #653


yschimke:
@dimosped what is the expected version for any of

- a clean git build off master
- a dirty build of master
- a tagged build

yschimke:
@dimosped exaplined its always just the version from setup.py

Issue #658, opened 2018-11-13
Write/append errors for Panel objects from older pandas versions


dimosped:
#### Arctic Version

```
1.72.0
```

#### Arctic Store

```
VersionStore
```

#### Description of problem and/or code sample that reproduces the issue

Writting pandas Panel objects fails with older pandas as they lack an 'index' attribute:
https://github.com/manahl/arctic/blob/master/arctic/store/_pandas_ndarray_store.py#L219

PR #659, opened 2018-11-13
Check if panel has index first

dimosped:
Solves issue #658

### DIFF
--- a/arctic/store/_pandas_ndarray_store.py
+++ b/arctic/store/_pandas_ndarray_store.py
@@ -216,7 +216,7 @@ class PandasPanelStore(PandasDataFrameStore):
     def can_write(self, version, symbol, data):
         if self.can_write_type(data):
             frame = data.to_frame(filter_observations=False)
-            if NP_OBJECT_DTYPE in frame.dtypes.values or data.index.dtype is NP_OBJECT_DTYPE:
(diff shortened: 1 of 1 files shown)


arctic/store/_pandas_ndarray_store.py
```diff
@@ -216,7 +216,7 @@ def can_write_type(data):
     def can_write(self, version, symbol, data):
         if self.can_write_type(data):
             frame = data.to_frame(filter_observations=False)
-            if NP_OBJECT_DTYPE in frame.dtypes.values or data.index.dtype is NP_OBJECT_DTYPE:
+            if NP_OBJECT_DTYPE in frame.dtypes.values or (hasattr(data, 'index') and data.index.dtype is NP_OBJECT_DTYPE):
```
yschimke: Do you need to check the cases above? PandasDataFrameStore.can_write, PandasSeriesStore.can_write?

dimosped: I haven't seen breakages internally for them, and AFAIK there can't be Dataframe/Series without an index attribute.
```
import pandas as pd

df = pd.DataFrame(data={'a':[1,2,3]}, index=None)
print("{}: {}".format(type(df), hasattr(df, 'index')))

ser = pd.Series(data=range(3), index=None)
print("{}: {}".format(type(ser), hasattr(df, 'index')))
```
prints:
```
<class 'pandas.core.frame.DataFrame'>: True
<class 'pandas.core.series.Series'>: True
```

Issue #665, opened 2018-11-27
Collect all arctic global configuration variables/flags in arctic._config and document their use


dimosped:
#### Arctic Version

```
1.72.0
```

#### Arctic Store

```
All stores
```

#### Description of problem and/or code sample that reproduces the issue

Collect all arctic global configuration variables/flags in arctic._config and document their use

yschimke:
I think we still need a further/later PR to support a different model of configuration, that allows for per connection etc.  But happy with this consolidation.

bmoscon:
Probably also need to document them as well, right?

dimosped:
@yschimke  @bmoscon I add here some ideas for future reference.

We currently have one MongoClient instance (which we mistakenly refer to as 'connection') per arctic instance.
https://github.com/manahl/arctic/blob/master/arctic/arctic.py#L139

In practice, each mongo client instance has an internal connection pool (1-thread/connection I believe), each connection using own underlying TCP socket.

We use a configuration of  4 for maxPoolSize:
https://github.com/manahl/arctic/blob/master/arctic/arctic.py#L140

So there is no notion of having a configuration per connection, but maybe instead could discuss having per:
- Arctic instance
- Library object instance
- Library globally (part of a lib's metadata stored in mongo)

Any of the above could e.g. receive upon construction a custom 'ArcticConfiguration' object instance, which includes all the necessary configuration options.

Keep in mind that there as some options that are module-level (i.e. some LZ4 ones thread pool size) and can't be custom per instance.

dimosped:
@bmoscon  this is now merged, and yes, all options in arctic/_config.py have been well documented.

shashank88:
Seems to be done, worth closing?

PR #666, opened 2018-11-27
moving all env variable configuration driven variables in arctic._config

dimosped:
Implements feature #665  

@yschimke  @jamesblackburn @bmoscon

### DIFF
--- a/benchmarks/fast_serializable_check.py
+++ b/benchmarks/fast_serializable_check.py
@@ -28,13 +28,13 @@ def _bench(rounds, input_df, fast):
 def assess_speed(df_kind):
     rounds = 100
     input_df = input_test_data()[df_kind][0]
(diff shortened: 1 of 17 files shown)


benchmarks/fast_serializable_check.py
```diff
@@ -28,13 +28,13 @@ def _bench(rounds, input_df, fast):
 def assess_speed(df_kind):
     rounds = 100
     input_df = input_test_data()[df_kind][0]
-    orig_config = anr._FAST_CHECK_DF_SERIALIZABLE
+    orig_config = anr.FAST_CHECK_DF_SERIALIZABLE
```
yschimke: Should this still point at anr? or at config?

dimosped: this is fine, as it will use the module level namespace which contains the import

PR #667, opened 2018-11-29
Add config docs

dimosped:
Adds documentation for #665

### DIFF
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -97,18 +97,20 @@ export ARCTIC_FORWARD_POINTERS_CFG=1
 
 The following table documents the compatibility for reading and writing data for all possible combinations of *ARCTIC_FORWARD_POINTERS_CFG*.
 
-|                                                  |   Version written with    | Version written with | Version written with |<br>
-|                                                  | legacy Arctic / DISABLED  |        HYBRID        |       ENABLED        |<br>
-| ------------------------------------------------ | ------------------------- | -------------------- | -------------------- |<br>
-|            Read with Arctic < v1.73.0            |            Y              |          Y           |          -           |<br>
-|  Read with ARCTIC_FORWARD_POINTERS_CFG=DISABLED  |            Y              |          Y           |          Y           |<br>
-|  Read with ARCTIC_FORWARD_POINTERS_CFG=HYBRID    |            Y              |          Y           |          Y           |<br>
-|  Read with ARCTIC_FORWARD_POINTERS_CFG=ENABLED   |            Y              |          Y           |          Y           |<br>
-|                                                  |                           |                      |                      |<br>
-|          Update with Arctic < v1.73.0            |            Y              |          Y           |          -           |<br>
-| Update with ARCTIC_FORWARD_POINTERS_CFG=DISABLED |            Y              |          Y           |          Y(*)        |<br>
-| Update with ARCTIC_FORWARD_POINTERS_CFG=HYBRID   |            Y              |          Y           |          Y(*)        |<br>
-| Update with ARCTIC_FORWARD_POINTERS_CFG=ENABLED  |            Y              |          Y           |          Y           |<br>
+```
+|                                                  |   Version written with    | Version written with | Version written with |
+|                                                  | legacy Arctic / DISABLED  |        HYBRID        |       ENABLED        |
+| ------------------------------------------------ | ------------------------- | -------------------- | -------------------- |
+|            Read with Arctic < v1.73.0            |            Y              |          Y           |          -           |
+|  Read with ARCTIC_FORWARD_POINTERS_CFG=DISABLED  |            Y              |          Y           |          Y           |
+|  Read with ARCTIC_FORWARD_POINTERS_CFG=HYBRID    |            Y              |          Y           |          Y           |
+|  Read with ARCTIC_FORWARD_POINTERS_CFG=ENABLED   |            Y              |          Y           |          Y           |
+|                                                  |                           |                      |                      |
+|          Update with Arctic < v1.73.0            |            Y              |          Y           |          -           |
+| Update with ARCTIC_FORWARD_POINTERS_CFG=DISABLED |            Y              |          Y           |          Y(*)        |
+| Update with ARCTIC_FORWARD_POINTERS_CFG=HYBRID   |            Y              |          Y           |          Y(*)        |
+| Update with ARCTIC_FORWARD_POINTERS_CFG=ENABLED  |            Y              |          Y           |          Y           |
+```
 
 (*) appends will be converted to a full write
 
--- a/docs/configuration.md
+++ b/docs/configuration.md
@@ -4,6 +4,9 @@ Arctic has several tuning knobs under [arctic/_config.py](https://github.com/man
 
 Most of these variables are initialized via environment variables, which are explained in the rest of this section.
 
+
+<br><br>
+
 ## VersionStore
 
 ### STRICT_WRITE_HANDLER_MATCH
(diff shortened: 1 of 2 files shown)


docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+Most of these variables are initialized via environment variables, which are explained in the rest of this section.
+
+
+<br><br>
+
+## VersionStore
+
+### STRICT_WRITE_HANDLER_MATCH
+
+Controls if Arctic can only match the intended handlers for the data type. If set to true it prevents falling back to pickling if the matching handler (based on the type of data) can't serialize without objects.
```
yschimke: Is there a typical error that users see when upgrading, if we could put this below and how to clean your data then it should hopefully come up in google results.

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+
+
+<br><br>
+
+
+## NdArrayStore
+
+### CHECK_CORRUPTION_ON_APPEND
+
+Enables more thorough sanity checks for detecting data corruption when issuing appends. The checks will introduce a 5-7% performance hit. This is disabled by default.
```
yschimke: Is there a programmatic way to do this check? If not, should there be?

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+```
+
+
+### MAX_DOCUMENT_SIZE
+
+This configuration variable is used only when ARCTIC_AUTO_EXPAND_CHUNK_SIZE is set and the user writes a DataFrame which has extremely large number of columns (exceed 2MB serialized).
+
+This value must be less than 16MB which is the maximum document size of MongoDB, taking into account the size of other fields, serialized as a BSON object.
+
+Its default value is configured as follows:
```
yschimke: configured confused me here, maybe calculated?

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+
+<br><br>
+
+
+
+## Forward pointers
+
+### ARCTIC_FORWARD_POINTERS_CFG
+
+This feature flag controls the mode of operation for the data segment references model.
```
yschimke: Should we mark this clearly e.g. orange text as experimental?

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+
+The original implementation of Arctic stores the *_id* values of version document inside the data segment documents (parent references). Therefore, even small updates (i.e. writes where data change very little) cause a large number of documents to be updated, and trigger fetches in the WiredTiger cache (e.g. when appending).
+
+Since Arctic version *1.73.0* the segment referencing model of VersionStore has included a new implementation, named as *forward pointers*. In this model, the segments no longer hold information about the versions which reference/use them, but instead, the list of segment SHAs is stored in the version document itself. This is beneficial in many ways. First, small updates resultin updates/writes only for the new data segments, reducing dramatically the number of affected documents by the update queries (faster execution). Second, the WiredTiger cache gets less polluted utilizing the necessary indexes, not fetching existing large data segments in cache. Finally, all the data information is now in one place, the version document itself, making the debugging of data integrity issues easier.
```
yschimke:
```suggestion
Since Arctic version *1.73.0* the segment referencing model of VersionStore has included a new implementation, named as *forward pointers*. In this model, the segments no longer hold information about the versions which reference/use them, but instead, the list of segment SHAs is stored in the version document itself. This is beneficial in many ways. First, small updates result in updates/writes only for the new data segments, reducing dramatically the number of affected documents by the update queries (faster execution). Second, the WiredTiger cache gets less polluted utilizing the necessary indexes, not fetching existing large data segments in cache. Finally, all the data information is now in one place, the version document itself, making the debugging of data integrity issues easier.
```

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+export ARCTIC_FORWARD_POINTERS_CFG=DISABLED
+
+# This mode of operation maintains both forward pointers and parent references in segments.
+# For reads the forward pointer segment references are preferred if they exist.
+export ARCTIC_FORWARD_POINTERS_CFG=HYBRID
+
+# In this mode of operation, only forward pointers are used, and the created versions are not
+# backwards compatible with older (< v1.73.0) Arctic versions.
+# Note that it is still possible to read versions written with older Arctic versions
+export ARCTIC_FORWARD_POINTERS_CFG=1
```
yschimke:
```suggestion
export ARCTIC_FORWARD_POINTERS_CFG=ENABLED
```

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+When enabled, the number of segments will be cross-verified between forward and legacy (parent) pointers. It is mostly used to verify the correct functionality of forward pointers as long as it is in experimental state.
+
+It has an effect only when *ARCTIC_FORWARD_POINTERS_CFG=HYBRID* and affects writes/appends.
+
+```
+export ARCTIC_FORWARD_POINTERS_RECONCILE=1
+```
+
+
+### FW_POINTERS_REFS_KEY
```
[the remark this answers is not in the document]
dimosped: yeah, was also thinking to remove them

docs/configuration.md
dimosped: will attach a stacktrace (also includes the error message), removing sensitive data

docs/configuration.md
dimosped: There are already two implementations, well documented:
https://github.com/manahl/arctic/blob/master/arctic/store/_version_store_utils.py#L321
and
https://github.com/manahl/arctic/blob/master/arctic/store/_version_store_utils.py#L344

I will have to slightly update this for the FW pointers = ENABLED case were we no longer have parents.

docs/configuration.md
dimosped: It is a code static variable, not configurable via os.env (on purpose).
In this context it seems better to convert to:

```
It is set to the following default value:
```

docs/configuration.md
dimosped: highlighted

docs/configuration.md
dimosped: thanks

docs/configuration.md
```diff
@@ -0,0 +1,214 @@
...
+Variable *ARCTIC_FORWARD_POINTERS_CFG* controls the three modes of operation as described below:
+
+```
+# This is the default mode of operation (i.e. same as not setting the variable).
+# Arctic operates identically to previous versions (<1.73.0).
+export ARCTIC_FORWARD_POINTERS_CFG=DISABLED
+
+# This mode of operation maintains both forward pointers and parent references in segments.
+# For reads the forward pointer segment references are preferred if they exist.
+export ARCTIC_FORWARD_POINTERS_CFG=HYBRID
```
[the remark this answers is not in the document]
dimosped: added this info

docs/configuration.md
dimosped: corrected

dimosped:
@yschimke  I addressed all comments, it is ready to merge.

docs/configuration.md
```diff
@@ -0,0 +1,228 @@
...
+
+### STRICT_WRITE_HANDLER_MATCH
+
+Controls if Arctic can only match the intended handlers for the data type. If set to true it prevents falling back to pickling if the matching handler (based on the type of data) can't serialize without objects.
+
+```
+export STRICT_WRITE_HANDLER_MATCH=1
+```
+
+When strict match is enabled and handler fails to serialize without objects, users will reveive the the foolowing error:
```
yschimke:
```suggestion
When strict match is enabled and handler fails to serialize without objects, users will receive the the following error:
```

docs/configuration.md
yschimke: Nice, I actually meant the opposite, e.g. which types of errors that users pickling data may see. But just a suggestion, ok with this.

yschimke:
LGTM

docs/configuration.md
dimosped: corrected, thanks

Issue #668, opened 2018-11-30
Asynchronous arctic requests (multithreading)


dimosped:
#### Arctic Version

```
1.73.0
```

#### Arctic Store

```
All arctic stores
```

#### Description of problem and/or code sample that reproduces the issue

Currently Arctic serves requests via blocking calls, which need to complete serially.
Implement asynchronous capabilities, where users may submit arctic requests, and use returned futures or callback to track completion of the requests.

shashank88:
Hey @dimosped ! Should we close this issue as the implementation seems to be merged or is there more stuff remaining for this?

Issue #655, opened 2018-11-12
[ChunkStore] Decompression error on concurrent read & write


ddebie:
#### Arctic Version

```
1.70.0
```

#### Arctic Store

```
ChunkStore
```

#### Platform and version

CentOS 7.2

#### Description of problem and/or code sample that reproduces the issue
I'm getting an error when I have two threads simultaneously reading and writing from one symbol. The reading thread will periodically throw a LZ4BlockError:

```LZ4BlockError: Decompression failed: corrupt input or insufficient space in destination buffer. Error code: 24049```

Here is the code to reproduce it:
```
def get_library():
    return Arctic('some_host:1234').get_library(LIB_NAME)

def write_loop():
    library = get_library()
    while True:
        data = create_data(100)  # Creates 100 rows of random data
        library.append('SYMBOL', data)
        
def read_loop():
    library = get_library()
    while True:
        df = library.read('SYMBOL')

proc = mp.Process(target=write_loop)
proc.start()

try:
    read_loop()
finally:
    proc.terminate()
```

From a quick check, it seems that the data being passed to decompress(_str) in _compression.py is not valid lz4 - could the the block metadata and data be out of sync?

**Comment by bmoscon** [comment]
I'm pretty sure none of the libraries are threadsafe. There are multiple documents that make up the stored data and you can't read and write to them at the same time (though you can write new data to chunkstore and read back older data that is not part of that chunk at the same time).

correct me if I'm wrong @jamesblackburn

**Comment by yschimke** [comment]
This seems like something Arctic should be handling, and have some documented expectations of what clients might observe.  Potentially with ways to documented ways resolve e.g. either retry on a class of errors (similar to SQL optimistic lock failures) or ways to resolve via config e.g. isolation levels.

**Comment by ddebie** [comment]
I've looked into this a bit more and from what I can tell, it's due to writes not being atomic - there is a short period of time where the metadata document is out of sync with the data. So it seems that the problem exists even when running from a single thread.

Even with a single thread, if an application dies while writing, it's possible to cause permanent data corruption (and I've been able to reproduce this pretty consistently). If I run just the writing code (below), and then randomly kill the process, eventually the symbol will become corrupted and unreadable.

```
def write_loop():
    library = get_library()
    while True:
        data = create_data(100)  # Creates 100 rows of random data
        library.append('SYMBOL', data)```

Is there any way around this? It seems to me that without writes being atomic, permanent data corruption is possible.

bmoscon:
[...]

PR #669, opened 2018-11-30
[EXPERIMENTAL] Initial implementation of the public arctic async api (Python 2 and 3)

dimosped:
Implements feature described in #668 .

This is compatible both with Python 2 and Python 3, without requiring third party libraries.

Currently uses multi-threading, not multi-processing.

The speed gains are roughly 2x for not too large data frames, and 1.5x for large:
```
Experiment results: async=False, chunks/write=2, writes/round=64, rounds=2, parallel_lz4=False, async_arctic_pool_workers=4: ['11.991', '0.000', '11.991', '11.991']
Experiment results: async=True, chunks/write=2, writes/round=64, rounds=2, parallel_lz4=False, async_arctic_pool_workers=2: ['7.028', '0.000', '7.028', '7.028']
Experiment results: async=True, chunks/write=2, writes/round=64, rounds=2, parallel_lz4=False, async_arctic_pool_workers=4: ['5.820', '0.000', '5.820', '5.820']
```
The API allows to get back a future, provide a callback, and run with mongo_retry.

May need an extra check for detecting a fork(), and resetting the pool.

The subclassing of AsyncArctic from the abstract LazySingletonWorkersPool, may seem odd, but this is because an upcoming PR (the internal incremental serialization/async-mongo work), reuses the code of LazySingletonWorkersPool. This upcoming PR is orthogonal, and speeds up individual arctic operations.

This PR doesn't cause any regression risks, it only adds new functionality, without affecting existing code paths. The user needs to explicitly invoke the async API, so no surprises for existing code which uses Arctic.

We could improve the usability of the API, but let's first get its nuts and bolts right and agreed.

### DIFF
--- a/arctic/async/__init__.py
+++ b/arctic/async/__init__.py
@@ -1,0 +1,2 @@
+from .async_arctic import ASYNC_ARCTIC, async_arctic_submit, async_wait_request, async_wait_requests, async_join_all, \
+    async_reset_pool, async_total_requests
--- a/arctic/async/__init__.py
+++ b/arctic/async/__init__.py
@@ -1,2 +1,3 @@
-from .async_arctic import ASYNC_ARCTIC, async_arctic_submit, async_wait_request, async_wait_requests, async_join_all, \
+from .async_arctic import ASYNC_ARCTIC, async_arctic_submit, \
+    async_wait_request, async_wait_requests, async_shutdown, async_await_termination, \
     async_reset_pool, async_total_requests
--- a/arctic/async/async_arctic.py
+++ b/arctic/async/async_arctic.py
@@ -6,7 +6,7 @@ from threading import RLock
 from concurrent.futures import FIRST_COMPLETED
 
 from .async_utils import AsyncRequestType, AsyncRequest
-from ._workers_pool import LazySingletonWorkersPool
+from ._workers_pool import LazySingletonTasksCoordinator
 from ..decorators import mongo_retry
 from ..exceptions import AsyncArcticException
 
@@ -30,7 +30,7 @@ def _arctic_task_exec(request):
     return result
 
 
-class AsyncArctic(LazySingletonWorkersPool):
+class AsyncArctic(LazySingletonTasksCoordinator):
     _instance = None
     _SINGLETON_LOCK = RLock()
     _POOL_LOCK = RLock()
@@ -45,6 +45,8 @@ class AsyncArctic(LazySingletonWorkersPool):
             super(AsyncArctic, self).__init__(pool_size)
             self.requests_per_library = defaultdict(lambda: defaultdict(lambda: defaultdict(list)))
             self.requests_by_id = dict()
+            self.local_shutdown = False
+            self.deferred_requests = list()
 
     def __reduce__(self):
         return "ASYNC_ARCTIC"
@@ -64,6 +66,10 @@ class AsyncArctic(LazySingletonWorkersPool):
         mongo_retry = bool(kwargs.get('mongo_retry'))
         return library_name, symbol, kind, callback, mongo_retry
 
+    def _is_clashing(self, request):
+        return bool(self._get_modifiers(request.library, request.symbol) or
+                    request.kind is AsyncRequestType.MODIFIER and self._get_accessors(request.library, request.symbol))
+
     def _add_request(self, request):
         self.requests_per_library[request.library][request.symbol][request.kind].append(request)
         self.requests_by_id[request.id] = request
@@ -73,6 +79,19 @@ class AsyncArctic(LazySingletonWorkersPool):
         if request.id in self.requests_by_id:
             del self.requests_by_id[request.id]
 
+    def _schedule_request(self, request):
+        # Update the state of tracked tasks
+        self._add_request(request)
+        try:
+            new_id, new_future = self.submit_task(False, _arctic_task_exec, request)
+            request.id = new_id
+            request.future = new_future
+            request.future.add_done_callback(lambda the_future: self._request_finished(request))
+        except Exception:
+            # clean up the state
+            self._remove_request(request)
+            raise
+
     def submit_arctic_request(self, store, fun, is_modifier, *args, **kwargs):
         lib_name, symbol, kind, callback, mongo_retry = AsyncArctic._verify_request(store, is_modifier, **kwargs)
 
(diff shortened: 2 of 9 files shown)


arctic/async/__init__.py
```diff
@@ -0,0 +1,2 @@
+from .async_arctic import ASYNC_ARCTIC, async_arctic_submit, async_wait_request, async_wait_requests, async_join_all, \
+    async_reset_pool, async_total_requests
```
yschimke:
```suggestion
    async_reset_pool, async_total_requests
    
```

arctic/async/async_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    def schedule_delay(self):
+        if None in (self.start_time, self.create_time):
+            raise RequestDurationException("{} can't provide an schedule_delay ({}).".format(
+                self, (self.start_time, self.create_time)))
+        return self.start_time - self.create_time
+
+    @property
+    def total_time(self):
+        if None in (self.end_time, self.create_time):
+            raise RequestDurationException("{} can't provide an total_time ({}).".format(
```
yschimke: Looks like it will double parenthethis this

>>> "a({})b".format((1, 2))
'a((1, 2))b'

and above

arctic/async/async_utils.py
```diff
@@ -0,0 +1,64 @@
...
+    def execution_duration(self):
+        if None in (self.start_time, self.end_time):
+            raise RequestDurationException("{} can't provide an execution_duration ({}).".format(
+                self, (self.start_time, self.end_time)))
+        return self.end_time - self.start_time
+
+    @property
+    def schedule_delay(self):
+        if None in (self.start_time, self.create_time):
+            raise RequestDurationException("{} can't provide an schedule_delay ({}).".format(
```
yschimke:
```suggestion
            raise RequestDurationException("{} can't provide a schedule_delay ({}).".format(
```

arctic/async/async_utils.py
yschimke:
```suggestion
            raise RequestDurationException("{} can't provide a total_time ({}).".format(
```

arctic/async/async_arctic.py
```diff
@@ -0,0 +1,187 @@
...
+            raise errored[0].exception
+
+    def join(self, timeout=None):
+        while len(self.requests_by_id):
+            try:
+                AsyncArctic.wait_requests(self.requests_by_id.values(), timeout=timeout)
+            except Exception as e:
+                logging.exception("Failed to join all requests")
+
+    def total_requests(self):
```
yschimke: total_requests isn't the right name, pending_requests?

arctic/async/async_arctic.py
```diff
@@ -0,0 +1,187 @@
...
+        return alive_requests, done_requests
+
+    @staticmethod
+    def raise_errored(requests):
+        errored = tuple(r for r in requests if r.is_completed and r.exception is not None)
+        if errored:
+            raise errored[0].exception
+
+    def join(self, timeout=None):
+        while len(self.requests_by_id):
```
yschimke: join is probably the wrong name here, and I'm not sure it makes sense for a singleton. multiple clients could be calling this adding requests so your while loop may never terminate.

arctic/async/async_arctic.py
```diff
@@ -0,0 +1,187 @@
...
+        done_requests = [r for r in requests if r.is_completed]
+        if do_raise:
+            AsyncArctic.raise_errored(done_requests)
+        return alive_requests, done_requests
+
+    @staticmethod
+    def raise_errored(requests):
+        errored = tuple(r for r in requests if r.is_completed and r.exception is not None)
+        if errored:
+            raise errored[0].exception
```
yschimke: Raising just one of the requests might hide the real cause, e.g. a first failing async request may leave some internal functionality in an inconsistent state, but may not be the exception that is finally raised.  Java has suppressed exceptions for exactly this reason, we should do something similar here to allow them all to be introspected.

arctic/async/async_arctic.py
```diff
@@ -0,0 +1,187 @@
...
+
+            # Submit the task
+            try:
+                new_id, new_future = self.submit_task(False, _arctic_task_exec, request)
+                request.id = new_id
+                request.future = new_future
+            except Exception:
+                # clean up the state
+                self._remove_request(request)
+                raise
```
yschimke: what exceptions do you expect here?  backpressure from the executor?  Should they be returned through the callback?

arctic/async/async_arctic.py
```diff
@@ -0,0 +1,187 @@
...
+
+    def submit_arctic_request(self, store, fun, is_modifier, *args, **kwargs):
+        lib_name, symbol, kind, callback, mongo_retry = AsyncArctic._verify_request(store, is_modifier, **kwargs)
+
+        for k in ('async_callback', 'mongo_retry'):
+            kwargs.pop(k, None)
+
+        with type(self)._POOL_LOCK:  # class level lock, since it is a Singleton
+            if lib_name:
+                if self._get_modifiers(lib_name, symbol):
```
yschimke: It seems awkward that in async code clients need to deal with this themselves.  Shouldn't the library manage this for callers?

What can a client processing a request do when receiving this error partially through a larger operation?  Seems like all that caller can do is make a *synchronous* blocking retry to continue.

yschimke: I would suggest deferring this part until we have a solid plan.  You can add locking above, but also it doesn't seem like it will stop a concurrent synchronous caller.  Nor a mutator in a different process/host.

arctic/async/async_utils.py
dimosped: indeed :-)

arctic/async/async_arctic.py
dimosped: good one

arctic/async/async_arctic.py
dimosped: yes, the very reason I added the while here, is because other callers/threads may keep adding requests

the intention is to wait until all requests have been served

what about something similar to Java's : 
shutdown() + awaitTermination() 

So we stop accepting new requests with shutdown(), and then we can await until all requests are finished?

arctic/async/async_arctic.py
dimosped: this is a fair point

Internally, the executor will indeed not fail/raise, but rather record the exception:
https://github.com/dimosped/arctic/blob/public-async-final/arctic/async/async_arctic.py#L25

"raise_errored(requests)" is only a convenience static method, not affecting functionally the executor/async_arctic. So it is up to the user to use it to raise any of the requests he provides in the list, it is to quickly raise from a large set of requests which have completed, i.e. raise if any had an exception. This is a separate call, e.g. 

```
requests = []
for s in syms:
    requests.append(async_arctic_submit(lib, lib.write, True, symbol=s, data=data)

async_wait_requests(requests)

try:
    raise_errored(requests)
except:
    # re-run or handle errored ones
    # errored = filter_errored(requests)
```

arctic/async/async_arctic.py
dimosped: this is scheduling related exceptions, not related to the task we are supposed to run, becuase all task exceptions are attached to the request object.

This means that the executor service failed to process/run the request, not that the request's body excecution failed.

we should be really raising this one

arctic/async/async_arctic.py
dimosped: Valid points, as we discussed, I will convert this to an execution deferral approach.

So we have a queue of jobs awaiting to be run which will get delayed as long as a conflicting task is running currently in the pool, to avoid concurrent modification errors like the ones described here: #655 

Note, we guaranty safety for concurrent read/write for the same symbol from multiple threads in the same process here, not across processes.

arctic/async/async_arctic.py
yschimke: Yep - sounds good.

arctic/async/__init__.py
yschimke: end with a newline

arctic/async/async_utils.py
yschimke: Do you want that?

bmoscon:
I'll take a look at this later today. Interested to review it as I'm a heavy asyncio user in Python3

yschimke:
Designwise my biggest concern is the use of subclassing e.g. extending LazySingletonWorkersPool.  Ideally AsyncArctic would just "have" a worker pool etc, and ultimately the public API for AsyncArctic would just be async_arctic_submit method.

Also I'm not convinced that AsyncArtic itself needs to guarantee itself being a singleton, it could just be via a central factory instead.

dimosped:
@bmoscon yes, agreed, it is not asynchronous in the sense of having an event loop (twisted/tornado/py3 async).

This allows to receive arctic requests without blocking until completion, and requests are served via a worker pool beneath the API.

As I have described in person to @yschimke  , indeed, motor and making things truly async is one alternative option I had in mind.

The benefit of this PR is that code don't block, and serving concurrently allows for an significant increase in speed.

There will be another PR (upcoming) which utilizes the incremental serializer and this is a great case scenario for motor (serialize/compress/write-to-mongo) asynchronously and in smaller units of work (instead of monolithic end-2-end design we have now).

Motor under the hood uses either tornado (AFAIK) for py2 comp, and also the py3 async. I am strongly inclined to merge this PR for now so that users can start having a top level API which allows for getting futures/callback and get speed improvements. 

Then, once the truly async/motor-based solution is ready, we can re-release complying to the same public async API, as we have discussed internally with Yury.

This is a quite flexible change, as noone is relying on this ATM and we don't break compatibility/regression. 

I plan to make explicit not in the documentation (to be added) that the async public API is highly experimental and can change in the next few months until we reach the final design.

Thanks for your reviews @bmoscon  @yschimke .
Will finalize this PR in the next few days.

yschimke:
Is there a better name than AsyncArctic for this work?  ConcurrentArctic, CallbackArctic?

Is there a smallest possible public API for this functionality (a single method signature) that would work the same across this impl, a py2 tornado impl and p3 asyncio?

yschimke:
@bmoscon To be honest, if this is good enough for us for code that will be running on python 2 for 18+ months.  Then I think the interesting part would be what would a python 3 only equivalent look like.

Is there value in splitting these 2 APIs completely and having a python 3 async/await API?  My last two non-python languages are Hack and Kotlin, so I'm 110% sold on the async/await world.

PR #671, opened 2018-12-24
Pep8 cleanups

shashank88:
Does some import ordering fixes, newline fixs and small pep8 issues.

### DIFF
--- a/tests/integration/store/test_version_store_corruption.py
+++ b/tests/integration/store/test_version_store_corruption.py
@@ -9,7 +9,7 @@ from tests.integration.chunkstore.test_utils import create_test_data
 symbol = 'TS1'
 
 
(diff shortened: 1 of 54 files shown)


shashank88:
Seems like another major commit before this one. Will rebase on manahl/master and resend it.

tests/integration/store/test_version_store_corruption.py
```diff
@@ -24,8 +24,8 @@ def n_append(library, library_name, total_appends, rows_per_append, bulk_data_ts
     return open_last_row
 
 
-def _corrupt_with_append_only(library, library_name):
-    def do_fail(version):
+def _corrupt_with_append_only(library, _library_name):
```
jamesblackburn: The `_` prefix on `library_name` will mean it doesn't match the pytest fixture, no?

PR #672, opened 2018-12-24
Optimize imports according to pep8 rules

shashank88:
Optimized imports according to:
https://www.python.org/dev/peps/pep-0008/#imports


bmoscon:
looks good - but lets wait until some builds pass before merging, plz. It would be pretty easy to miss something with a diffset of this size :)

shashank88:
@bmoscon had to leave in some imports which were unused in the module they were in buy apparently being used in the global scope by other modules. Seems fine now. Can merge now once I have an ack.

Issue #674, opened 2019-01-03
More formatting fixes, and adding minimal pep8 checks to ensure it's being followed.


shashank88:
#### Description of problem and/or code sample that reproduces the issue

Similar to the #670 more formatting changes (eg. indent issues, more unused stuff) are required for pep8 compliance. Will add some minimal pep8 checking to ensure the checking is integrated in travis

yschimke:
@shashank88 can you remove the boilerplate from the issue you raise?

PR #676, opened 2019-01-06
Multiple small formatting changes and enabling pycodestyle in ci

shashank88:
This has multiple small changes made as was browsing the code. Fixes mostly fall into these categories:
* double quote in docstrings
* remove unnecessary brackets
* marking unused imports
* initializing dicts in one go
* Indents
* Variable renames

Also enabled pycodestyle with some disabled flags (eg. line length etc are disabled). The advantage being if any of the fixed issues now break, it will not allow that change. 

`pycodestyle --ignore=E122,E126,E223,E501,E731,W503,E722,W504 arctic`
should pass from now on.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,6 +8,8 @@ python:
 services:
    - mongodb
 
+before_install:
+  - pip install pycodestyle
 install:
   - mongo --version
   - pip install --upgrade pip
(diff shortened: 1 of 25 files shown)


shashank88:
Now ci also contains a basic code styling check (pycodestyle which is the new name for pep8). Build seems fine.

shashank88:
Closing https://github.com/manahl/arctic/issues/674

Issue #678, opened 2019-01-06
Builds fail with pytest 4.1


shashank88:
Looking at: https://github.com/pytest-dev/pytest/issues/4546
this breaks our builds as the error iirc reflected missing get_marker.

shashank88:
https://github.com/pytest-dev/pytest-cov/pull/253 should fix it once they release.

shashank88:
We can lock pytest to the previous successful version (4.0.2 from travis) if the release is not done by tomorrow.

PR #679, opened 2019-01-06
Fix build by pinning pytest<4.1

shashank88:
Pinning pytest to <4.1 to fix this: 
https://github.com/manahl/arctic/issues/678

Not sure when pytest-cov will release next, this should unblock us for now.


shashank88:
Builds passed with this, Merging

Issue #587, opened 2018-07-23
Add cache for list_libraries


jamesblackburn:
#### Description of problem and/or code sample that reproduces the issue

List libraries is O(n) for number of libraries, so is slow for large numbers of databases and libraries, and over high latency links.  Provide a cache of the libraries in a top-level database to speed up initial connections.

dimosped:
@jamesblackburn  reg the "high latency links", is the RTT "payed" multiple times during the qyery, (i.e. iterative), or this is simply for the TCP in order to transfer a large response?

jamesblackburn:
For each db, it lists all the collections:

```
    @mongo_retry
    def list_libraries(self):
        """
        Returns
        -------
        list of Arctic library names
        """
        libs = []
        for db in self._conn.database_names():
            if db.startswith(self.DB_PREFIX + '_'):
                for coll in self._conn[db].collection_names():
                    if coll.endswith(self.METADATA_COLL):
                        libs.append(db[len(self.DB_PREFIX) + 1:] + "." + coll[:-1 * len(self.METADATA_COLL) - 1])
            elif db == self.DB_PREFIX:
                for coll in self._conn[db].collection_names():
                    if coll.endswith(self.METADATA_COLL):
                        libs.append(coll[:-1 * len(self.METADATA_COLL) - 1])
        return libs
```

yschimke:
We are also discussing using this to avoid locking up when there are global locks on the database cc @dimosped @rob256

shashank88:
Have something simple'ish working, just need to add some tests.

PR #690, opened 2019-01-11
Add a cached version of list_libraries

shashank88:
list_libraries is a frequently called operation that can get
quite slow with a high amount of libraries. I noticed 4-5s+
for 7-8k libraries. Also the number of libraries is fairly
constant, so not much point in refetching the entire list every
time.

This adds a new collection in mongo which acts as a global cache
for all queries. For now it's called list_libraries_cache but I
guess long term we can have: list_libraries(cached=True).

### DIFF
--- a/arctic/_config.py
+++ b/arctic/_config.py
@@ -92,3 +92,9 @@ BENCHMARK_MODE = False
 # ---------------------------
 # Configures the size of the workers pools used for async arctic requests
 ARCTIC_ASYNC_NWORKERS = os.environ.get('ARCTIC_ASYNC_NWORKERS', 4)
+
+
+# -------------------------------
+# Flag used for indicating caching levels. For now just for list_libraries.
+# -------------------------------
+ENABLE_CACHE = bool(os.environ.get('ENABLE_CACHE'))
--- a/arctic/arctic.py
+++ b/arctic/arctic.py
@@ -191,7 +191,13 @@ class Arctic(object):
         -------
         list of Arctic library names
         """
-        return self._list_libraries_cached() if ENABLE_CACHE else self._list_libraries()
+        if ENABLE_CACHE:
+            try:
+                return self._list_libraries_cached()
+            except OperationFailure as exc:
+                logging.warning("Could not fetch from cache due to %s", exc)
+
+        return self._list_libraries()
 
     @mongo_retry
     def _list_libraries(self):
--- a/arctic/arctic.py
+++ b/arctic/arctic.py
@@ -2,11 +2,13 @@ import logging
 import os
 import re
 import threading
+from datetime import datetime
 
 import pymongo
 from pymongo.errors import OperationFailure, AutoReconnect
 from six import string_types
 
+from ._config import ENABLE_CACHE
 from ._util import indent
 from .auth import authenticate, get_auth
 from .chunkstore import chunkstore
@@ -183,13 +185,16 @@ class Arctic(object):
     def __setstate__(self, state):
         return Arctic.__init__(self, **state)
 
-    @mongo_retry
     def list_libraries(self):
         """
         Returns
         -------
         list of Arctic library names
         """
+        return self._list_libraries_cached() if ENABLE_CACHE else self._list_libraries()
+
+    @mongo_retry
+    def _list_libraries(self):
         libs = []
         for db in self._conn.list_database_names():
             if db.startswith(self.DB_PREFIX + '_'):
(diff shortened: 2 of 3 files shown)


shashank88:
For https://github.com/manahl/arctic/issues/587

arctic/arctic.py
```diff
@@ -202,6 +203,42 @@ def list_libraries(self):
...
+    def invalidate_libraries_cache(self):
+        """
+        Deletes the data in the collection used to cache the result
+        of list_libraries
+        :return:
+        """
+        self._conn.cache_db.cached_lib_names.drop()
+
+    @mongo_retry
+    def list_libraries_cached(self):
```
yschimke: I suggest caching should be a flag for the list libraries call, instead of a separate method.  Maybe sensible values are USE, REFRESH, IGNORE?

arctic/arctic.py
```diff
@@ -202,6 +203,42 @@ def list_libraries(self):
...
         return libs
 
+    @mongo_retry
+    def invalidate_libraries_cache(self):
+        """
+        Deletes the data in the collection used to cache the result
+        of list_libraries
+        :return:
+        """
+        self._conn.cache_db.cached_lib_names.drop()
```
yschimke: I presume dropping a collection is heavyweight, which might go against the idea of a cache here.

yschimke:
As this is a cache, can we write with the lowest write acknowledgement requirements?

arctic/arctic.py
shashank88: So my mongo knowledge is limited but I changed my remove({}) to drop based on https://stackoverflow.com/questions/12147783/mongodb-trade-offs-of-dropping-a-collection-vs-removing-all-of-its-documents/12148106
Although I guess in a 1 element db remove might be cheaper.

bmoscon:
this isnt something that currently affects me, so I defer to @jamesblackburn in terms of if this is the solution he had in mind

arctic/arctic.py
shashank88: Do you want to have some thing specific for list_libraries or a generic flag for all caching, if we do cache more stuff in the future?

arctic/arctic.py
```diff
@@ -202,6 +203,42 @@ def list_libraries(self):
...
+    @mongo_retry
+    def list_libraries_cached(self):
+        """
+        Returns
+        -------
+        List of Arctic library names from a cached collection (global per mongo cluster) in mongo.
+        Long term list_libraries should have a use_cached argument.
+        """
+        mongo_db = self.__conn.cache_db
+        col_name = 'cached_lib_names'
```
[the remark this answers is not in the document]
shashank88: Even if we do have that we can have a separate database called (say) meta_db and have a new collection for each type of metadata/accounting stuff rather than reusing this.

arctic/arctic.py
```diff
@@ -202,6 +207,44 @@ def list_libraries(self):
...
         return libs
 
+    @mongo_retry
+    def invalidate_cache(self):
+        """
+        Deletes the data in the collection used to cache the result
+        of list_libraries
+        :return:
+        """
+        self._conn.meta_db.cache.drop()
```
yschimke: Let's discuss tomorrow, I think have the cache exist over time is still preferable and simpler than deleting dbs or collections.

arctic/_config.py
```diff
@@ -92,3 +92,9 @@ class FwPointersCfg(Enum):
 # ---------------------------
 # Configures the size of the workers pools used for async arctic requests
 ARCTIC_ASYNC_NWORKERS = os.environ.get('ARCTIC_ASYNC_NWORKERS', 4)
+
+
+# -------------------------------
+# Flag used for indicating caching levels. For now just for list_libraries.
+# -------------------------------
+ENABLE_CACHE = bool(os.environ.get('ENABLE_CACHE'))
```
yschimke: Should we change the default for this to on?

shashank88: Yeah, it's not a risky thing to turn on by default. If we are doing that then maybe we should have cache invalidation when a new library is initialized / renamed given a common flow for people might be to list_libraries on creating a new one, and it might be confusing to not see their newly created libs.

yschimke: @shashank88 Good idea.  nit: instead of basic cache invalidation, write the updated results.

yschimke: My thinking is, that we can probably have a reasonably high TTL on the cache if most mutations eventually update as well.

arctic/arctic.py
```diff
@@ -183,13 +185,16 @@ def __getstate__(self):
...
         return Arctic.__init__(self, **state)
 
-    @mongo_retry
     def list_libraries(self):
         """
         Returns
         -------
         list of Arctic library names
         """
+        return self._list_libraries_cached() if ENABLE_CACHE else self._list_libraries()
```
rob256: Should we check for authorisation errors? This assumes we have `read`, and sometimes `dbOwner` rights to `meta_db`.

shashank88:
@jamesblackburn any comments on this one?

arctic/arctic.py
```diff
@@ -202,6 +213,44 @@ def list_libraries(self):
...
         return libs
 
+    @mongo_retry
+    def invalidate_cache(self):
+        """
+        Deletes the data in the collection used to cache the result
+        of list_libraries
+        :return:
+        """
+        self._conn.meta_db.cache.drop()
```
jamesblackburn: Not sure about drop, here - it may be reasonably heavyweight in the database. Can we just delete_many or similar?

arctic/arctic.py
```diff
@@ -202,6 +213,44 @@ def list_libraries(self):
...
+        List of Arctic library names from a cached collection (global per mongo cluster) in mongo.
+        Long term list_libraries should have a use_cached argument.
+        TODO: Maybe convert this to a decorator so this code can be used in a generic way.
+        """
+        mongo_db = self._conn.meta_db
+        col_name = 'cache'
+        if col_name not in mongo_db.collection_names():
+            mongo_db.create_collection(col_name).create_index("date", expireAfterSeconds=3600)
+
+        cache_col = mongo_db[col_name]
```
jamesblackburn: How does this interact with authentication on the cluster?

arctic/arctic.py
```diff
@@ -202,6 +213,44 @@ def list_libraries(self):
...
+        coll_data = cache_col.find_one({"type": "list_libraries"})
+        if coll_data:
+            logger.debug('Library names are in cache.')
+            return coll_data['data']
+        library_names = self._list_libraries()
+        # TODO: Check if ret is serializable to BSON by mongo. For now it's a list of strings.
+        cache_col.update_one(
```
jamesblackburn: Assumes write access to this database which most users won't have => lead to an exception? Is it worth separating out the cache update from the usage?

arctic/arctic.py
shashank88: I discussed this with @rob256, and we are going to create a role similar to list_databases and grant readWrite to all users for this db. Till the permissions are granted it should just throw an exception and use the slow path (the old list_libraries).

arctic/arctic.py
shashank88: Yeah, it will throw an exception but will be caught in the list_libraries code and default to the slow path. I am not sure if it will be trivial to schedule it in the background. An option is to regenerate it on initialize_library and have a very high expiry time, but that slows initialize library every time.

arctic/arctic.py
```diff
@@ -202,6 +213,44 @@ def list_libraries(self):
...
+            {"type": "list_libraries"},
+            {"$set": {"type": "list_libraries", "date": datetime.utcnow(), "data": library_names}},
```
jamesblackburn: (this limits us to 16MB of library_names by length - which is probably quite a lot :)

shashank88:
Closing this as I have opened https://github.com/manahl/arctic/pull/717 to fix this

PR #697, opened 2019-01-18
Add basic benchmarking for simple loads with fwd_pointers

shashank88:
This adds a simple benchmark with pytest-benchmark with writing
a largeish (400 * 400) dataframe of random floats for multiple
symbols.
Then it appends 1 row for each symbol for 10 days followed by
reading all. The idea is to replicate a flow that's closer to
use cases eg. here it's once a day data.

Will add a similar test for one minute data like load as well.

The code has some duplications, but I want to push the benchmarks
else it just keeps growing locally.

### DIFF
--- a/benchmarks/fwd_pointers/fwd_benchmarks.py
+++ b/benchmarks/fwd_pointers/fwd_benchmarks.py
@@ -57,7 +57,6 @@ def gen_sparse_rows_for_range(n_rows, low, high, dense):
         value = float(random.randrange(low, high))
         repetitions = min(random.randint(0, 20), n_rows - current)
         rows.extend([value] * repetitions)
-        print('current=', current, 'repetitions=', repetitions)
         current += repetitions
 
     print('lensparse=', len(rows))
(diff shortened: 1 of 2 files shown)


benchmarks/fwd_pointers/fwd_benchmarks.py
```diff
@@ -0,0 +1,128 @@
...
+from arctic import Arctic
+from arctic._config import FwPointersCfg
+
+DATAFRAME_DIM = 1000
+N_SYMBOLS = 5
+N_APPENDS = 10
+
+
+class FwPointersCtx:
+    def __init__(self, value_to_test, do_reconcile=True):
```
yschimke: why is do_reconcile=True for the benchmark?

benchmarks/fwd_pointers/fwd_benchmarks.py
```diff
@@ -0,0 +1,128 @@
...
+def test_benchmark_reads(benchmark, fw_pointers_cfg):
+    with FwPointersCtx(fw_pointers_cfg):
+        benchmark(read_all_symbols, fw_pointers_cfg)
+
+# For 10 symbols, 400 * 400 dataframe of dense non compressible floats
+# ------------------------------------------------------------------------------------------------------------- benchmark: 6 tests ------------------------------------------------------------------------------------------------------------
+# Name (time in ms)                                                Min                    Max                   Mean              StdDev                 Median                   IQR            Outliers     OPS            Rounds  Iterations
+# ---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
+# test_benchmark_reads[FwPointersCfg.DISABLED]                450.5870 (1.0)         761.6360 (1.0)         613.6972 (1.0)      142.5660 (1.78)        672.2870 (1.0)        256.6150 (2.06)          2;0  1.6295 (1.0)           5           1
```
yschimke: Should we remove the benchmark results until re-run?

shashank88:
Have changed this a lot, will resend another pull request later.

Issue #699, opened 2019-01-25
BSONStores - remove default sharding


shashank88:
From @rob256 
```Currently on Arctic, BSONStores shard by _id. This is bad for a number of reasons. Here are a couple of good guidelines on sharding:
https://www.mongodb.com/blog/post/on-selecting-a-shard-key-for-mongodb
https://blog.bugsnag.com/mongo-shard-key/

BSONStores shouldn't shard by default. The sharding will need to added afterwards manually which is dependent on the information it stores and how it is used.```

shashank88:
This would also require a collection creation which is not created in the case of BSONStore but created in the case of version store as it does it while creating an index

PR #701, opened 2019-01-25
Bsonstore

shashank88:
to fix https://github.com/manahl/arctic/issues/699
example to show it initializes the relevant collections on initialize:
>>> a.initialize_library('testbson', 'BSONStore')

show collections
test1
test1.ARCTIC
test1.changes
test1.snapshots
test1.version_nums
test1.versions
test2
test2.ARCTIC
test2.changes
test2.snapshots
test2.version_nums
test2.versions
testbson
testbson.ARCTIC

### DIFF
--- a/arctic/store/bson_store.py
+++ b/arctic/store/bson_store.py
@@ -29,17 +29,23 @@ class BSONStore(object):
         self._arctic_lib = arctic_lib
         self._reset()
 
-    @classmethod
-    def initialize_library(cls, arctic_lib, hashed=True, **kwargs):
+    def enable_sharding(self):
         logger.info("Trying to enable sharding...")
+        arctic_lib = self._arctic_lib
         try:
-            if not hashed:
-                logger.warning("Ignored hashed=False when enabling sharding, only hashed=True "
-                               " makes sense when they key is an ObjectId")
             enable_sharding(arctic_lib.arctic, arctic_lib.get_name(), hashed=True, key='_id')
         except OperationFailure as exception:
-            logger.warning(("Library created, but couldn't enable sharding: "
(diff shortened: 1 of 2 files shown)


arctic/store/bson_store.py
```diff
@@ -29,17 +29,23 @@ def __init__(self, arctic_lib):
...
-                logger.warning("Ignored hashed=False when enabling sharding, only hashed=True "
-                               " makes sense when they key is an ObjectId")
             enable_sharding(arctic_lib.arctic, arctic_lib.get_name(), hashed=True, key='_id')
         except OperationFailure as exception:
-            logger.warning(("Library created, but couldn't enable sharding: "
-                            "%s. This is OK if you're not 'admin'"), exception)
+            logger.warning("Could not enable sharding: %s, you probably need admin permissions.", exception)
+
+    @classmethod
+    def initialize_library(cls, arctic_lib, hashed=True, **kwargs):
```
yschimke: nit: should you keep the warning about the useless hashed flag?

shashank88:
Tests seem to be failing in the build which look unrelated on first glance, will take a look

arctic/store/bson_store.py
rob256: I would remove it. It doesn't add any value, and you would never use ranged sharding for ObjectIDs.

shashank88: Removing it would kinda change the signature even though kwargs should swallow hashed=True up this even if provided after removing it from the function signature ... wdyt?

shashank88: Removed the hashed=True argument.

shashank88:
Merging this since the build breakage does not seem related to this PR.

PR #708, opened 2019-02-01
Add basic benchmarks for forward pointers

shashank88:
This can be improved and cleaned up quite a bit, but want to commit
at a checkpoint rather than doing everything locally.

### DIFF
--- a/benchmarks/fwd_pointers/fwd_benchmarks.py
+++ b/benchmarks/fwd_pointers/fwd_benchmarks.py
@@ -1,0 +1,209 @@
+from __future__ import print_function
+
+import argparse
+import random
+from collections import defaultdict
+from datetime import datetime as dt
+from datetime import timedelta as td
+
+import numpy as np
+import pandas as pd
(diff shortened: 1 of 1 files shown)


benchmarks/fwd_pointers/fwd_benchmarks.py
```diff
@@ -0,0 +1,209 @@
...
+    'ASKSIZE': (0.0, 400.0),
+    'BIDSIZE': (0.0, 400.0),
+    'TICK_COUNT': (1.0, 50.0),
+    'VOLUME': (0.0, 1000.0),
+}
+
+APPEND_NROWS = 10
+
+
+class FwPointersCtx:
```
pkittenis: Use new style classes: `class FwPointersCtx(object)`

shashank88: Yeah, I am used to the py3 style classes without the object, forgot about py2.

PR #711, opened 2019-02-06
Fixes #710: Initialize SEGMENT_SHA in versions for writes and appends

### LINKED ISSUE #710: Forward pointers: Upgrading to hybrid mode breaks due to missing SEGMENT_SHAS
Upgraded to arctic 1.73 from 1.67.1 and it caused issues due to a field being missing in the version collection.
`
  File "build/bdist.linux-x86_64/egg/arctic/store/_ndarray_store.py", line 447, in append
    self._do_append(collection, version, symbol, item, previous_version, dirty_append)
  File "build/bdist.linux-x86_64/egg/arctic/store/_ndarray_store.py", line 494, in _do_append
    shas_to_add=(sha,))
  File "build/bdist.linux-x86_64/egg/arctic/store/_ndarray_store.py", line 147, in _update_fw_pointers
    version_shas.update(previous_version[FW_POINTERS_REFS_KEY])
`

shashank88:
This would help preventing issues where we are accessing this key
in versions without being initialized.

This is only done for cases where it's either HYBRID or ENABLED
mode as this field should not be used in other cases anyways.

### DIFF
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -409,6 +409,9 @@ class NdarrayStore(object):
 
         version['type'] = self.TYPE
         version[FW_POINTERS_CONFIG_KEY] = ARCTIC_FORWARD_POINTERS_CFG.name
+        # Create an empty entry to prevent cases where this field is accessed without being there. (#710)
+        if version[FW_POINTERS_CONFIG_KEY] != FwPointersCfg.DISABLED.name:
+            version[FW_POINTERS_REFS_KEY] = list()
 
         if str(dtype) != previous_version['dtype'] or \
                 _fw_pointers_convert_append_to_write(previous_version):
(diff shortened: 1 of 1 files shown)


yschimke:
Some of the code seems like it calls get for this key with a default value of [], which would suggest the intention was to be optional.

Should we either
1) rely on that optionality and use get with default
or
2) always write, and remove the default fallback?

shashank88:
> Some of the code seems like it calls get for this key with a default value of [], which would suggest the intention was to be optional.
> 
> Should we either
> 
> 1. rely on that optionality and use get with default
>    or
> 2. always write, and remove the default fallback?

@yschimke I was looking into the .get( KEY, []) instances of SEGMENT_SHA, and it's there because for disabled mode it still won't be initialized and thus it's fine to have an empty list.

PR #717, opened 2019-02-14
Fix #587 Add a best effort cache option for list_libraries

### LINKED ISSUE #587: Add cache for list_libraries

shashank88:
list_libraries is a frequently called operation that can get
quite slow with a high amount of libraries. I noticed 4-5s+
for 7-8k libraries. Also the number of libraries is fairly
constant, so not much point in refetching the entire list every
time.

This adds a new collection in mongo which acts as a global cache
for all queries. The meta_db can be used for more accounting and
metadata storage in future, and this collection can be used for
caching other stuff as well.

It's best effort as if you don't have permissions or the data
does not exist in the cache it will just use the list_libraries.
The idea is to not require admin for reads and not expect the user
to have write access to a collection on writes.

### DIFF
--- a/arctic/_cache.py
+++ b/arctic/_cache.py
@@ -1,0 +1,67 @@
+import logging
+from datetime import datetime, timedelta
+
+from pymongo.errors import OperationFailure
+
+from ._config import CACHE_COLL, CACHE_DB
+
+logger = logging.getLogger(__name__)
+
+
+class Cache:
+    def __init__(self, client, cache_expiry=3600, cache_db=CACHE_DB, cache_col=CACHE_COLL):
+        self._client = client
+        self._cachedb = client[cache_db]
+        self._cachecol = None
+        try:
+            if cache_col not in self._cachedb.collection_names():
+                self._cachedb.create_collection(cache_col).create_index("date", expireAfterSeconds=cache_expiry)
+        except OperationFailure as op:
+            logging.debug("This is fine if you are not admin. The collection should already be created for you: %s", op)
+
+        self._cachecol = self._cachedb[cache_col]
+
+    def get(self, key, newer_than_secs=-1):
+        """
+
+        :param key: Key for the dataset. eg. list_libraries.
+        :param newer_than_secs: -1 to indicate use cache if available. Used to indicate what level of staleness
+        in seconds is tolerable.
+        :return: None unless if there is non stale data present in the cache.
+        """
+        try:
+            if not self._cachecol:
+                # Collection not created or no permissions to read from it.
+                return None
+            coll_data = self._cachecol.find_one({"type": key})
+            # Check that there is data in cache and it's not stale.
+            if coll_data and (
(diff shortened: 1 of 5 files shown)


arctic/arctic.py
```diff
@@ -183,13 +188,17 @@ def __getstate__(self):
     def __setstate__(self, state):
         return Arctic.__init__(self, **state)
 
-    @mongo_retry
-    def list_libraries(self):
+    def list_libraries(self, newer_than_secs=120):
```
yschimke: It seems like there are a lot of dials for caching.  I think it's worth defaulting to the happy cache path.

e.g. This list_libraries seems useful to have a way to invalidate, but could be defaulted to -1.  That the cache value is always assumed to be ok.

It seems like there are a lot of ways e.g. mismatch between values where your cache will get ignored.

arctic/_cache.py
```diff
@@ -0,0 +1,57 @@
...
+        self._cachecol = None
+        try:
+            if cache_col not in self._cachedb.collection_names():
+                self._cachedb.create_collection(cache_col).create_index("date", expireAfterSeconds=cache_expiry)
+        except OperationFailure as op:
+            logging.debug("This is fine if you are not admin. The collection should already be created for you: %s", op)
+
+        self._cachecol = self._cachedb[cache_col]
+
+    def get(self, key, newer_than_secs=120):
```
yschimke: suggest -1 here be the default which means to always trust the cache

arctic/_config.py
```diff
@@ -92,3 +92,11 @@ class FwPointersCfg(Enum):
 # ---------------------------
 # Configures the size of the workers pools used for async arctic requests
 ARCTIC_ASYNC_NWORKERS = os.environ.get('ARCTIC_ASYNC_NWORKERS', 4)
+
+
+# -------------------------------
+# Flag used for indicating caching levels. For now just for list_libraries.
+# -------------------------------
+ENABLE_CACHE = bool(os.environ.get('ENABLE_CACHE'))
```
yschimke: Can this be defaulted to true?

arctic/arctic.py
```diff
@@ -183,13 +188,17 @@ def __getstate__(self):
...
         """
         Returns
         -------
         list of Arctic library names
         """
+        _ = self._conn  # Ensure the connection exists and cache is initialized with it.
+        return self._list_libraries_cached(newer_than_secs) if ENABLE_CACHE else self._list_libraries()
```
yschimke: newer_than_secs == 0 probably means avoid checking the cache

arctic/arctic.py
shashank88: Makes sense. Will default to the cached version with the default in config also to enable cache

PR #719, opened 2019-02-19
add to cache even for the first item

shashank88:
Was thinking about it, I think upsert might be better here as for someone trying out arctic, list libraries won't work as it won't add the first item.

### DIFF
--- a/arctic/_cache.py
+++ b/arctic/_cache.py
@@ -57,11 +57,15 @@ class Cache:
             logging.debug("This operation is to be run with admin permissions. Should be fine: %s", op)
 
     def append(self, key, append_data):
-        # Not upserting here as this is meant to be use when there is already data for this key.
         try:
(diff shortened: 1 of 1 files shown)


arctic/_cache.py
```diff
@@ -61,7 +61,8 @@ def append(self, key, append_data):
         try:
             self._cachecol.update_one(
                 {'type': key},
-                {'$push': {'data': append_data}}
+                {'$push': {'data': append_data}},
```
yschimke: Do you need setOnInsert also for date and type?

https://stackoverflow.com/a/47004771/1542667

yschimke: Also remove the comment on line 60

shashank88:
Will resend this PR as I want to add some more cases. Not sure how to mark it as draft so will close it for now.

Issue #725, opened 2019-02-26
Backwards incompatibility with pandas < 0.23


shashank88:
With https://github.com/manahl/arctic/pull/724 it uses features that are not in pandas <0.23 which would cause builds to fail or people using older versions of pandas to fail without warning. 

@bmoscon we hit this as sort parameter is not compatible with pandas 0.22 which we use internally.

bmoscon:
i'll get a travis config file running that will run pandas on 0.22.0 on a specific build so these things will be more obvious at commit time

bmoscon:
sounds like i should add a partial or something to bind the kwarg if the pandas version is > 0.24 otherwise leave it as is

shashank88:
sort was added as a possible param in 0.23.0 I think, so yeah a partial should work for +0.23.0 pandas.

shashank88:
https://github.com/manahl/arctic/pull/729 fixes this

PR #729, opened 2019-02-27
Version Fixes and new travis setup

bmoscon:
* Fixes pandas compatibility issues
* Removes reserved word from package name
* Updates travis to use an older version of pandas (0.22.0) for python 2.7 tests


bmoscon:
@shashank88 setting pandas to 0.22.0 for python2.7 seems to break several things. Not sure if we want to support a version that old or not (or older?).

yschimke:
@bmoscon happy to defer decision to you 3, but as an open source maintainer, people wanting newer versions of Arctic probably want newer version of pandas (or should be tolerant of it).

v0.22.0 (December 29, 2017)

bmoscon:
@yschimke I agree 100%. I'm fine saying anything older than 0.23.0 may or may not work (i.e. unsupported). The fact that @shashank88 opened this ticket makes me think that some may still be on ancient versions of pandas: https://github.com/manahl/arctic/issues/725 ;)

shashank88:
@bmoscon yeah, internally we are using pandas 0.22.0 from the build logs which is also a requirement for the release job which is why I raised the ticket.

bmoscon:
@shashank88 looks fixed now!

shashank88:
Thanks @bmoscon ! Will release 1.74 tomorrow morning

PR #740, opened 2019-03-26
Fix the case of moving from disabled to hybrid / fwd mode

shashank88:
Currently we have issues when moving from disabled -> fwd mode with preexisting data. This was because:
a) usage of 'is' vs '=='
b) using the returned find results vs the 'sha' of the objects.


shashank88:
Closing this as we have decided to disable and revert the forward pointers stream of work for now.

PR #743, opened 2019-04-03
Chunkstore Improvements + fixes

bmoscon:
* Improve column list copy + add test
* fix exception when empty symbol has get_info called on it


bmoscon:
The ticket I was looking into was already fixed, but I made some improvements here anyways

bmoscon:
@shashank88 @jamesblackburn @yschimke should be good to merge, take a look if you want

PR #747, opened 2019-04-19
reduce chunkstore memory footprint

TomTaylorLondon:


arctic/serialization/numpy_arrays.py
```diff
@@ -218,16 +222,19 @@ def deserialize(self, data, columns=None):
...
-            df = pd.concat([self.converter.objify(d, columns) for d in data], ignore_index=not index)
+            dfs = []
+            while len(data):
+                dfs.append(self.converter.objify(data.pop(0), columns))
+            df = pd.concat(dfs, ignore_index=not index)
+            del dfs
 
         if index:
-            df = df.set_index(meta[INDEX])
+            df = df.set_index(meta[INDEX], inplace=True)
```
bmoscon: did you test this? i thought if `inplace` was true it didn't need a return (it may still return something, I've always just omitted the variable assignment while doing inplace operations on dataframes)

yschimke:
What's the memory saving?  Have you measured it? 50% only 1 copy of data instead of 2?

Would be great to have a way to show the saving, and automated test to avoid any accidental regressions.

arctic/serialization/numpy_arrays.py
```diff
@@ -218,16 +222,19 @@ def deserialize(self, data, columns=None):
...
-            if len(columns) > len(set(columns)):
-                raise Exception("Duplicate columns specified, cannot de-serialize")
+                columns = list(set(columns))
 
         if not isinstance(data, list):
             df = self.converter.objify(data, columns)
         else:
-            df = pd.concat([self.converter.objify(d, columns) for d in data], ignore_index=not index)
+            dfs = []
+            while len(data):
```
bmoscon: is there a reason you modify the loop control variable in place during iteration? Is this for memory savings?

arctic/serialization/numpy_arrays.py
yschimke: why reassigned to df?

arctic/serialization/numpy_arrays.py
```diff
@@ -218,16 +222,19 @@ def deserialize(self, data, columns=None):
             if index:
                 columns = columns[:]
                 columns.extend(meta[INDEX])
-            if len(columns) > len(set(columns)):
-                raise Exception("Duplicate columns specified, cannot de-serialize")
+                columns = list(set(columns))
```
bmoscon: I'm not sure I see this as a win. It seems like the caller may have a bug if they are specifying duplicate columns, we're just hiding the error now

arctic/serialization/numpy_arrays.py
```diff
@@ -203,13 +203,17 @@ def deserialize(self, data, columns=None):
...
+            this will modify data
 
         Returns
         -------
         pandas dataframe or series
         """
         if not data:
             return pd.DataFrame()
+        if not inplace:
+            data = data[:]
```
yschimke: Does this hurt the case when there is no index?  e.g. you are making a defnesive copy even though you won't hit the set_index call?

arctic/serialization/numpy_arrays.py
TomTaylorLondon: Good catch.

arctic/serialization/numpy_arrays.py
TomTaylorLondon: The current logic is confusing when subsetting data frames with indexes. For example, if you have the data frame:
index: date, security
columns: price, volume

The logic works if the user passes:
['price']
Raises a duplicate columns error when passing:
['date','security','price']

I don't see the value in the check - it should just do the right thing..

arctic/serialization/numpy_arrays.py
TomTaylorLondon: You save having two copies of the data in memory. Will attach memory profile.

arctic/serialization/numpy_arrays.py
TomTaylorLondon: fixed

TomTaylorLondon: fixed.

arctic/serialization/numpy_arrays.py
TomTaylorLondon: Attached.

arctic/serialization/numpy_arrays.py
TomTaylorLondon: data is a list of the raw bytes. To save memory I allow the user to convert the list into dataframes in-place. This switch copies list first (shallow) so the caller doesn't get their reference to data changed. No impact to the index.

arctic/serialization/numpy_arrays.py
bmoscon: Using pandas nomenclature the columns and the index are separate. If there is an index, you always get it back, even if you specify a subset of columns (and even if they do not include the index columns). Maybe the documentation should be improved. If for example, you specify price and security, you'll still get date as well as price and security, so your fix would only introduce more weirdness (in my opinion).

TomTaylorLondon: We could remove index columns from columns and then check for duplicates. This keeps the nomenclature but keeps the user interface 'minimum surprise'. Or raise an error saying they have included index columns in the column list.

bmoscon: The result would be the same though, no? You'd supply index columns and it wont complain. I foresee someone opening a bug complaining they only specified 1 of 3 index columns but still got all 3 back.

bmoscon: @TomTaylorLondon not trying to hold up the PR. Is allowing the user to specify all columns (even index columns) something you need? Could emit a warning log and allow them to use the index columns. Would that be ok?

arctic/serialization/numpy_arrays.py
bmoscon: there are some errors in the tests, so I'm thinking this will need to be tweaked a bit more

arctic/serialization/numpy_arrays.py
TomTaylorLondon: Yes - it's bitten us a few times. As a user as your never quite sure what is in the index and what is in the columns, so it's hard to know exactly what to specify here.

bmoscon: ok, so lets leave your change in then. I'll probably tweak `get_info` to separate out the index labels from the columns so then you will be able to know. Whenever i get that done, I'll probably change this back to columns only (no index labels). Sound ok?

TomTaylorLondon: Sure

TomTaylorLondon: Actually, in retrospect, that means breaking the API for clients. How about we keep the fuzziness for clients and simply output a warning instead?

bmoscon: ok, lets see a warning :) but i still think that get info should change, otherwise how would you ever know how to rid yourself of the warning?

TomTaylorLondon: Agree with get_info change

bmoscon: so sounds like you just need to fix the broken tests and add the log and we're all set :D

bmoscon:
@TomTaylorLondon are you going to have the bandiwdth to finish this or would you like me to resolve it?

shashank88:
Hi @TomTaylorLondon any luck with this?

bmoscon:
@shashank88 I spoke with @TomTaylorLondon and am going to take this over from him. I'll get it all fixed up later this week(end).

shashank88:
> @shashank88 I spoke with @TomTaylorLondon and am going to take this over from him. I'll get it all fixed up later this week(end).

👍

PR #750, opened 2019-04-26
Fix issue with pickled numpy arrays in py2 read back in py3

shashank88:
We have an issue where datetimes / numpy arrays etc written in py2
can't be read back now. This tries to use 'bytes' as encoding and
if that does not work, it tries to use latin. This method is given
in the official docs:
https://docs.python.org/3/library/pickle.html#pickle.load

### DIFF
--- a/arctic/store/_pickle_store.py
+++ b/arctic/store/_pickle_store.py
@@ -3,6 +3,7 @@ import logging
 from operator import itemgetter
 
 import bson
+import six
 from bson.binary import Binary
 from bson.errors import InvalidDocument
 from six.moves import cPickle, xrange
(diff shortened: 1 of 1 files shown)


arctic/store/_pickle_store.py
```diff
@@ -54,7 +55,19 @@ def read(self, mongoose_lib, version, symbol, **kwargs):
...
                 except:
                     logger.error("Failed to read symbol %s" % symbol)
-            return pickle_compat_load(io.BytesIO(data))
+
+            if six.PY2:
+                # Providing encoding is not possible on PY2
+                return pickle_compat_load(io.BytesIO(data))
+            else:
+                try:
+                    return pickle_compat_load(io.BytesIO(data), encoding='bytes')
```
yschimke: How could this ever fail?  Isn't this not decoding?

shashank88:
https://stackoverflow.com/questions/28218466/unpickling-a-python-2-object-with-python-3?answertab=oldest#tab-top

arctic/store/_pickle_store.py
yschimke: Dicsussed, not the bytes -> str conversion, but the pickled object has other conversions that fail on bytes.  So try the fast option first.  Then fallback to assuming latin which could be unsafe for conversions but makes the other conversions work.

PR #754, opened 2019-04-29
Fixes 753: Add details to bulk write failures

shashank88:
Adds an error message with details when this happens.

### DIFF
--- a/arctic/store/_ndarray_store.py
+++ b/arctic/store/_ndarray_store.py
@@ -5,7 +5,7 @@ from operator import itemgetter
 import numpy as np
 import pymongo
 from bson.binary import Binary
-from pymongo.errors import OperationFailure, DuplicateKeyError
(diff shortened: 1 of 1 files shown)


yschimke:
Isn't a better exception to make the final exception handling more aware of detail?  Logging errors at multiple layers seems like an anti-pattern.

In Java you would normally log when the exception is finally caught, including in a retry handler etc, use suppressions for chaining or not losing errors with retries.

shashank88:
> Isn't a better exception to make the final exception handling more aware of detail? Logging errors at multiple layers seems like an anti-pattern.
> 
> In Java you would normally log when the exception is finally caught, including in a retry handler etc, use suppressions for chaining or not losing errors with retries.

I don't think there is any final exception handling here. The jobs that failed yesterday due to this error, just crash without logging anything and by default the exception is not informative.

PR #756, opened 2019-04-30
Add a way to force bytes columns to unicode via a config flag

shashank88:
This adds a config flag which allows users to convert the py2 str
type columns (which are bytes) to be unicode when read back. This
is useful for arctic users migrating from python2 to python3 and
expecting back a py3 string type type which is unicode.


shashank88:
Needs another fix. Will resend

PR #757, opened 2019-04-30
Add a way to force bytes columns to unicode via a config flag

shashank88:
This adds a config flag which allows users to convert the py2 str
type columns (which are bytes) to be unicode when read back. This
is useful for arctic users migrating from python2 to python3 and
expecting back a py3 string type type which is unicode.

### DIFF
--- a/arctic/store/_pandas_ndarray_store.py
+++ b/arctic/store/_pandas_ndarray_store.py
@@ -197,7 +197,7 @@ class PandasDataFrameStore(PandasStore):
 
     def read(self, arctic_lib, version, symbol, **kwargs):
         item = super(PandasDataFrameStore, self).read(arctic_lib, version, symbol, **kwargs)
-        return self.SERIALIZER.deserialize(item)
+        return self.SERIALIZER.deserialize(item, force_bytes_to_unicode=kwargs.get('force_bytes_to_unicode', False))
 
     def read_options(self):
         return super(PandasDataFrameStore, self).read_options()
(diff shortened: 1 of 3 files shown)


yschimke:
Do these environment variables work well for us? Should this be at least exposed as a global that can be toggled? My concern is apps that need to set it but may lose a dependency race and config is already read

shashank88:
> Do these environment variables work well for us? Should this be at least exposed as a global that can be toggled? My concern is apps that need to set it but may lose a dependency race and config is already read
> 
> Sent with [GitHawk](http://githawk.com)

I agree with this. I find it a bit confusing to have a config that depends on the order of imports by importlib. I generally use this as this seems to be the unified system we did the configuration for Arctic. For this case, I am happy to have a global var that's inited when Arctic is initialized and can be toggled.

arctic/store/_pandas_ndarray_store.py
```diff
@@ -197,7 +198,7 @@ def append(self, arctic_lib, version, symbol, item, previous_version, **kwargs):
 
     def read(self, arctic_lib, version, symbol, **kwargs):
         item = super(PandasDataFrameStore, self).read(arctic_lib, version, symbol, **kwargs)
-        return self.SERIALIZER.deserialize(item)
+        return self.SERIALIZER.deserialize(item, force_bytes_to_unicode=FORCE_BYTES_TO_UNICODE)
```
pablojim: is it worth allowing this to be specified/overidden as part of the kwargs?

shashank88: I was discussing this with @yschimke whether we should do kwargs args in read vs global, but I assumed the use case was so common that people will just set it as a global once and use it like that. What do you prefer?

shashank88:
Fabian mentioned the indexes and column names should also undergo the same change if force is set. Will make that change and resend.

PR #763, opened 2019-05-01
Do not use bytes when loading back pickled objects

shashank88:
This causes an issue with unicode strings in py2. Instead I am using
something similar to:

https://github.com/pandas-dev/pandas/blob/master/pandas/io/pickle.py#L157

where the default is the better option rather than forcing it to use
bytes.

### DIFF
--- a/arctic/store/_pickle_store.py
+++ b/arctic/store/_pickle_store.py
@@ -61,13 +61,14 @@ class PickleStore(object):
                 return pickle_compat_load(io.BytesIO(data))
             else:
                 try:
-                    return pickle_compat_load(io.BytesIO(data), encoding='bytes')
+                    # The default encoding is ascii.
+                    return pickle_compat_load(io.BytesIO(data))
                 except UnicodeDecodeError as ue:
-                    logger.error("Could not load Pickled object in PY3 with bytes encoding: %s" % ue)
-                # Using encoding='latin1' is required for unpickling NumPy arrays and instances of datetime, date and
-                # time pickled by Python 2: https://docs.python.org/3/library/pickle.html#pickle.load
-                encoding = kwargs.get('encoding', 'latin_1')  # Check if someone has manually specified encoding.
-                return pickle_compat_load(io.BytesIO(data), encoding=encoding)
+                    # Using encoding='latin1' is required for unpickling NumPy arrays and instances of datetime, date
+                    # and time pickled by Python 2: https://docs.python.org/3/library/pickle.html#pickle.load
+                    logger.info("Could not Unpickle with ascii, Using latin1.")
+                    encoding = kwargs.get('encoding', 'latin_1')  # Check if someone has manually specified encoding.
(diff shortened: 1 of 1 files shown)


shashank88:
Will add more info to this when I reach home.

yschimke:
This is a bit of a can of worms. Is it worth changing it further eg using config to select the encoding (including none)? Or just make this one specific change?

shashank88:
@yschimke @pablojim 
So I spent a fair bit of time trying to understand this and did some experiments with pickling non-ascii chars/date objects. 
Summary: 
By default when the Unpickler comes across 8-bit string instances it defaults to assuming ‘ascii’ as the encoding of the pickled object and decodes it to a python3 style string. The issue this can cause is if you wanted to keep it like py2 / py3 bytes style strings it would still convert it. To prevent It you can say encoding=‘bytes’ and it will now read it as a bytes object. eg. you want to dictionary keys to stay as bytes or if you have image data and you want it to be read as bytes and don't want to attempt interpreting it as character data. This also means `encoding=bytes` is not a normal use case for us.

Often you have objects pickled which have characters which don’t fit in ascii, i.e in 7 bits and require 8 bits to be decoded. For example when we get an error like:
```UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 1: ordinal not in range(128)```
This basically says 0xe3 = 227 doesn't fit in ascii and tells you to pass in a different encoding scheme so it can be decoded properly. Latin1 is a pretty common superset of ascii that works for chars up to 256. This happens with numpy arrays and date time/date objects as they probably are using these characters while serializing. 

Examples:
Python2 write:
```
>>> import pickle
>>> from datetime import datetime
>>> s = {'a': datetime.now(), 'b': 1, 'c': u'q'}
>>> pickle.dump(a, open('test.p', 'w'))
>>> pickle.dump(s, open('test.p', 'w'))
```

Python3 reading back: 
```>>> s = pickle.load(open('test.p','r'), encoding='bytes')
  File "<stdin>", line 1, in <module>
TypeError: a bytes-like object is required, not 'str'
>>> s = pickle.load(open('test.p','rb'))
  File "<stdin>", line 1, in <module>
UnicodeDecodeError: 'ascii' codec can't decode byte 0xe3 in position 1: ordinal not in range(128)
>>> s = pickle.load(open('test.p','rb'), encoding='latin1')
>>> s
{'a': datetime.datetime(2019, 5, 1, 23, 57, 35, 777654), 'c': 'q', 'b': 1}```

shashank88:
The conclusion based on above and what I saw with a case with unpickling some class with characters in the range 128-256, was attempting to decode with the default (ascii) and then if that fails trying latin1 should work for the most likely cases.

shashank88:
Build fails due to indent. WIll fix merge tomorrow morning

yschimke:
Thanks, nice analysis. And appreciate the extra mile on this. 

If we can’t make it reliably handle all the real world data we have. Maybe the best aim is for predictable and explainable. And get people away from pickling long term data. 

PR #765, opened 2019-05-02
Use astype vs using str accessor to convert to unicode

shashank88:
When read back with py3, the str accessor for pandas 0.22.0 does not
work for indexes if all data is bytes type. I added a test for py3
reading as well, and astype works for all these cases.

### DIFF
--- a/arctic/serialization/numpy_records.py
+++ b/arctic/serialization/numpy_records.py
@@ -278,18 +278,21 @@ class DataFrameSerializer(PandasSerializer):
             # of people migrating to py3. # https://github.com/manahl/arctic/issues/598
             # This should not be used for a normal flow, and you should instead of writing unicode strings
             # if you want to work with str in py3.,
-            def convert_pandas_column_to_unicode(col):
-                return col.str.decode('utf-8')
 
             for c in df.select_dtypes(object):
+                # The conversion is not using astype similar to the index as pandas has a bug where it tries to convert
+                # the data columns to a unicode string, and the object in this case would be bytes, eg. b'abc'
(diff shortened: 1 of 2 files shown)


arctic/serialization/numpy_records.py
```diff
@@ -279,7 +279,7 @@ def deserialize(self, item, force_bytes_to_unicode=False):
             # This should not be used for a normal flow, and you should instead of writing unicode strings
             # if you want to work with str in py3.,
             def convert_pandas_column_to_unicode(col):
-                return col.str.decode('utf-8')
+                return col.astype('unicode')
```
yschimke: Discussed F2F, +1 to your plan to add/update the explanation link

shashank88:
With py3, trying to convert a bytes column to unicode with: index.str.decode('utf-8') breaks with:
```
/users/is/ahlpypi/medusa_egg_cache/36-1/p/pandas-0.22.0+ahl1-py3.6-linux-x86_64.egg/pandas/core/strings.py in _make_accessor(cls, data)
   1920                            "(i.e. inferred_type is 'string', 'unicode' or "
   1921                            "'mixed')")
-> 1922                 raise AttributeError(message)
   1923             if data.nlevels > 1:
   1924                 message = ("Can only use .str accessor with Index, not "

AttributeError: Can only use .str accessor with string values (i.e. inferred_type is 'string', 'unicode' or 'mixed')
```
https://pandas.pydata.org/pandas-docs/version/0.22/generated/pandas.DataFrame.astype.html given here seems to be the proper way to do this that works in py2 / py3

PR #767, opened 2019-05-03
Handle multi indexes when converting to unicode

shashank88:
+ add a test + new version info

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,8 @@
 ## Changelog
 
+### 1.79.1 (2019-05-03)
+  * Bugfix: Pandas convert multiindexes to unicode if flag set.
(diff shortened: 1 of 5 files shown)


CHANGES.md
```diff
@@ -1,5 +1,8 @@
 ## Changelog
 
+### 1.79.1 (2019-05-03)
```
yschimke: Why not 1.80.0?

yschimke: Discussed F2F, if the policy is minor bug fix == patch, then I presume new feature = minor, and major functional (breaking) change is major.  Can you add to the README?

shashank88: Added to releasing.md_

PR #772, opened 2019-05-10
Add global settings for caching list_libraries

shashank88:
This allows us to enable and disable the cache on the fly for list_libraries.
Also allows us to change the expiry period for returning the results.

If user does not have write permissions to meta_db, it will use the slow path. To fix this, you can give permissions using:

db.getSiblingDB('admin').grantRolesToUser("research", [{role: "readWrite", db: "meta_db"}])

### DIFF
--- a/arctic/_cache.py
+++ b/arctic/_cache.py
@@ -8,6 +8,11 @@ logger = logging.getLogger(__name__)
 CACHE_COLL = 'cache'
 CACHE_DB = 'meta_db'
 CACHE_SETTINGS = 'cache_settings'
+"""
+Sample cache_settings collection entry:
+meta_db.cache_settings.insertOne({"enabled": true, "cache_expiry": 600})
+meta_db.cache_settings.find(): { "_id" : ObjectId("5cd5388b9fddfbe6e968f11b"), "enabled" : false, "cache_expiry" : 600 }
+"""
 DEFAULT_CACHE_EXPIRY = 3600
 
 
--- a/arctic/_cache.py
+++ b/arctic/_cache.py
@@ -3,13 +3,16 @@ from datetime import datetime, timedelta
 
 from pymongo.errors import OperationFailure
 
-from ._config import CACHE_COLL, CACHE_DB
-
 logger = logging.getLogger(__name__)
+
+CACHE_COLL = 'cache'
+CACHE_DB = 'meta_db'
+CACHE_SETTINGS = 'cache_settings'
+DEFAULT_CACHE_EXPIRY = 3600
 
 
 class Cache:
-    def __init__(self, client, cache_expiry=3600, cache_db=CACHE_DB, cache_col=CACHE_COLL):
+    def __init__(self, client, cache_expiry=DEFAULT_CACHE_EXPIRY, cache_db=CACHE_DB, cache_col=CACHE_COLL):
         self._client = client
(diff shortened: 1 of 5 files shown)


arctic/_cache.py
```diff
@@ -83,3 +119,11 @@ def update_item_for_key(self, key, old, new):
         # This op is not atomic, but given the rarity of renaming a lib, it should not cause issues.
         self.delete_item_from_key(key, old)
         self.append(key, new)
+
+    def is_caching_enabled(self):
+        # Caching is enabled unless explicitly disabled.
+        cache_settings = self._get_cache_settings()
+        print('cache settings = ', cache_settings)
```
yschimke: remove?

arctic/_config.py
```diff
@@ -94,13 +94,6 @@ class FwPointersCfg(Enum):
 ARCTIC_ASYNC_NWORKERS = os.environ.get('ARCTIC_ASYNC_NWORKERS', 4)
 
 
-# -------------------------------
-# Flag used for indicating caching levels. For now just for list_libraries.
-# -------------------------------
-ENABLE_CACHE = not bool(os.environ.get('DISABLE_CACHE'))
```
yschimke: Probably the right call to move away from system properties here.  But make sure Ray knows.

arctic/_cache.py
```diff
@@ -3,43 +3,79 @@
...
-
 logger = logging.getLogger(__name__)
 
+CACHE_COLL = 'cache'
+CACHE_DB = 'meta_db'
+CACHE_SETTINGS = 'cache_settings'
+DEFAULT_CACHE_EXPIRY = 3600
+
 
 class Cache:
```
yschimke: I suggest showing the config entries as an example here.  So it's clear that enabled is a key etc.

arctic/_cache.py
shashank88: Done

arctic/_cache.py
shashank88: Done.

arctic/_cache.py
```diff
@@ -3,43 +3,84 @@
...
+            return
+
+        if CACHE_SETTINGS not in self._cachedb.list_collection_names():
+            logging.info("Creating %s collection for cache settings" % CACHE_SETTINGS)
+            self._cachedb[CACHE_SETTINGS].insert_one({
+                'enabled': enabled,
+                'cache_expiry': DEFAULT_CACHE_EXPIRY
+            })
+        else:
+            self._cachedb[CACHE_SETTINGS].update_one({}, {'$set': {'enabled': enabled}})
```
rob256: This forces `cache_settings` to only ever have a single document.

My vision was to have a `meta_db.global_settings` collection. In here you can store individual documents for the settings you want:
```
{
  _id: 'library_cache_enabled',
  value: False
},
{
  _id: 'library_cache_expiry',
  value: 600
}
```

Then we can put all global settings into a single collection. What do you think?

shashank88: So I generally like having one document for a setting as this is read fairly often so I didn't want to get into writing / reading atomically and having 2 reads instead of 1. But yeah, I could have one settings doc per application in the same global_settings collection. Your call. Don't feel too strongly about it. 

I can change the name to just meta_db.settings for now and move throttling to use this collection later?

PR #778, opened 2019-05-20
Black formatting v2

shashank88:
As @jamesblackburn suggested I have skipped the string normalization for now and changed to 120. Have only formatted src files for now and not tests as that requires more work eg. Moving stuff to data etc.

### DIFF
--- a/arctic/__init__.py
+++ b/arctic/__init__.py
@@ -3,11 +3,16 @@
 from .arctic import Arctic, register_library_type
 from .arctic import VERSION_STORE, TICK_STORE, CHUNK_STORE
 from .store._ndarray_store import NdarrayStore
-from .store._pandas_ndarray_store import PandasDataFrameStore, PandasSeriesStore, PandasPanelStore
+from .store._pandas_ndarray_store import (
+    PandasDataFrameStore,
+    PandasSeriesStore,
+    PandasPanelStore,
+)
 from .store.version_store import register_versioned_storage, register_version
 
 try:
     from pkg_resources import get_distribution
+
     str_version = get_distribution(__name__).version.strip()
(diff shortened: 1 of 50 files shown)


bmoscon:
I'll refrain from reviewing because I'm kind of anti-black. Line like this to me invalidate the argument that this formatting makes things better or easier to read. Its a personal choice 🤷‍♂ 
https://github.com/manahl/arctic/pull/778/files#diff-63f2797d12f7fd46e9c8345736f1ac36R37

shashank88:
> I'll refrain from reviewing because I'm kind of anti-black. Line like this to me invalidate the argument that this formatting makes things better or easier to read. Its a personal choice 🤷‍♂
> https://github.com/manahl/arctic/pull/778/files#diff-63f2797d12f7fd46e9c8345736f1ac36R37

In my last place, we had the same rules as the formatting rule you pointed to. Basically, if something goes multiline you split each argument to a separate line. That's the thing with autoformatters, that no 2 people will fully agree with the end result but the end goal is to be good enough and save time in the long term. gofmt was something I loved when I did some Golang, even though I didn't agree with some of the formattings it did as everyone did the same thing

bmoscon:
true. I actually happen to like go's formatting. Thats part of the language standard though, black is not

shashank88:
This is funny. pycodestyle check failed after running black formatting. Did not expect this 😆

yschimke:
Would we disable that, if we use black?

shashank88:
> Would we disable that, if we use black?

We shouldn't have to. pycodestyle is the python standard pep8 checker and black is supposed to be compatible with pep8. Will take a look later today.

shashank88:
Closing this. Will recreate a PR after reverting forward pointers.

PR #780, opened 2019-05-21
Fix #777: Handle zero columns when converting to unicode

### LINKED ISSUE #777: DataFrameSerializer.deserialize function should handle empty columns
#### Arctic Version

```
arctic-1.79.1-
```

#### Arctic Store

```
VersionStore
```

#### Platform and version

All

https://github.com/manahl/arctic/blob/master/arctic/serialization/numpy_records.py#L304  needs a check for length of columns. A DataFrame can have non-zero length and still 0 columns.
Currently the code breaks for such cases.

shashank88:
Currently it will break for empty columns due to the missing check.

### DIFF
--- a/arctic/serialization/numpy_records.py
+++ b/arctic/serialization/numpy_records.py
@@ -301,8 +301,8 @@ class DataFrameSerializer(PandasSerializer):
                 if type(df.index[0]) == bytes:
                     df.index = df.index.astype('unicode')
 
-            if type(df.columns[0]) == bytes:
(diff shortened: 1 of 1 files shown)


shashank88:
> Seems worthy of a test?
> 
> https://github.com/manahl/arctic/blob/master/tests/unit/serialization/test_numpy_records.py

I added integration tests (but not unit tests) for this here: https://github.com/manahl/arctic/blob/master/tests/integration/store/test_pandas_store.py#L1061

PR #790, opened 2019-07-02
Convert log to debug

shashank88:
This log can be quite spammy as people generally do not not have listCollections on this db.


yschimke:
LGTM

PR #791, opened 2019-07-03
RFC: Admit invalid bson keys in PickleStore

scriada:
When saving BSON encodable objects to the `PickleStore`, mongo will
raise an `InvalidName` if any of the fields contain "." or "$". Since we
don't actually need to query these fields (afaik), we can relax this constraint.

### DIFF
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -526,7 +526,7 @@ class VersionStore(object):
             # upon intermittent Mongo errors
             # If, however, we get a DuplicateKeyError, suppress it and raise OperationFailure, so that the method-scoped
             # mongo_retry re-tries and creates a new version, to overcome the issue.
-            mongo_retry(self._versions.insert_one)(version)
+            mongo_retry(self._versions.insert)(version, check_keys=False)
         except DuplicateKeyError as err:
             logger.exception(err)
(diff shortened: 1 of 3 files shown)


arctic/store/version_store.py
```diff
@@ -526,7 +526,7 @@ def _insert_version(self, version):
             # upon intermittent Mongo errors
             # If, however, we get a DuplicateKeyError, suppress it and raise OperationFailure, so that the method-scoped
             # mongo_retry re-tries and creates a new version, to overcome the issue.
-            mongo_retry(self._versions.insert_one)(version)
+            mongo_retry(self._versions.insert)(version, check_keys=False)
```
scriada: `insert_one()` doesn't support `check_keys`

richardbounds: Unfortunately I suspect the `check_keys` functionality is deprecated along with `insert`, as it doesn't appear in any of the new methods.

scriada: Good point. Reading around it seems that while mongo supports dots in field names, they are strongly discouraged. Given it's not supported in the new API (and I can't find any hints that it will ever be), this change could come back to bite us.

shashank88:
Discussed with @scriada in person, we will skip this and instead just skip the bson encoding with an env flag. If it's pickled and stored in the data collection we should be fine.

Issue #792, opened 2019-07-04
Unable to save dict with "x.y" fields in VersionStore


scriada:
#### Arctic Version

```
1.79.2
```

#### Arctic Store

```
VersionStore
```

#### Platform and version

python 2.7

#### Description of problem and/or code sample that reproduces the issue

Trying to save a dictionary with keys like "x.y" in a `VersionStore` will raise an `InvalidDocument` error from mongo e.g.

```py
blob = {'foo.bar': {'baz': 1}}
library.write('BLOB', blob)
```

This is because for small bson-compatible objects the `PickleStore` writes the data to the version collection (https://github.com/manahl/arctic/blob/master/arctic/store/_pickle_store.py#L83) so mongo treats the dict as a document.,

shashank88:
So what I am thinking is maybe having a flag that tells PickleStore to not store stuff in the version collections (I am not sure why it even does it in the first place?). That would cause the dict to be pickled and compressed instead of being bson encoded and then fail when writing.

shashank88:
@scriada merged. Closing this

PR #794, opened 2019-07-05
Add a config option to skip bson encoding in pickle store

shashank88:
Currently we try to bson encode if the data is less than a given size and store it in
the version collection, but pickling might be preferable if we have characters that don't
play well with the bson encoder or if you always want your data in the data collection.


shashank88:
For: https://github.com/manahl/arctic/issues/792

scriada:
thanks @shashank88 , lgtm!

Issue #796, opened 2019-07-05
Investigate failing py3 build with multiple test failures


shashank88:
`
E       AssertionError: assert 'Library test' in '<ExceptionInfo LibraryNotFoundException tblen=3>'
E        +  where '<ExceptionInfo LibraryNotFoundException tblen=3>' = str(<ExceptionInfo LibraryNotFoundException tblen=3>)
`
Multiple tests are failing with the above.

shashank88:
I think the issue is that pytest.raises now returns: https://docs.pytest.org/en/latest/reference.html#exceptioninfo
and str(e) no longer returns just the stringified error code. Possible that we might have to use . getrepr to get the same thing? Possibly breaking due to a new pytest version for py3?

shashank88:
This broke the tests:  https://github.com/pytest-dev/pytest/pull/5479
pytest 5.0.1 changes pytest.raises behaviour and it's a py3 only release.

shashank88:
`
In [5]: def a():
   ...:     raise BlahException("testing 12 333")

In [10]: with pytest.raises(Exception) as e:
    ...:     a()
    ...:
    ...: print(str(e))
    ...: print(type(e))
    ...:
    ...:
    ...:
<ExceptionInfo BlahException tblen=2>
<class '_pytest._code.code.ExceptionInfo'>

`

shashank88:
Need to change the the code to use str(e.value) instead of str(e) for py3. Need to check if the same works with py2.

shashank88:
Fixed with PR: #797

PR #797, opened 2019-07-05
Fix tests broken due to https://github.com/pytest-dev/pytest/issues/5412

shashank88:
This breaks our tests as we use str(e) everywhere and as the above
PR changes the behaviour of str to be the same as repr which is to
print the ExceptionInfo Object, which breaks our tests as we expect str(e) to contain the message


shashank88:
Fixes https://github.com/manahl/arctic/issues/796

PR #800, opened 2019-07-09
Revert "Forward pointers implementation (#651)"

shashank88:
Attempt v1

This reverts commit 6735c0400540bfbed3b2e8c61bbff2d856139f33.


bmoscon:
any reason we cant revert these things? They also cause a slew of test failures

PR #821, opened 2019-10-16
Fix read with date_range where start but no end

scoriiu:


shashank88:
@scoriiu can you add the test as @bmoscon mentioned?

scoriiu:
Hi guys, will try adding a test these days. Very buys with other projects :(

PR #823, opened 2019-10-18
Don't retry BulkWriteErrors

eeaston:
These can contain index and key errors that must be raised immediately. 
If we retry, then the error that is raised is often different to the one that caused the original problem - for example duplicate key error for entries that have just been added.

### DIFF
--- a/arctic/decorators.py
+++ b/arctic/decorators.py
@@ -3,7 +3,8 @@ import sys
 from functools import wraps
 from time import sleep
 
-from pymongo.errors import AutoReconnect, OperationFailure, DuplicateKeyError, ServerSelectionTimeoutError
+from pymongo.errors import (AutoReconnect, OperationFailure, DuplicateKeyError, ServerSelectionTimeoutError,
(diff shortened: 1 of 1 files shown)


arctic/decorators.py
```diff
@@ -47,7 +48,7 @@ def f_retry(*args, **kwargs):
             while True:
                 try:
                     return f(*args, **kwargs)
-                except (DuplicateKeyError, ServerSelectionTimeoutError) as e:
+                except (DuplicateKeyError, ServerSelectionTimeoutError, BulkWriteError) as e:
```
jamesblackburn: Out of interest, what was the underlying error? Should we inspect `e.details` and handle the different cases?

jamesblackburn: Agreed, i think this change is sensible.

shashank88:
@eeaston can you rerun your build? Have fixed it in master. Let's merge this.

bmoscon:
@eeaston any reason to not merge? build passes

PR #825, opened 2019-10-25
Make the maximum bson encode size configurable

ChrisBradshaw1:


arctic/_config.py
```diff
@@ -102,3 +102,8 @@ class FwPointersCfg(Enum):
 # play well with the bson encoder or if you always want your data in the data collection.
 # -------------------------------
 SKIP_BSON_ENCODE_PICKLE_STORE = bool(os.environ.get('SKIP_BSON_ENCODE_PICKLE_STORE'))
+
+# -------------------------------
+# Maximum size to bson encode.
```
shashank88: Can you add a longer description Maybe something like "Maximum size up to which the input will be bson_encoded and stored in the version doc instead of being pickled in the version store"

arctic/store/_pickle_store.py
```diff
@@ -85,7 +84,7 @@ def write(self, arctic_lib, version, symbol, item, _previous_version):
             try:
                 # If it's encodeable, then ship it
                 b = bson.BSON.encode({'data': item})
-                if len(b) < _MAX_BSON_ENCODE:
+                if len(b) < MAX_BSON_ENCODE:
```
shashank88: Should we add a check that falls back to using pickle if the object size is larger than say 10MB, given this will be stored in the single version document?

shashank88:
Looks good, just some small comments.

arctic/_config.py
ChrisBradshaw1: Ok

arctic/store/_pickle_store.py
ChrisBradshaw1: Sure

shashank88:
The test is failing for some unrelated issue, let me rebuild.

shashank88:
Merging this @ChrisBradshaw1

PR #826, opened 2019-10-29
Disable test for now to fix build

shashank88:
This works locally for me, I wonder if it's due to a pymongo version
change going by the "cannot find enablesharding" in the test logs.


shashank88:
@bmoscon I saw you removed the docstring for compat on pandas-dev, apparently, it's breaking our build since we do use that arg.

PR #841, opened 2020-02-06
Updated SeriesSerializer to work with pandas 1.0

helingai:
Changed `Series.from_array` to `Series` since the former no longer works with pandas 1.0.


bmoscon:
thanks for this pr @helingai - I fixed all the build errors so this wont be necessary.

PR #846, opened 2020-02-18
Allow writing of `None` data in transactions

tjw45:
Previously attempting to Copy a symbol which had a value of None stored against it would silently fail due to a difference between the treatment of write(<symbol>, None) in transactions and the VersionStore itself.


bmoscon:
hey @tjw45 thanks for the PR. Unfortunately the library is in a bit of a bad state with all the deprecations and warnings. I'm working on a clean up that should resolve a lot of the pandas issues that are causing the build to fail. Once thats squared away we can re-run the tests here and hopefully get this merged.

PR #849, opened 2020-03-24
Adding license boilerplate

ceallen:
Using the abridged version of boilerplate that's suggested in LGPL 3.0.  

Only question I'd have is if we want to change the Acknowledgements section in the README to Authors - otherwise we're referring to a section in the readme that doesn't explicitly exist.


dunckerr:
needs rebasing.

PR #851, opened 2020-04-02
Optimize Chunkstore mongo read query

BaiBaiHi:
Optimize chunkstore read query by joining metadata collection instead of querying for metadata on each individual date which accounted for about 40% of the processing time. Profile results below are run on data with 365 chunks. 

**Original Profile**
image

**New Profile**
image


BaiBaiHi:
Looks like tests are failing since it's trying to build with mongodb version 3.4.10. 
Syntax for the $lookup/$let pattern is only available after mongodb version 3.6...

bmoscon:
@BaiBaiHi  there are other things deprecated that are breaking the build as well. I'm working on fixing them. @jamesblackburn whats the min mongodb version we need to support? 3.4 is quite old, can we bump it to 3.6 or 4.0+?

jamesblackburn:
Yep I think we can bump to 4.0.x (CC @rob256 )

arctic/chunkstore/chunkstore.py
```diff
@@ -262,22 +262,28 @@ def read(self, symbol, chunk_range=None, filter_data=True, **kwargs):
...
+                'pipeline': [
+                    {'$match': {
+                        '$expr': {
+                            '$and': [{'$eq': ['${}'.format(SYMBOL), '$$symbol']},
+                                     {'$eq': ['${}'.format(START), '$$start']}]}}},
+                    {'$project': {'_id': 0, 'm': 0}}], 
+                'as': METADATA}},
+            {'$sort': {START: 1,
+                       SEGMENT: 1}}],
+        allowDiskUse=True)
```
jamesblackburn, line 281: Would be useful to have some comments on what the pipeline is doing.  Particularly if we're spilling to disk would be good to understand why this is needed?

rob256:
> Yep I think we can bump to 4.0.x (CC @rob256 )

Yep, let's go to the latest 4.0 (4.0.17).

bmoscon:
I've already updated the version of mongo on my branch that addresses a lot of the deprecations and what not, so once thats merged in, this branch can be rebased off that

bmoscon:
@BaiBaiHi can you rebase your fork off mainline? I updated many things to fix the build, so assuming after that the tests pass on this PR we can get it merged

BaiBaiHi:
@bmoscon 
Looks like the Python 3 build is failing for tests/integration/tickstore/test_ts_write.py: test_ts_write_pandas due to column order. 

The change I made should have no effect on tickstore and it looks like it's only an issue in the python 3 build.

bmoscon:
@BaiBaiHi sure but can you fix the style issues?

arctic/chunkstore/chunkstore.py:278:55: W291 trailing whitespace
arctic/chunkstore/chunkstore.py:282:27: E261 at least two spaces before inline comment
arctic/chunkstore/chunkstore.py:282:134: W291 trailing whitespace

arctic/chunkstore/chunkstore.py
BaiBaiHi: Updated with comments. The aggregation pipeline has a memory limit for each step so in order to enable it for large datasets, we set the allowDiskUse flag. 

https://docs.mongodb.com/manual/reference/method/db.collection.aggregate/#example-aggregate-method-external-sort

BaiBaiHi:
Yeah of course. My bad. Just set up my IDE on this machine. Didn't realize that my auto-formatter/checker wasn't turned on.

bmoscon:
@jamesblackburn it seems reasonable and it passes all the tests. I'm not sure how it will perform if actually need to spill out onto disk

bmoscon:
@BaiBaiHi @jamesblackburn what do you want to do with this?

bmoscon:
@jamesblackburn - just another ping - I think this seems reasonable from my testing

shashank88:
Should we merge this?

crazy25000:
If it's good to go we should merge this. If not, are there outstanding issues that I could help with? Would love to have consistent performance.

TomTaylorLondon:
Can we hold off merging

You are correct in the issue diagnostics but the root cause of this is a missing index for the query. I have a fix which is simply adding the correct index and should be more efficient. Let me put out a PR and you can test locally? Let me know if this works!

Tom

TomTaylorLondon (changes requested):
Test new fix

arctic/chunkstore/chunkstore.py
```diff
@@ -262,22 +262,29 @@ def read(self, symbol, chunk_range=None, filter_data=True, **kwargs):
...
-            # when len(segments) == 1, this is essentially a no-op
-            # otherwise, take all segments and reassemble the data to one chunk
-            chunk_data = b''.join([doc[DATA] for doc in segments])
-            chunks[segments[0][SYMBOL]].append({DATA: chunk_data, METADATA: mdata})
+        # Join metadata collection to top level collection on "sym" and "s"(start)
+        joined = self._collection.aggregate([
+            {'$match': spec},
+            {'$lookup': {
+                'from': self._mdata.name,
+                'let': {'symbol': '${}'.format(SYMBOL),
```
bmoscon, line 271: I forget, are we still supporting python 2.7? If not, can we use f-strings here?

bmoscon (approved):

arctic/chunkstore/chunkstore.py
bmoscon: @TomTaylorLondon - assuming you're using something like this in your code, the comment above is for you

TomTaylorLondon:
See https://github.com/man-group/arctic/pull/902

vietlq:
@BaiBaiHi this is an exciting PR. Do you want to carry on and test against fix #902?

PR #853, opened 2020-04-04
Fix Build

bmoscon:
This updates a few things in the code, uses a newer version of ubuntu for the tests, which also uses a newer version of mongodb, and allows for use of python 3.7

Please review only - do not merge. Need to finalize what version of pandas needs to be used on 2.7

### DIFF
--- a/arctic/serialization/numpy_records.py
+++ b/arctic/serialization/numpy_records.py
@@ -3,6 +3,7 @@ import logging
 
 import numpy as np
 from pandas import DataFrame, MultiIndex, Series, DatetimeIndex, Index
+import pandas as pd
 
 # Used in global scope, do not remove.
 from .._config import FAST_CHECK_DF_SERIALIZABLE
@@ -20,7 +21,7 @@ except ImportError:
 
 
 log = logging.getLogger(__name__)
-
+PD_VER = pd.__version__
 DTN64_DTYPE = 'datetime64[ns]'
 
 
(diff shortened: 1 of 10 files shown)


bmoscon:
@jamesblackburn @shashank88 this should be good to review and merge now - take a look and let me know if you have any questions.

arctic/serialization/numpy_records.py
```diff
@@ -253,7 +254,10 @@ def deserialize(self, item, force_bytes_to_unicode=False):
                 if len(index) and type(index[0]) == bytes:
                     index = index.astype('unicode')
 
-        return Series.from_array(data, index=index, name=name)
+        if PD_VER < '0.23.0':
```
jamesblackburn, line 257: Note lexicographic version comparisons don't work right in general for all versions... I guess this is OK for the pandas versions we care about?

bmoscon: I tested for 0.22.0, 0.23.x, 0.24.x and 1.x.x and it worked as expected

jamesblackburn: Ok cool. I wonder at some point if we can abstract out the pandas compat code so it doesn't need to exist everywhere, or maybe it'll all go away once we hit 1.x *hopes*

bmoscon: if you can get to 0.24.0 these will go away :D

PR #856, opened 2020-05-02
fix(list_symbols): Use IXSCAN queries for the versions collection when querying symbols

rob256:
This is mainly a reversion on https://github.com/man-group/arctic/pull/520, but we add another index to avoid a FETCH stage which gives a massive speedup.

### DIFF
--- a/arctic/store/version_store.py
+++ b/arctic/store/version_store.py
@@ -79,6 +79,8 @@ class VersionStore(object):
                                          background=True)
         collection.versions.create_index([('symbol', pymongo.ASCENDING), ('version', pymongo.DESCENDING)], unique=True,
                                          background=True)
+        collection.versions.create_index([('symbol', pymongo.ASCENDING), ('version', pymongo.DESCENDING), ('metadata.deleted', pymongo.ASCENDING)],
+                                         background=True)
         collection.version_nums.create_index('symbol', unique=True, background=True)
         for th in _TYPE_HANDLERS:
             th._ensure_index(collection)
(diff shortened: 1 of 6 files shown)


arctic/store/version_store.py
```diff
@@ -79,6 +79,8 @@ def _ensure_index(self):
                                          background=True)
         collection.versions.create_index([('symbol', pymongo.ASCENDING), ('version', pymongo.DESCENDING)], unique=True,
                                          background=True)
+        collection.versions.create_index([('symbol', pymongo.ASCENDING), ('version', pymongo.DESCENDING), ('metadata.deleted', pymongo.ASCENDING)],
```
jamesblackburn, line 82: As this compound index extends the previous one, I think you can remove the one on line 80-81 as this will be used instead (and means you don't need to maintain 2 indexes).

rob256:
> Very nice - presumably this fixes quite a longstanding performance blooper?

Yeah. We moved from an IXSCAN+FETCH to a COLLSCAN, and this will allow us to do a straight IXSCAN. I've seen long queries on a collection which has a million documents, and this gives it a really nice speed up.

arctic/store/version_store.py
rob256: The index on 80-81 is a unique index. We can't make this one unique as it will allow both a deleted and active version to exist. And we can't remove the index on 80-81 for a similar reason of allowing 2 of the same version to exist.

bmoscon:
@rob256 any reason to not merge this?

rob256:
> @rob256 any reason to not merge this?

Hi @bmoscon, I can't see a reason why not! I did re-run the checks but a lot of the tests are failing (similar to the failures on `master`).

I can update the changes.md once we're in a position to merge.

PR #881, opened 2021-01-13
Fix for issue #815

enricodetoma:
I produced a fix for the import errors/deprecations in issue #815.
With a recent version of Pandas, importing Panel results in an error, not just a future warning.
Please have a look at it.

PR #887, opened 2021-02-05
Additional fix for missing pandas.Panel

enricodetoma:
I made an additional fix to manage the case of missing `Panel` in new `pandas` versions.
In fact I discovered that in few cases (e.g. `library.delete(symbol)`), `PandasPanelStore` static method `can_write_type` was still called.
With this commit, `PandasPanelStore` is not even registered as storage class if `Panel` is missing from `pandas`.


shashank88:
@enricodetoma you need to rebase off the latest master after pinning pandas and numpy

enricodetoma:
@shashank88 I rebased off the latest master, but I'm not sure what you mean about pinning pandas and numpy.
This pull request completes my previous pull request #881 which was meant to allow to use Arctic with recent pandas versions, though without Panel support (but then without the need to pin it to a specific version, I guess).

enricodetoma:
Actually I'm already using Arctic successfully with pandas 1.2.3 with this change, without Panel support

crazy25000:
> Actually I'm already using Arctic successfully with pandas 1.2.3 with this change, without Panel support

Same here, using it with latest Pandas. I don't use panels explicitly, just using tick store. Chunk store too slow to use.

samuelrohr:
Any update on this review?

dunckerr (commented):
Hi Enrico - we haven’t removed Panel support because we are still supporting old versions of Pandas. Am I correct that you’re forking arctic repo and removing Panel support?  We are always interested to hear about any improvements to the arctic codebase.
Thanks
Duncan

enricodetoma:
Hi Duncan, I'm simply skipping importing Panel if it doesn't exist:

    try:
        from pandas import Panel
        register_versioned_storage(PandasPanelStore)
    except ImportError:
        pass

PR #898, opened 2021-04-09
Handle uninitialized cache object

shashank88:
If _cache is not initialized - I have seen this in some contexts with forking involved - just return the uncached list_libraries.

Travis seems to be completely busted, checking if it's working for new  PR's, hence closing and reopening one

### DIFF
--- a/arctic/store/_pickle_store.py
+++ b/arctic/store/_pickle_store.py
@@ -95,7 +95,11 @@ class PickleStore(object):
         collection = arctic_lib.get_top_level_collection()
         # Try to pickle it. This is best effort
         version['blob'] = _MAGIC_CHUNKEDV2
-        pickled = cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL)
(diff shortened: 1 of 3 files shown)


arctic/store/_pickle_store.py
```diff
@@ -95,7 +95,11 @@ def write(self, arctic_lib, version, symbol, item, _previous_version):
         collection = arctic_lib.get_top_level_collection()
         # Try to pickle it. This is best effort
         version['blob'] = _MAGIC_CHUNKEDV2
-        pickled = cPickle.dumps(item, protocol=cPickle.HIGHEST_PROTOCOL)
+        # Python 3.8 onwards uses protocol 5 which cannot be unpickled in Python versions below that, so limiting
+        # it to use a maximum of protocol 4 in Python which is understood by 3.4 onwards and is still fairly efficient.
+        # The min() allows lower versions to be used in py2 (which supports a max of 2)
```
rob256, line 100: I imagine we can start deprecating uses of py2 as 2.7 is now EOL.

rob256 (approved):

PR #910, opened 2021-07-21
Bugfix:  Fix column subsetting bug in numpy deserializer

BaiBaiHi:
The following change https://github.com/man-group/arctic/pull/909 introduced a bug with column subsetting. 
**Changes**
* Added tests for column subsetting

**Bugfixes**
* Fixed column subsetting issue 
* Fixed pre-existing error where objify would error out if dataFrame had no index and column requested did not exist in a given chunk.

### DIFF
--- a/arctic/serialization/numpy_arrays.py
+++ b/arctic/serialization/numpy_arrays.py
@@ -145,7 +145,7 @@ class FrameConverter(object):
         """
         cols = columns or doc[METADATA][COLUMNS]
         data = {}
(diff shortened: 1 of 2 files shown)


bmoscon (approved):

bmoscon:
@TomTaylorLondon have a look and then we can merge

TomTaylorLondon (approved):
As discussed, happy with the change to missing columns

PR #928, opened 2021-11-30
Import Panel in the method where it is needed to prevent errors

yaniboshev:
Importing `Panel` in the method where it is explicitly used prevents errors in case of `ImportError`.


tigerstyle80:
Is this the thing holding back support of pandas >=1.1*? If so I think many people will appreciate the change. Thanks.

chicago-joe (approved):

PR #940, opened 2022-01-18
Handle zero rows_per_chunk for ndarray_store

pratverma:
For large ndarrays, rows_size comes out to be greater than CHUNKS_SIZE, thus the value becomes 0. To handle division by zero, add 1 to rows_per_chunk which will not make a difference in the general case


dunckerr:
Pat - why not just add this?
```
if rows_per_chunk == 0:
    rows_per_chunk = 1
```

PR #949, opened 2022-01-28
Improve CI

akkashk:
- Remove Travis CI
- Tidy up circle CI file


akkashk:
@dunckerr PR is ready for review, I've split the changes into self-contained commits since I'm reformatting a lot of files as part of black changes

bmoscon:
I thought in the past someone tried a massive reformatting of everything similar to this (maybe it was using black, maybe not, was years ago) and we opted to not accept it because it made a lot of changes that polluted the change history/blame for little gain. What is the reason behind this change? @jamesblackburn I believe you were involved in the past ticket/PR to which I am referring.

bmoscon:
I'm not opposed to circle, but why not use github actions? They seem much more powerful and flexible than anything I've seen on Travis or Circle

akkashk:
Happy to remove my commit with black changes.

 >does $VERSION get passed into the the individual build sections at the bottom of the file
@dunckerr before we were setting the env var $VERSION in a separate step and writing it to the file in a different step. Now it's all done as part of the same step which should fix the issue

akkashk:
> I'm not opposed to circle, but why not use github actions?

Don't have a strong preference. Our other open source projects use circleci and it's nice being able to share some CI config code between them.

PR #988, opened 2023-03-03
When serializing DateTimeIndex, cast .tz property to str instead of u…

chazkii:
…sing pandas tslib get_timezone


chazkii:
Opening new PR that merges directly into master

PR #990, opened 2023-03-09
When serializing DateTimeIndex, cast .tz property to str instead of u…

chazkii:
…sing pandas tslib get_timezone

### DIFF
--- a/arctic/serialization/numpy_records.py
+++ b/arctic/serialization/numpy_records.py
@@ -61,7 +61,7 @@ def _multi_index_to_records(index, empty_index):
             index_names[i] = 'level_%d' % count
             count += 1
             log.info("Level in MultiIndex has no name, defaulting to %s" % index_names[i])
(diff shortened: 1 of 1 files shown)


chazkii:
@dunckerr good point - I looked through the commit history and use of `get_timezone` was in [the very first commit in 2015](https://github.com/man-group/arctic/commit/31021e7728a077100857bcf01f139cd937e0bb82#diff-9924710cc1a0a0ac30db589a8705abb06745cd8b23f54239ad0490eab1cbda80#L47), meaning initial introduction maybe even older given Arctic was worked on internally.

I've looked at the implementation of [`get_timezone`](https://github.com/pandas-dev/pandas/blame/5a372d892a8c45a2442ab1e744aea3241d2c26a8/pandas/_libs/tslibs/timezones.pyx#L105-L142) and to make things safer, I will only cast to string if it isn't a string already.

PR #1011, opened 2023-06-29
compatibility with newer numpy/pandas

morotti:
* update the pinning from numpy<1.19.0 pandas<1.1.0  to numpy<2 pandas <2
* fix all the errors due to new versions
* fix all the depreciation warnings
* update tests to run on python 3.6 + 3.7 + 3.8 with gradually increasing versions of numpy/pandas.

recommend to review each commit individually, most of the changes are straightforward
except for the get_timezone() and the _arrays_to_mgr()


tests/integration/chunkstore/test_chunkstore.py
```diff
@@ -465,7 +465,7 @@ def helper(chunkstore_lib, chunk_size, name, df, append):
 
         chunkstore_lib.update(name, append)
 
-        if chunk_size is not 'A':
+        if chunk_size != 'A':
```
dunckerr, line 468: is this performant?

mehertz (approved):

arctic/store/_pandas_ndarray_store.py
```diff
@@ -220,6 +213,8 @@ class PandasPanelStore(PandasDataFrameStore):
 
     @staticmethod
     def can_write_type(data):
+        # late import, Panel is removed in pandas 1
+        from pandas import Panel
```
dunckerr, line 217: Will the import happen every time can_write_type gets called?

CHANGES.md
```diff
@@ -1,5 +1,17 @@
 ## Changelog
 
+### HEAD
+ * Compatibility: allow pandas version newer than 1.0, less than 2.0
```
dunckerr, line 4: why "newer than 1.0"?  How is that constraint made?  
if it is made, then do we still need to support pandas 0.22 for the old medusa/pegasus environments?

CHANGES.md
```diff
@@ -1,5 +1,17 @@
...
+ * Compatibility: allow numpy version newer than 1.18
+ * Build: specify which version of python/numpy/pandas to test
+ * Bugfix: resolve import error on pandas.Panel, Panel was removed in pandas 1.0
```
dunckerr, line 7: It'd be nice to have issue numbers for these, but not super important.  
Once merged i'll go thru historic issues fixed by this PR.

arctic/serialization/numpy_records.py
```diff
@@ -63,9 +63,23 @@ def _multi_index_to_records(index, empty_index):
...
+                # https://github.com/pandas-dev/pandas/blob/v0.25.3/pandas/_libs/tslibs/timezones.pyx
+                # https://github.com/pandas-dev/pandas/blob/v1.0.5/pandas/_libs/tslibs/timezones.pyx
+                # https://github.com/pandas-dev/pandas/blob/v1.3.5/pandas/_libs/tslibs/timezones.pyx
+                # get_timezone(mktz("UTC")) returns 'dateutil//usr/share/zoneinfo/UTC' string before pandas 1.3
+                # get_timezone(mktz("UTC")) returns the timezone object in pandas 1.3+
+                # we need a stable string to serialize and to deserialize,
+                # so let's always encode the filename like pandas is doing for standard datetime objects.
+                # NOTE: constrary to what it looks like, the path is not dependent on the machine,
+                #       there's logic in pandas and datetime to strip the directory prefix.
+                tmp = "dateutil/" + i.tz._filename
```
dunckerr, line 81: will this work on windows?

benchmarks/async/async_benchmark.py
dunckerr, line 1: was this ever run as part of testpack?

tests/integration/chunkstore/test_chunkstore.py
morotti: it's a warning from the linter, should use `!=` for comparisons instead of `is not`

arctic/store/_pandas_ndarray_store.py
morotti: a module is only loaded once by the interpreter. all subsequent imports are returning the already-loaded module, it's very quick.

CHANGES.md
morotti: I have adjusted the message: update pinning from pandas<1.1, to pandas<2

we're still testing on python 3.6 with pandas 0.x

arctic/serialization/numpy_records.py
morotti: I think so, this is the line of code that was returned by `pandas.get_timezone()`. 

it's encoding timezone to that string for serialization/deserialization, across all platforms.

benchmarks/async/async_benchmark.py
morotti: This file is not importable, invalid python syntax. 
It's not run by any test, or we'd have noticed it was broken.

best guess, this was a manually run benchmark that hasn't been used for many years.

CHANGES.md
morotti: I've added numbers for the Panel import error, it's the only bug that had open PRs.

dunckerr (approved):
Looks good

PR #1013, opened 2023-10-06
fix: Handle index timezones of dateutil.timezone.tzutc()

lwfitzgerald:
Due to an earlier change in this area, a regression was introduced where `dateutil.timezone.tzutc()` was serialized as the string `"tzutc()"`. `pandas` is unable to create a timezone from this, so data could not be reconstructed.

This change restores the old behaviour where `dateutil.timezone.tzutc()` is serialized as the string `"UTC"`.

Another prior fix to consistently serialize `dateutil.timezone.gettz("UTC")` for multi-level indexes is also extended in this change to cover single level indexes.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,6 +1,8 @@
 ## Changelog
 
 ### HEAD
+ * Bugfix: #1013 Fix serialized tz name for index timezones of dateutil.timezone.tzutc()
+ * Bugfix: #1013 Fix serialized tz name for single-level index timezones of dateutil.timezone.gettz("UTC")
   
 ### 1.82.0 (2023-07-24)
  * Release: #1012 version 1.82.0
--- a/arctic/serialization/numpy_records.py
+++ b/arctic/serialization/numpy_records.py
@@ -1,5 +1,8 @@
+import datetime
 import logging
+from typing import List, Optional, Union
 
+import dateutil
 import numpy as np
 from pandas import DataFrame, MultiIndex, Series, DatetimeIndex, Index
 import pandas as pd
@@ -8,6 +11,12 @@ import pandas as pd
 from .._config import FAST_CHECK_DF_SERIALIZABLE
 from .._util import NP_OBJECT_DTYPE
 from ..exceptions import ArcticException
+
+try:  # 1.3.0+ Compatibility
+    from pandas._libs.tslibs.timezones import is_utc
+except ImportError:  # < 1.3.0, this function is unavailable, but should never be called. Stub to satisfy linting.
+    def is_utc(tz):
+        raise AssertionError("Should never be called for Pandas versions where this is not an exported function")
 
 try:  # 0.21+ Compatibility
     from pandas._libs.tslib import Timestamp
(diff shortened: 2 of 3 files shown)


CHANGES.md
dunckerr, line 1: Can you add the issue nuber (1013) to changelog, for consistency.

arctic/serialization/numpy_records.py
```diff
@@ -47,6 +56,55 @@ def _to_primitive(arr, string_max_len=None, forced_dtype=None):
...
+    #   `pandas` 0.22.0 - https://github.com/pandas-dev/pandas/blob/v0.22.0/pandas/_libs/tslibs/timezones.pyx#L71-L72
+    #      - Returns "UTC" for UTC `tzinfo`s, except:
+    #        - `dateutil.tz.gettz("UTC")` - Returns a string like "dateutil//usr/share/zoneinfo/UTC"
+    #   `pandas` 0.24.0 - https://github.com/pandas-dev/pandas/blob/v0.24.0/pandas/_libs/tslibs/timezones.pyx#L59-L60
+    #      - Returns the `tzinfo` object for UTC `tzinfo`s except:
+    #        - `dateutil.tz.gettz("UTC")` - Returns a string like "dateutil//usr/share/zoneinfo/UTC"
+    #   `pandas` 1.3.0 -  https://github.com/pandas-dev/pandas/blob/v1.3.0/pandas/_libs/tslibs/timezones.pyx#L53
+    #      - Returns the `tzinfo` object for UTC `tzinfo`s (including `dateutil.tz.gettz("UTC")`)
+    if PD_VER < "0.24.0":
+        return str(get_timezone(tz))
```
dunckerr, line 91: Can we get to this point with tz==None?

arctic/serialization/numpy_records.py
```diff
@@ -91,7 +134,7 @@ class PandasSerializer(object):
     def _index_to_records(self, df):
         metadata = {}
         index = df.index
-        index_tz = None
+        index_tz: Optional[Union[str, List[str]]]
```
morotti, line 137: the variable is not created, should it be created to None?

`index_tz: Optional[Union[str, List[str]]] = None`

arctic/serialization/numpy_records.py
lwfitzgerald: This function is only called for non-`None` `tz`s. The type hint reflects this: https://github.com/man-group/arctic/pull/1013/files#diff-1c730dd7d8c45baa8edaa77fcdd8b32c1c10c91c370a9aa46b218f59fbe07a0eR69

arctic/serialization/numpy_records.py
lwfitzgerald: `index_tz` is guaranteed to be set by the following code. The type hint here just declares the type of the variable.

> ```py
> # You don't need to initialize a variable to annotate it
> a: int  # Ok (no value at runtime until assigned)
> ```
https://mypy.readthedocs.io/en/stable/cheat_sheet_py3.html

lwfitzgerald: Just spotted it does need a small correction though to:
```py
index_tz: Union[Optional[str], List[Optional[str]]]
```

as if it is a `List`, the entries can be `None` as well.

dunckerr (approved):

PR #1021, opened 2024-02-01
Speedup tickstore read date index conversion

BaiBaiHi:
Speedup of Tickstore read datetime index conversion by ~40x.

Overall runtime improvement is around 3-4x on average, but this can vary depending on the mongo latency. 

```
%timeit  pd.DatetimeIndex(np.concatenate(rtn[INDEX]).astype('datetime64[ms]'), tz='UTC')
1.76 s ± 59.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
%timeit pd.to_datetime(np.concatenate(rtn[INDEX]), utc=True, unit='ms')
1min 16s ± 1.74 s per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

**In the current implementation, the datetime conversion takes ~60% of the total runtime:**

image

**The new implementation drops it down to ~5%**

image

**Output of both implementations is identical (tested on 22.5 million values).** 

image

### DIFF
--- a/arctic/tickstore/tickstore.py
+++ b/arctic/tickstore/tickstore.py
@@ -334,7 +334,7 @@ class TickStore(object):
             raise NoDataFoundException("No Data found for {} in range: {}".format(symbol, date_range))
         rtn = self._pad_and_fix_dtypes(rtn, column_dtypes)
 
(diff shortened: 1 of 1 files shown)


jamesmunro (approved):

jamesmunro:
Thanks!

